PR #9, opened 2022-08-27
Add ml library for LLBC

giltho:
I'd like to try Charon as a front-end for Gillian-Rust, which is also written in OCaml.
I believe the LLBC library should also live here, as to stay synchronised with Charon itself.
If the PR is merged, Aeneas should also be migrated to using this library instead.


giltho:
I also took the liberty of adding a github action that runs the tests
The main reason I did it is to make sure that charon and ocaml's llbc-lib stay in sync before merging PRs, but I can definitely revert that if you're against it of couse

msprotz:
Hi Sacha,

Thanks for the PR! Do I understand correctly that the goal is to:
- move the LLBC de-serializer (from JSON), helpers and AST definitions to this repository to keep them closer to the Charon serializer?
- expose an OPAM package called llbc-lib
- submit a subsequent PR to Aeneas to make sure it uses the new llbc-lib?

This looks overall like a good idea, especially if these definitions are stable and day-to-day development on Aeneas is unlikely to require constantly recompiling, re-installing and updating Aeneas. @sonmarcho is on vacation right now so let's wait until he's back to comment.

Can you comment a little bit on what the workflow would look like hacking on Aeneas with this new setup? Maybe `opam pin` for the llbc-lib package?

Thanks again!

Jonathan

PS: I have a few minor tidbits that I'm going to add in the review

.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+        toolchain: nightly-2022-01-29
+        components: rustc-dev, llvm-tools-preview
+    - name: "build rs"
+      run: make
+    - name: "test rs"
+      run: make test
+    - name: "Setup ocaml 4.12.1"
+      uses: ocaml/setup-ocaml@v2
+      with:
+        ocaml-compiler: 4.12.x
```
msprotz, line 33: why not the latest?

Makefile
```diff
@@ -4,6 +4,13 @@ all: build build-tests build-tests-nll
 .PHONY: build
 build:
 	cd charon && make
+	
+ml: ocaml-build
+
+ml-test: ocaml-build ocaml-fmt-check ocaml-tests
+	
+ocaml-build:
```
msprotz, line 12: technically this one should be declared phony as well

ml/dune-project
```diff
@@ -0,0 +1,3 @@
+(lang dune 3.0)
+
+(cram enable)
```
msprotz, line 3: I don't think this is mandatory anymore as of dune 3.0

Makefile
```diff
@@ -52,3 +60,11 @@ test-nll-%: TESTS=../tests-nll
 test-nll-%: OPTIONS += --nll
 test-nll-%:
 	cd charon && cargo run $(SRC)/$*.rs $(OPTIONS)
+
+ocaml-tests:
```
msprotz, line 64: also phony and the one below too: make is not aware of what it takes to regenerate dune builds

.github/workflows/build.yml
giltho: OCaml maintains backward compatibility, so as a rule of thumbs we might as well use an older version.
I chose 4.12 because that's what [Aeneas says it uses](https://github.com/AeneasVerif/aeneas#installation--build), but  anything is fine there

giltho:
> Can you comment a little bit on what the workflow would look like hacking on Aeneas with this new setup? Maybe opam pin for the llbc-lib package?

I think there are 2 choices there:
- `opam pin`
- adding Charon as a submodule in Aeneas.

I think that's basically up to you guys on that one. I think `opam pin` works quite well for dev packages, at the cost of an annoying one-time setup

ml/dune-project
giltho: correct, TIL

giltho:
(I'm on the wrong computer right now, I'll fix this tonight at home)

giltho:
Note that I'm not sure at the moment that LLBC corresponds to my needs - I'm verifying unsafe code, so I'd need a strict superset of LLBC (I already have a PR ready with raw pointers implemented)
I'll try to gather data about using LLBC for my use case, figure out the necessary changes and write a full document about it. Also, I'll be in Paris in October and I already plan on visiting Son, so we can maybe talk about that :)

msprotz:
Yes, it would be great to talk. Our ambition had always been to have LLBC serve as a platform for other verification tools, so if it turns out that you can build upon the existing work and extend the supported subset of Rust, it'd be ideal! Will you be at ICFP?

giltho:
I unfortunately will not be at ICFP :(

ml/tests/dune
```diff
@@ -0,0 +1,6 @@
+(executable
+ (name e2e)
```
sonmarcho, line 2: What does `e2e` stand for?

ml/src/values.ml
```diff
@@ -0,0 +1,841 @@
...
+(** A constant value *)
+type constant_value =
+  | Scalar of scalar_value
+  | Bool of bool
+  | Char of char
+  | String of string
+[@@deriving show]
+
+(** The kind of a symbolic value, which precises how the value was generated *)
+type sv_kind =
```
sonmarcho, line 47: There are many things here which are specific to Aeneas (not Charon). We should split the file - only the constant values should be used on Charon's side.

sonmarcho (changes requested):
Thanks a lot for the PR!

It indeed makes sense to move the deserialization utilities to Charon, and turn them into a library.
Also thanks for the Github actions: this is one of the many extremely useful things I hadn't had time to setup...

I just have the following comments:

- The `values.ml` file should be split, because it contains many definitions which are specific to Aeneas and its symbolic execution. We should only keep the definitions relative to `constant_value` (and put them in a `ConstantValue.ml` file).

- Why change the file names from Caml case to snake case? I thought the convention was Caml case for the filenames (please correct me if I'm wrong).

- It would make sense to also move the printing utilities for LLBC to this library (they're currently still in the Aeneas project).

I can take care of those modifications myself.


Finally, I suggest proceeding as follows:

- I want to merge this [PR](https://github.com/AeneasVerif/charon/pull/7) first, because it was submitted first and has been stalled for too long (the Charon PR is ready, but I need to review and merge the corresponding changes on the Aeneas side)

-  once this is done, I want to update Aeneas to use this ML library (and make sure we didn't break anything at the same time)

sonmarcho:
I'm adding a small note: the reason why I work with OCaml 4.12.1 is because F* doesn't support the latest versions of OCaml. I will update the code to use the latest version of OCaml and switch version when building my various projects.

sonmarcho:
Also, I'm wondering whether Aeneas should use a sub module or `opam pin`. Which one do you think is the easier to use to do incremental development @msprotz ?

msprotz:
`opam pin` will work better with the opam ecosystem, I think

ml/tests/dune
giltho: "End to end", because these are not unit test, it executes the thing entirely and check for any failure. It'd be nice to have better tests but that requires more work...

giltho:
Sorry it took me time to come back to that, end of internship and all that. I'll answer these comments now and update the PR soon.

> The values.ml file should be split, because it contains many definitions which are specific to Aeneas and its symbolic execution. We should only keep the definitions relative to constant_value (and put them in a ConstantValue.ml file).

Interesting, I haven't looked too closely into it. Are you sure none of that can be reused by someone else?

> Why change the file names from Caml case to snake case? I thought the convention was Caml case for the filenames (please correct me if I'm wrong).

It depends on sub-communities etc, but the adoption of dune everywhere has kind of pushed snake_case. See [the code of dune itself](https://github.com/ocaml/dune/tree/main/src). The way I remember it is that python uses caml case and OCaml uses snake case 🙃 

> It would make sense to also move the printing utilities for LLBC to this library (they're currently still in the Aeneas project).

Definitely! Didn't notice that.

> Finally, I suggest proceeding as follows

That sounds good

> I can take care of those modifications myself

Ah, as you prefer, I don't mind making the changes myself. If you think it'll be difficult for me to separate the Aeneas-specific values and the rest, you can do it

giltho:
Closing because Son implemented everything properly I believe

sonmarcho:
I had forgotten to comment this PR after merging https://github.com/AeneasVerif/charon/pull/18: I still need to add the CI (github actions), but I will do that later

PR #10, opened 2022-09-27
Some more logic to work on OSX

msprotz:
This is just a small oddity I found while trying to get things up and running on my machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,25 @@
+ifeq (3.81,$(MAKE_VERSION))
+  $(error You seem to be using the OSX antiquated Make version. Hint: brew \
+    install make, then invoke gmake instead of make)
+endif
+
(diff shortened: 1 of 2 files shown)


msprotz:
Some comments.
- OSX ships a super old version of make because of gplv2 vs gplv3
- that old version of make does not support the shortest stem rule
- you rely on it when you expect "foo-nll-%" to take precedence over "foo-%" for a target of the form "foo-nll-bar"

The recursive calls need to be done through `$(MAKE)` so that you really invoke the same make executable with the same parallel options.

sonmarcho:
I made a minor modification: can you check that it still works?

R1kM:
With the latest fix, the build works on my mac M1

sonmarcho (approved):

PR #11, opened 2022-10-07
Add Nix support

pnmadelaine:


msprotz (approved):
Dear @pnmadelaine 

Thank you so much for submitting this PR! As you know, I am a big fan of Nix and it's always a delight to see more and more projects adopt it.

Overall, the PR looks good. A few minor comments:
- do we need lock files everywhere?
- if so, why no lock file for the macros?
- your flake seems to be replicating logic from the Makefile -- how can we keep the two in sync? any way to just invoke `make` in the buildPhase?

Thanks again, and long live Nix.

pnmadelaine:
Dear @msprotz, I'm really happy to see the interest you have in seeing Nix broadly adopted.
To answer your very interesting questions:

-do we need lock files everywhere?

Yes, if we want to benefit from the full capabilities of Nix regarding reproducibility.

-if so, why no lock file for the macros?

Because the macros crate is not meant to be built, only imported by the other crates in Charon.

-any way to invoke `make` in the buildPhase?

This is exactly what we've been discussing with @sonmarcho, it should be possible with a few minor changes to the Makefile.

Don't hesitate to contact me if you have more questions regarding this PR in particular, or if you simply want to discuss other Nix endeavours.

sonmarcho:
@msprotz I'm indeed working on invoking `make` in the build phase. But first I've been fixing the way the Charon driver invokes `rustc`, which as discussed was not robust (especially with regards to the handling of the external dependencies) and prevented us from running some tests with Nix. I updated Charon to piggyback on `cargo` by taking some inspiration from [rust-clippy](https://github.com/rust-lang/rust-clippy): it seems to work well so far - I just have a few modifications left to do.

sonmarcho:
I'm closing this as it is now superseded by https://github.com/AeneasVerif/charon/pull/12

PR #12, opened 2022-10-20
Piggy back on Cargo and implement Nix support

sonmarcho:
This PR started as https://github.com/AeneasVerif/charon/pull/11 (thanks a lot @pnmadelaine for initiating this and all your support when debugging Nix) and ended up a lot bigger (in particular I **had** to update the way I call Rustc to make the Nix build succeed). It does the following:
* completely change the way we call Rustc by piggy-backing on Cargo (inspired by [clippy](https://github.com/rust-lang/rust-clippy)) so that we handle external dependencies in a more robust way. This works by having two binaries: `cargo-charon` which calls Cargo while setting the environment variable `RUST_WORKSPACE_WRAPPER`, and `charon-driver` (the Rustc driver) which is later called by Cargo with proper arguments and links itself to Rustc (while transmitting the arguments given by Cargo)
* implement support for Nix: build the project and run the tests
* add a `--release` option to extract the MIR compiled for release (changes a few things with the way we simplify some operations - a lot of dynamic checks are removed, but not all of them)
* add an option to extract the optimized MIR (the lowest-level MIR) rather than the built MIR (the highest-level MIR), together with some tests. It is important to maintain our extraction of optimized MIR: some people need it, and the extraction is not done the same way at all in many places, in particular with regards to the handling of constants (see [`translate_functions_to_im.rs`](https://github.com/AeneasVerif/charon/blob/son_flake/charon/src/translate_functions_to_im.rs))
* add a library `charon`, so that the project can be reused in other applications
* build a proper documentation for the various targets (`macros`, `cargo-charon`, `charon-driver` and `charon`)

I initially also wanted to upgrade the pinned version of Rustc as well as the dependencies, but updating the code to account for the changes in Rustc proved to be a lot more work than expected (they completely change the way constants are handled, and in optimised MIR the handling of boxes is desugared to some very low level code). Another PR is coming for that.


sonmarcho:
Tagging @msprotz for the information, and @pnmadelaine so that you can check that the `flake.nix` file reaches your standard of quality :)

pnmadelaine:
looks good to me :)

PR #30, opened 2023-06-19
Add support for array aggregates

msprotz:
This adds support for serializing array aggregates. Some notes:
- there seemed to be an issue in the serialization functions where the AggregateKind::Adt was using the same variant number as AggregateKind::Options... was this intentional? (reading the doc, I would say no)
- I pointed out one potential inefficiency as a TODO but I don't think it's a big deal.

### DIFF
--- a/charon/src/expressions.rs
+++ b/charon/src/expressions.rs
@@ -211,7 +211,7 @@ pub enum Rvalue {
     Global(GlobalDeclId::Id),
 }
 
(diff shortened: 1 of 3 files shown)


sonmarcho:
Indeed, there was a bug in the serialization. I'm not sure why it never got triggered...

sonmarcho (approved):

PR #44, opened 2023-09-22
Add Charon support for Shl/Shr

R1kM:
Adds Charon support for the bitshifts operators, as well as a small basic test.
To be merged with the companion Aeneas PR.


sonmarcho:
I made minor modifications and updated the tests to use the other bitwise operations (and, or, xor) as well.

PR #87, opened 2024-03-18
Handle binary and

msprotz:
For some reason I had this in my stash -- I thought I had fixed it, but couldn't find the commit anywhere.

Please check it looks ok? Not sure which other operators we should add to the list.

### DIFF
--- a/charon/src/remove_dynamic_checks.rs
+++ b/charon/src/remove_dynamic_checks.rs
@@ -210,7 +210,9 @@ impl<'tcx, 'ctx, 'a> RemoveDynChecks<'tcx, 'ctx, 'a> {
                 {
                     // We don't check that the second operand is 0 in
                     // case we are in the division/remainder case
-                    if matches!(binop, BinOp::Eq) && is_assert_move(dest_p, s1, false) {
+                    if matches!(binop, BinOp::Eq | BinOp::BitAnd)
+                        && is_assert_move(dest_p, s1, false)
+                    {
                         // This should be the division/remainder case
                         // Eliminate the first two statements
                         take(s, |s| {


msprotz:
here's a testcase from Franziskus in case you want to add it to the test suite (which prompted me to resurrect my fix):

```
pub fn modulus(value: &[i32]) {
    let mut result = [0i32];
    for i in 0..value.len() {
        result[i] = value[i] % 3329;
    }
}
```

franziskuskiefer:
Looks like this doesn't working for the actual code we have.

https://github.com/cryspen/libcrux/blob/baf808ecaa0b265b668d30b748919cac4fde463e/src/kem/kyber/serialize.rs#L117-L124

sonmarcho:
Can you provide the code which triggers an error? I can have a look.

sonmarcho:
I see what's going on: working on it.

sonmarcho:
The binops simplification is incorrect when using signed integers. In particular, for the following code:
```rust
pub fn modulus(result: &mut [i32], i: usize, x: i32) {
    // Bugs
    result[i] = x / 3329;
}
```
We get the LLBC:
```
fn tests::modulus<'_0>(@1: &'_0 mut (Slice<i32>), @2: usize, @3: i32)
{
    let @0: (); // return
    let result@1: &'_ mut (Slice<i32>); // arg #1
    let i@2: usize; // arg #2
    let x@3: i32; // arg #3
    let @4: i32; // anonymous local
    let @5: bool; // anonymous local
    let @6: bool; // anonymous local
    let @7: usize; // anonymous local
    let @8: (); // anonymous local
    let @9: &'_ mut (Slice<i32>); // anonymous local
    let @10: &'_ mut (i32); // anonymous local

    @4 := copy (x@3)
    @5 := const (3329 : i32) == const (-1 : i32) // SHOULDN'T BE HERE
    @6 := copy (@4) == const (-2147483648 : i32) // SHOULDN'T BE HERE
    @7 := copy (i@2)
    @9 := &mut *(result@1)
    @10 := @SliceIndexMut<'_, i32>(move (@9), copy (@7))
    *(@10) := move (@4) / const (3329 : i32)
    drop @4
    drop @7
    @8 := ()
    @0 := move (@8)
    @0 := ()
    return
}
```

If we use the type `u32` instead:
```rust
pub fn modulus1(result: &mut [u32], i: usize, x: u32) {
    // Succeeds
    result[i] = x / 3329;
}
```

We get a valid LLBC:
```
fn tests4::modulus1<'_0>(@1: &'_0 mut (Slice<u32>), @2: usize, @3: u32)
{
    let @0: (); // return
    let result@1: &'_ mut (Slice<u32>); // arg #1
    let i@2: usize; // arg #2
    let x@3: u32; // arg #3
    let @4: u32; // anonymous local
    let @5: usize; // anonymous local
    let @6: (); // anonymous local
    let @7: &'_ mut (Slice<u32>); // anonymous local
    let @8: &'_ mut (u32); // anonymous local

    @4 := copy (x@3)
    @5 := copy (i@2)
    @7 := &mut *(result@1)
    @8 := @SliceIndexMut<'_, u32>(move (@7), copy (@5))
    *(@8) := move (@4) / const (3329 : u32)
    drop @4
    drop @5
    @6 := ()
    @0 := move (@6)
    @0 := ()
    return
}
```

sonmarcho:
@Nadrieril can you take care of this? It requires modifying the micro pass `remove_dynamic_checks.rs`

Nadrieril:
What's happening is that the micro-pass expects the overflow checks to happen just before the division, but here there's some array indexing in between. We don't actually need the division to be there though, I'll try to fix that

Nadrieril:
https://github.com/AeneasVerif/charon/pull/124 should fix your issue

PR #98, opened 2024-04-03
Add a test harness for the llbc output

Nadrieril:
The charon-ml tests only check the structure of the output; until now there were no tests in this repo that tested the contents. This adds a test harness so we can check the contents a little bit.


sonmarcho (approved):
This is really great, thanks for this!

Issue #63, opened 2023-12-20
Add support for attributes


sonmarcho:
It would be good to have the possibility of annotating the Rust definitions with attributes, for instance to mark some definitions as `opaque`, to tweak the name of the definitions for the extraction, etc.

PR #99, opened 2024-04-03
Track item attributes

Nadrieril:
Fixes https://github.com/AeneasVerif/charon/issues/63

This breaks aeneas because it adds new fields and aeneas does exhaustive destructurings.


Nadrieril:
will do the `ItemMeta` refactor first

PR #102, opened 2024-04-04
Add test harness for llbc pretty output

Nadrieril:
This will make it a lot easier to debug issues like https://github.com/AeneasVerif/aeneas/issues/97

### DIFF
--- a/charon/src/charon-driver.rs
+++ b/charon/src/charon-driver.rs
@@ -210,7 +210,7 @@ fn main() {
     let mut callback = CharonCallbacks::new(options);
     let mut res = callback.run_compiler(compiler_args);
     if let Some(crate_data) = &callback.crate_data {
-        if !callback.options.no_serialize_llbc {
+        if !callback.options.no_serialize {
             // # Final step: generate the files.
             // `crate_data` is `None` on the first call of the driver.
             res = res.and_then(|()| {
@@ -233,7 +233,7 @@ fn main() {
             // Standard rust panic error code.
             std::process::exit(101);
         }
-        Err(_) => {
+        Err(CharonFailure::RustcError(_) | CharonFailure::Serialize) => {
             assert!(!errors_as_warnings);
             let msg = format!("The extraction encountered {} errors", callback.error_count);
             log::error!("{}", msg);
--- a/charon/src/charon-driver.rs
+++ b/charon/src/charon-driver.rs
@@ -87,7 +87,6 @@ mod values;
 mod values_utils;
 
 use crate::driver::{arg_value, get_args_crate_index, get_args_source_index, CharonCallbacks};
-use rustc_driver::RunCompiler;
 
 fn main() {
     // Initialize the logger
@@ -101,9 +100,6 @@ fn main() {
         "Impossible: zero arguments on the command-line!"
     );
     trace!("original arguments (computed by cargo): {:?}", origin_args);
-
-    // The execution path (the path to the current binary) is the first argument
-    let exec_path = origin_args[0].clone();
 
     // Retrieve the Charon options by deserializing them from the environment variable
     // (cargo-charon serialized the arguments and stored them in a specific environment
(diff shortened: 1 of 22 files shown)


Nadrieril:
I managed to break CI!

Nadrieril:
Converting to draft while I improve it

Nadrieril:
It's ready!

charon/src/cli_options.rs
```diff
@@ -110,27 +123,39 @@ performs: `y := (x as E2).1`). Producing a better reconstruction is non-trivial.
...
+    #[serde(default)]
     pub errors_as_warnings: bool,
+    #[structopt(
+        long = "no-serialize-llbc",
+        help = "
+Don't serialize the final LLBC to a file.
+"
+    )]
+    #[serde(default)]
+    pub no_serialize_llbc: bool,
```
sonmarcho, line 135: It would be better to name this `no-serialize` as we can serialize ULLBC as well?

charon/src/charon-driver.rs
```diff
@@ -231,6 +229,10 @@ fn main() {
                 log::warn!("{}", msg);
             }
         }
+        Err(CharonFailure::Panic) => {
+            // Standard rust panic error code.
+            std::process::exit(101);
+        }
         Err(_) => {
```
sonmarcho, line 236: It is good to make the `CharonFailure` kind explicit here (useful when we update this enumeration, to make sure we don't forget any case)

charon/tests/ui.rs
```diff
@@ -0,0 +1,177 @@
+//! Ui tests for the charon compiler. Each `<file>.rs` file in `ui/` will be passed to the charon
+//! driver. The corresponding pretty-printed llbc output will be stored in `<file>.llbc`, and CI
+//! will ensure these stay up-to-date.
+//!
+//! Files can start with special comments that affect the test behavior. Supported magic comments:
+//! see `HELP_STRING`.
```
sonmarcho, line 6: I think there is a different syntax to refer to a definition in a comment? Something like [HELP_STRING]?...

charon/tests/ui.rs
```diff
@@ -0,0 +1,177 @@
...
+
+enum TestKind {
+    PrettyLlbc,
+    KnownPanic,
+    Unspecified,
+}
+
+struct MagicComments {
+    test_kind: TestKind,
+}
```
sonmarcho, line 29: Why "magic"?

charon/tests/ui.rs
```diff
@@ -0,0 +1,177 @@
...
+        if !line.starts_with("//@") {
+            break;
+        }
+        let line = line[3..].trim();
+        if line == "known-panic" {
+            kind = TestKind::KnownPanic;
+        } else if line == "output=pretty-llbc" {
+            kind = TestKind::PrettyLlbc;
+        } else {
+        }
```
sonmarcho, line 51: Why not return after we successfully parse the first magic comment?

charon/tests/ui/issue-92-nonpositive-variant-indices.rs
```diff
@@ -0,0 +1,13 @@
+//@ output=pretty-llbc
```
sonmarcho, line 1: My understanding is that this PR requires https://github.com/AeneasVerif/charon/pull/107 to be merged before?

charon/tests/ui/issue-97-missing-parent-item-clause.rs
```diff
@@ -0,0 +1,18 @@
+//@ output=pretty-llbc
```
sonmarcho, line 1: Same: we need issue 97 to be merged before the current PR?

charon/tests/ui/issue-92-nonpositive-variant-indices.rs
Nadrieril: No, I just added a few test cases from open issues. As you can see, CI passes (as you can see, the match is weird)

sonmarcho: Indeed

charon/tests/ui.rs
Nadrieril: I think this is what the ui-test crate calls them. I could call them something else if you have an idea

charon/tests/ui.rs
Nadrieril: Yeah fair the loop doesn't make sense yet. I wrote it like this because I plan to add other options, e.g. `//@ driver-opts=--no-code-duplication` depending on what comes useful.

charon/src/cli_options.rs
```diff
@@ -110,27 +123,39 @@ performs: `y := (x as E2).1`). Producing a better reconstruction is non-trivial.
 Report the errors as warnings
 "
     )]
+    #[serde(default)]
     pub errors_as_warnings: bool,
+    #[structopt(
+        long = "no-serialize",
+        help = "
+Don't serialize the final LLBC to a file.
```
sonmarcho, line 131: Update this message: it can be the ULLBC

charon/tests/ui.rs
sonmarcho: Makes sense, and it's not that important anyway

sonmarcho (approved):
Looks good to me! I just have one comment I want to see you address: once it's done, you can directly merge the PR (no need to wait for me to reapprove and merge).

Issue #69, opened 2023-12-20
Extract inlining information


sonmarcho:
This issue will probably require https://github.com/AeneasVerif/charon/issues/63

Issue #68, opened 2023-12-20
Extract visibility information


sonmarcho:
We need to extract visibility information like `pub` or `pub(crate)`, etc.

PR #105, opened 2024-04-08
Track item attributes and visibility

Nadrieril:
Fixes https://github.com/AeneasVerif/charon/issues/63. Fixes https://github.com/AeneasVerif/charon/issues/69. Fixes https://github.com/AeneasVerif/charon/issues/68.

I made `attributes` be a list of strings for now, we can have something more structured if needed later. I made `public` a boolean, we can gather something more precise if needed. Note the comment about pub-in-priv; whether that's ok or not depends on what we need the info for.


Nadrieril:
Corresponding aeneas PR: https://github.com/AeneasVerif/aeneas/pull/116

charon-ml/src/Meta.ml
```diff
@@ -43,3 +43,17 @@ type meta = {
...
 [@@deriving show, ord]
+
+type inline_attr =
+  | Hint  (** `#[inline]` **)
+  | Never  (** `#[inline(never)]` **)
+  | Always  (** `#[inline(always)]` **)
+[@@deriving show, ord]
+
+type item_meta = {
+  meta : meta;
```
R1kM, line 54: Given that there are now several types of meta, could we rename this one in something more explicit? e.g., span_meta? @sonmarcho , wdyt?

sonmarcho: I agree. Let's do that in a subsequent PR (the diff will be huge). @Nadrieril can you open a PR for this?

Nadrieril: you mean an issue? I can also open a PR but that'll take a moment ^^

charon/src/translate_ctx.rs
```diff
@@ -482,6 +492,90 @@ impl<'tcx, 'ctx> TransCtx<'tcx, 'ctx> {
...
+            | ForeignMod
+            | Generator
+            | GlobalAsm
+            | Impl { .. }
+            | ImplTraitPlaceholder
+            | InlineConst
+            | LifetimeParam
+            | OpaqueTy
+            | TyParam
+            | Variant => false,
```
sonmarcho, line 575: Are we sure we want the default to be `false`? I think that given the places where you call `translate_visibility_from_rid` we should simply panic here.

sonmarcho (changes requested):

charon/src/translate_ctx.rs
Nadrieril: That would mean we can't reuse `translate_attributes_from_rid` for e.g. trait impls

Nadrieril: We could have visibility be `Option<bool>` to indicate missing info

sonmarcho: I guess the fact that an `impl` should be considered public or not depends on the trait and the type for which it is implemented.
I think we should: 1. add a comment here to explain we need to treat trait impls separately (and explain why), and that we return `false`by default for them; 2. panic for the others; 3. add a comment to the `public` field of `ItemMeta` to explain that in case of trait implementations it has no meaning (and is `false` by default).

Nadrieril: > I guess the fact that an impl should be considered public or not depends on the trait and the type for which it is implemented (correct?).

In a sense, but we also have non-trait impl blocks and, I just discovered, closures.

Agreed on the steps to take.

charon/src/meta.rs
```diff
@@ -92,8 +92,9 @@ pub struct ItemMeta {
     pub attributes: Vec<Attribute>,
     /// Inline hints (on functions only).
     pub inline: Option<InlineAttr>,
-    /// Whether this item is public. Note that this considers pub-in-priv items to be public.
-    /// Computing actual reachability is harder.
+    /// Whether this item is public. Note that this considers pub-in-priv items to be public;
```
sonmarcho, line 95: I still don't understand what "pub-in-priv" means

charon/src/translate_ctx.rs
```diff
@@ -553,24 +558,24 @@ impl<'tcx, 'ctx> TransCtx<'tcx, 'ctx> {
...
             | Generator
             | GlobalAsm
-            | Impl { .. }
             | ImplTraitPlaceholder
             | InlineConst
             | LifetimeParam
             | OpaqueTy
             | TyParam
-            | Variant => false,
+            | Variant => panic!("Called `translate_visibility_from_rid` on `{def_kind:?}`"),
```
sonmarcho, line 578: Having an option looks better.
Could you also use the `error_or_panic` macro so that we can recover from failure?

sonmarcho (changes requested):

charon/src/translate_ctx.rs
Nadrieril: If we use an option, we won't catch if this function is called on an unexpected kind (because we `unwrap_or(false)`)

charon/src/meta.rs
```diff
@@ -70,6 +70,33 @@ pub struct Meta {
...
+/// Meta information about an item (function, trait decl, trait impl, type decl, global).
+#[derive(Debug, Clone, Serialize)]
+pub struct ItemMeta {
+    pub meta: Meta,
+    /// Attributes (`#[...]`).
+    pub attributes: Vec<Attribute>,
+    /// Inline hints (on functions only).
+    pub inline: Option<InlineAttr>,
+    /// Whether this item is public. Note that this considers pub-in-priv items to be public.
+    /// Computing actual reachability is harder.
```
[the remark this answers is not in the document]
Nadrieril:
```rust
mod foo {
  pub fn bar() {}
}
```
That's a function declared `pub` inside a module that is private. This is commonly called "pub-in-priv". This function is not present in the public API, yet it is declared `pub` so `pub use foo:bar;` would be valid. In other words, to determine whether this function appears in the public API of the crate we need to look through every module and import so see if there's a path that exposes it.

For types, it's worse: we also need to look through type signatures:
```rut
mod foo {
    pub struct X;
}
mod bar {
    pub fn something(_x: super::foo::X) {}
}
pub use bar::something; // exposes `X`
```

charon/src/translate_ctx.rs
sonmarcho: I meant: the fact that you're using an option is better than the previous version.

sonmarcho (approved):
Looks good, thanks! If you add the comment above in the code, you're free to merge

PR #107, opened 2024-04-09
Unconfuse variant indices and discriminants

Nadrieril:
They're conceptually different entities but happen to coincide for basic enums, hence the confusion.

Fixes https://github.com/AeneasVerif/charon/issues/92.

### DIFF
--- a/charon/src/remove_read_discriminant.rs
+++ b/charon/src/remove_read_discriminant.rs
@@ -42,13 +42,29 @@ impl<'a, 'tcx, 'ctx> Visitor<'a, 'tcx, 'ctx> {
                 let (meta2, switch, st3_opt) = match st2.content {
                     RawStatement::Sequence(
                         box Statement {
-                            content: RawStatement::Switch(switch),
+                            content: RawStatement::Switch(switch @ Switch::SwitchInt(..)),
                             meta: meta2,
                         },
                         box st3,
                     ) => (meta2, switch, Some(st3)),
-                    RawStatement::Switch(switch) => (st2.meta, switch, None),
-                    _ => unreachable!("A discriminant read must be followed by a `Switch`"),
+                    RawStatement::Switch(switch @ Switch::SwitchInt(..)) => {
+                        (st2.meta, switch, None)
+                    }
+                    _ => {
+                        register_error_or_panic!(
+                            self.ctx,
+                            st.meta.span.rust_span_data.span(),
+                            "A discriminant read must be followed by a `SwitchInt`"
+                        );
+                        // An error occurred. We can't keep the `Rvalue::Discriminant` around so we
+                        // `Nop` the whole statement sequence.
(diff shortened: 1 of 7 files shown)


charon/src/types.rs
```diff
@@ -371,6 +371,9 @@ pub struct Variant {
     pub meta: Meta,
     pub name: String,
     pub fields: FieldId::Vector<Field>,
+    #[serde(skip)]
+    /// The discriminant used at runtime.
```
sonmarcho, line 375: Can you add more information in this comment, to explain that we only need this in remove_read_discriminant to properly match the integer discriminant values to the variant ids when reconstructing the matches?

charon/src/remove_read_discriminant.rs
```diff
@@ -48,67 +48,74 @@ impl<'a, 'tcx, 'ctx> Visitor<'a, 'tcx, 'ctx> {
...
+                        (
+                            v.into_iter()
+                                .map(|x| {
+                                    let discr = x.to_bits();
+                                    covered_discriminants.insert(discr);
+                                    discr_to_id.get(&discr).unwrap_or_else(|| {
+                                        panic!(
+                                            "Found incorrect discriminant {discr} for enum {}",
+                                            adt_id
+                                        )
```
sonmarcho, line 108: Can you use a register_error_or_panic so that we can recover from the error and print a message to the user?

charon/src/remove_read_discriminant.rs
```diff
@@ -48,67 +48,74 @@ impl<'a, 'tcx, 'ctx> Visitor<'a, 'tcx, 'ctx> {
                         box st3,
                     ) => (meta2, switch, Some(st3)),
                     RawStatement::Switch(switch) => (st2.meta, switch, None),
-                    _ => unreachable!(),
+                    _ => unreachable!("A discriminant read must be followed by a `Switch`"),
```
sonmarcho, line 51: I hadn't noticed this one: can you make this error not a hard error?

charon/src/remove_read_discriminant.rs
```diff
@@ -48,67 +48,74 @@ impl<'a, 'tcx, 'ctx> Visitor<'a, 'tcx, 'ctx> {
...
+                            TypeDeclKind::Error(_) => None,
+                            TypeDeclKind::Enum(variants) => Some(variants),
                         }
                     }
                 };
+                let Some(variants) = variants else {
+                    // An error occurred. We can't keep the `Rvalue::Discriminant` around so we
+                    // `Nop` the whole statement sequence.
+                    assert!(self.ctx.error_count > 0);
+                    *st = Statement {
```
sonmarcho, line 82: I think we should add an `Error` case in the statements. Could you put a TODO here and open an issue?

sonmarcho (changes requested):

sonmarcho (approved):

PR #109, opened 2024-04-09
Run `charon-driver` in tests

Nadrieril:
We avoid invoking the cargo machinery when not needed. `tests-polonius` still need `cargo` because they depend on `serde`.

### DIFF
--- a/charon/src/charon-driver.rs
+++ b/charon/src/charon-driver.rs
@@ -89,6 +89,7 @@ mod values_utils;
 use crate::driver::{
     arg_value, get_args_crate_index, get_args_source_index, CharonCallbacks, CharonFailure,
 };
+use crate::export::CrateData;
 
 fn main() {
     // Initialize the logger
@@ -220,12 +221,20 @@ fn main() {
     let mut callback = CharonCallbacks::new(options);
     let mut res = callback.run_compiler(compiler_args);
     if let Some(crate_data) = &callback.crate_data {
+        // `crate_data` is `None` on the first call of the driver.
         if !callback.options.no_serialize {
             // # Final step: generate the files.
-            // `crate_data` is `None` on the first call of the driver.
             res = res.and_then(|()| {
+                let dest_file = callback.options.dest_file.clone().unwrap_or_else(|| {
+                    let (crate_name, extension) = match crate_data {
+                        CrateData::ULLBC(d) => (&d.name, "ullbc"),
+                        CrateData::LLBC(d) => (&d.name, "llbc"),
+                    };
+                    format!("{crate_name}.{extension}").into()
+                });
+                trace!("Target file: {:?}", dest_file);
                 crate_data
-                    .serialize_to_file(&callback.options.dest_dir)
+                    .serialize_to_file(&dest_file)
                     .map_err(|()| CharonFailure::Serialize)
             });
         }
--- a/charon/src/charon-driver.rs
+++ b/charon/src/charon-driver.rs
@@ -212,16 +212,14 @@ fn main() {
     // on Cargo, the charon-driver is only called on the target (and not its
     // dependencies).
     //
-    // Note that the first call to the driver is with "--crate-name ___" and no
-    // source file, for Cargo to retrieve some information about the crate.
-    // We don't need to check this case in order to use the default Rustc callbacks
-    // instead of the Charon callback: because there is nothing to build, Rustc will
-    // take care of everything and actually not call us back.
+    // Cargo calls the driver twice. The first call to the driver is with "--crate-name ___" and no
+    // source file, for Cargo to retrieve some information about the crate. The compiler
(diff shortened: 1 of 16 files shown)


charon/src/charon-driver.rs
```diff
@@ -210,12 +221,20 @@ fn main() {
     let mut callback = CharonCallbacks::new(options);
     let mut res = callback.run_compiler(compiler_args);
     if let Some(crate_data) = &callback.crate_data {
+        // `crate_data` is `None` on the first call of the driver.
```
sonmarcho, line 224:
```suggestion
        // `crate_data` is initialized to `None`, then updated to `Some(...)` after running the compiler
```
I think this is more accurate?

flake.nix
```diff
@@ -62,7 +62,7 @@
             inherit cargoArtifacts;
             buildPhase = ''
               # Run the tests for Charon
-              DEST=$out CHARON="${charon}/bin/charon --cargo-no-rust-version" \
+              DEST=$out CARGO_NO_RUST_VERSION=1 CHARON_DRIVER="${charon}/bin/charon-driver" \
```
sonmarcho, line 65: Do we agree that with your changes `charon` runs cargo while `charon-driver`doesn't (and in particular `charon` is still valid)?

charon/src/cli_options.rs
```diff
@@ -11,51 +11,50 @@ use std::path::PathBuf;
...
     /// extract part of a crate for instance).
-    #[structopt(long = "input", parse(from_os_str))]
+    #[clap(long = "input", parse(from_os_str))]
     #[serde(default)]
     pub input_file: Option<PathBuf>,
-    /// The destination directory, if we don't want to generate the output
-    /// .llbc files in the same directory as the input .rs files.
-    #[structopt(long = "dest", parse(from_os_str))]
+    /// The destination file. By default `<crate_name>.llbc`.
+    #[clap(long = "dest-file", parse(from_os_str))]
```
sonmarcho, line 49: I would like to keep a `--dest` or `--dest-dir`option: I think it convenient to be able to only specify the output directory (I often do it). You can have two options and check that at most one is set.

tests/Makefile
```diff
@@ -72,17 +73,17 @@ clean:
...
+#	$(CHARON_CMD) --dest-file $(DEST)/llbc_prom/$*.llbc --mir_promoted
 # TODO: this fails for now (there is some very low-level desugaring happening)
-#	$(CHARON_CMD) --dest $(DEST)/llbc_opt --mir_optimized
+#	$(CHARON_CMD) --dest-file $(DEST)/llbc_opt/$*.llbc --mir_optimized
 
 else
 
 test-%:
-	$(CHARON_CMD) --dest $(DEST)/llbc
+	$(CHARON_CMD) --dest-file $(DEST)/llbc/$*.llbc
```
sonmarcho, line 87: If we keep a `--dest`/`--dest-dir` option then there is no need to modify those rules

sonmarcho (changes requested):

charon/src/charon-driver.rs
Nadrieril: That's not what I meant, I'll clarify

flake.nix
Nadrieril: Yes, `charon` is unchanged. We could probably rename it to `cargo-charon` tbh

Nadrieril:
I force pushed because I rebased on `main` but I didn't change any of the original commits. The last two commits are new

charon/src/cli_options.rs
```diff
@@ -11,51 +11,56 @@ use std::path::PathBuf;
...
     /// This is needed if you want to define a custom entry point (to only
     /// extract part of a crate for instance).
-    #[structopt(long = "input", parse(from_os_str))]
+    #[clap(long = "input", parse(from_os_str))]
     #[serde(default)]
     pub input_file: Option<PathBuf>,
-    /// The destination directory, if we don't want to generate the output
-    /// .llbc files in the same directory as the input .rs files.
-    #[structopt(long = "dest", parse(from_os_str))]
+    /// The destination directory. Files fill be generated as `<dest_dir>/<crate_name>.llbc`,
```
sonmarcho, line 48: Update to `...<crate_name>.{u}llbc`

charon/src/cli_options.rs
```diff
@@ -11,51 +11,56 @@ use std::path::PathBuf;
...
+    /// unless `dest_file` is set. `dest_dir` defaults to the current directory.
+    #[clap(long = "dest", parse(from_os_str))]
     #[serde(default)]
     pub dest_dir: Option<PathBuf>,
+    /// The destination file. By default `<dest_dir>/<crate_name>.llbc`. If this is set we gnore
```
sonmarcho, line 53: "ignore"

sonmarcho (approved):
This looks good. I have two minor comments: you can merge directly once you have addressed them.

PR #110, opened 2024-04-10
Clean up the indexing macro

Nadrieril:
This replaces most of it with usageof the `index_vec` crate, extracts what remains into a macro-by-example and a new module, and finally cleans up a bit to avoid generators when not needed.


charon/src/ids/generator.rs
```diff
@@ -0,0 +1,59 @@
+use index_vec::Idx;
+use std::{collections::HashMap, hash::Hash, marker::PhantomData};
+
+#[derive(Debug, Clone, Copy)]
+pub struct Generator<I: Idx> {
+    counter: usize,
+    phantom: PhantomData<I>,
```
sonmarcho, line 7: Is this the standard way of doing (rather than storing an `I` and converting it to a `usize` when needs be)?

sonmarcho (changes requested):
Looks good. Could you locally test Aeneas (in particular, that there are no diffs in the generated code). Also, could you add a check in Charon that the LLBC for the `closures.rs` test is stable?

charon/src/ids/generator.rs
Nadrieril: I didn't think very hard, I think this is fine

PR #113, opened 2024-04-11
ci: don't run the same job twice

Nadrieril:
I noticed that hen making a PR the same CI job is run twice. This is an experiment to avoid that. It uses https://github.com/marketplace/actions/skip-duplicate-actions. I'm not sure how to test that this work apart from merging and finding out.


Nadrieril:
It's a bit noisy but seems to work

Nadrieril:
My understanding of how this works is that `skip_duplicate_job` queries the github API to see if there's another identical job. I don't know how it handles concurrency, but `same_content_newer` ensures we run at least one version of the job (the oldest).

> As a side note: only one of the nix jobs was skipped.

Well yeah, we want to run one of them.

> Also, could we rename the `skip_duplicate_job`?

What name do you want?

pnmadelaine:
Yeah, I've been frustrated with the duplication of jobs, but I'm not really found of this solution either as it is indeed quite noisy.
Since we use Nix, one of the job is basically going to be a noop anyway, so I don't think the benefit of skipping one job is worth the clutter…

sonmarcho:
> My understanding of how this works is that `skip_duplicate_job` queries the github API to see if there's another identical job. I don't know how it handles concurrency, but `same_content_newer` ensures we run at least one version of the job (the oldest).
> 
> > As a side note: only one of the nix jobs was skipped.
> 
> Well yeah, we want to run one of them.
I meant: you successfully skipped one `nix` job, but there are two `skip_duplicate_job`. Is it expected?

> > Also, could we rename the `skip_duplicate_job`?
> 
> What name do you want?

Something like `check_if_skip_duplicate_job` (this is what it does, right?)

sonmarcho:
> Yeah, I've been frustrated with the duplication of jobs, but I'm not really found of this solution either as it is indeed quite noisy.
> Since we use Nix, one of the job is basically going to be a noop anyway, so I don't think the benefit of skipping one job is worth the clutter…

Good remark. But it would be useful to have it on the side of Aeneas, as we can't nixify the Lean tests and they are already expensive (~10 minutes).

pnmadelaine:
> Good remark. But it would be useful to have it on the side of Aeneas, as we can't nixify the Lean tests and they are already expensive (~10 minutes).

Definitely! Also, thanks @Nadrieril for bringing this up, I did not know about this github action :) (which, frankly, should not even *have* to exist)

Nadrieril:
> Since we use Nix, one of the jobs is basically going to be a noop anyway

Oh, I hadn't noticed but you're right the second job takes a few seconds only. There's no point adding two extra jobs just to skip 10s of computation. Let's do it only on the aeneas side then (unless we manage to nixify the Lean tests ? :angel:)

> frankly, should not even have to exist

I see why it does though: github actions can have side-effects e.g. call to an API. Makes perfect sense to run the job twice for distinct events.

pnmadelaine:
> unless we manage to nixify the Lean tests ? 👼

👀

> I see why it does though: github actions can have side-effects e.g. call to an API. Makes perfect sense to run the job twice for distinct events.

Well, imho it would have been more reasonable to have two events triggered at the exact same time be somehow merged into one…

Nadrieril:
They're not at the same time: the first one is triggered when I push the branch, the second when I make the pull request. Also, the job is allowed to check what triggered it so it can distinguish push from pull_request.

Btw one way to avoid duplicate tests I think is to make PRs from a fork instead of from this repo. Another is to disable check on push except on `main`. I'm up for that unless one of you relies on the `push` CI outside of making PRs

pnmadelaine:
> They're not at the same time: the first one is triggered when I push the branch, the second when I make the pull request. Also, the job is allowed to check what triggered it so it can distinguish push from pull_request.

When you push to an open PR, you do get the two events at once. (or am I mistaken?)

> Another is to disable check on push except on `main`. I'm up for that unless one of you relies on the `push` CI outside of making PRs.

I personally like having CI running on all branches of a project, when the workload is not too big, but that is a good point. @sonmarcho what do you think?

Nadrieril:
> When you push to an open PR, you do get the two events at once. (or am I mistaken?)

Oh, when the PR is already open indeed, good point

Nadrieril:
Actually, this is useful even with nix: when the tests *fail* then we run them twice

Nadrieril:
Oh wow, turns out the `pull_request` test doesn't run on the head of the branch, it runs on a merge commit between `main` and the branch! This is good, we should really not skip that.

sonmarcho:
> Actually, this is useful even with nix: when the tests _fail_ today, we run them twice.

This is a good reason of having this actually.

sonmarcho (approved):

Issue #114, opened 2024-04-11
Add an option to extract the bodies of the external definitions


sonmarcho:
For now, whenever possible we treat foreign definitions as opaque, that is we treat all external functions as opaque and thus don't look at their bodies, treat external structures without public fields as opaque, etc. It is sometimes useful to extract the bodies of the external definitions, and Eurydice needs this: we need to add a CLI option for this. Also, it might be good to be able to tweak this behavior, for instance by extracting only the bodies of the functions marked as `#[inline]`, etc.

Nadrieril:
Ok, so foreign functions/methods, structs that have private fields, anything else?

Nadrieril:
This is harder than expected, because the only MIR we can access for foreign functions is optimized MIR which we don't support well

Issue #120, opened 2024-04-12
Support `Rvalue::Discriminant` not followed by `SwitchInt`


Nadrieril:
This can show up in optimized MIR, for example the optimized mir of `Option::is_some()`. It's not too hard to support: we can replace `_x = Discriminant(_y)` with
```rust
match _y {
    Variant1 => _x = <discriminant value for variant 1>,
    ...
}
```

PR #121, opened 2024-04-12
Add option to extract opaque bodies

Nadrieril:
With `--extract-opaque-bodies`, we now also extract foreign globals, methods, trait provided methods, and ADT definitions. This is limited by two things:
- not all function bodies have a MIR in the crate data (only inlineable and generic functions do);
- when function have MIR available it's only optimized MIR, which we don't support as well.

This PR includes a few tests that fail because of other problems. Two of them are hax limitations that I'm ignoring until we're back on hax master. I opened issues for the other two: #123, #120.

The PR also includes many improvements to the ui tests that I needed along the way.

Fixes https://github.com/AeneasVerif/charon/issues/114


charon/src/translate_traits.rs
```diff
@@ -304,7 +304,7 @@ impl<'tcx, 'ctx> TransCtx<'tcx, 'ctx> {
                     // but still remember their name.
```
sonmarcho, line 304: Update the comment to mention the CLI option which changes this behavior?

charon/src/translate_functions_to_ullbc.rs
```diff
@@ -1887,14 +1883,12 @@ impl<'tcx, 'ctx> TransCtx<'tcx, 'ctx> {
...
 
-        let body = if rust_id.is_local() && is_transparent {
-            // It's a local and transparent global: we extract its body as for functions.
-            match bt_ctx.translate_body(rust_id.expect_local(), 0) {
-                Err(_) => {
-                    // Error case: we could have a specific variant
-                    None
-                }
-                Ok(body) => Some(body),
+        let body = if is_transparent {
```
sonmarcho, line 1886: Maybe add a comment here to explain that we test whether the definition is local or not in `translate_body`(I was quite puzzled when reviewing this change at first)

charon/src/translate_functions_to_ullbc.rs
```diff
@@ -1429,16 +1429,6 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
...
         for arg in args {
-            // There should only be moved arguments, or constants
-            match arg {
-                hax::Operand::Move(_) | hax::Operand::Constant(_) => {
-                    // OK
-                }
-                hax::Operand::Copy(_) => {
-                    unreachable!();
-                }
-            }
```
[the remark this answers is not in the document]
Nadrieril: Yep, in one of the std functions exported by one of the tests, I didn't find out which one exactly

charon/tests/ui/issue-114-opaque-bodies-aux.rs
```diff
@@ -0,0 +1,21 @@
+//@ skip
```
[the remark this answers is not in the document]
Nadrieril: Yep

charon/src/translate_functions_to_ullbc.rs
Nadrieril: We could do the `is_transparent` check in `translate_body` instead?

Nadrieril:
We could pass `-Zalways-encode-mir` to dependencies, I think this ensures we get the MIR for everything (while we're at it, `-Zno-codegen` avoids generating code we don't need).

This doesn't work for `std` though; for that, this proposal might be the solution, as it proposes shipping a std with full MIR: https://github.com/rust-lang/compiler-team/issues/738. Or we could ship our own MIR-only sysroot.

charon/src/translate_functions_to_ullbc.rs
Nadrieril: I did that, lmk if you like it

zhassan-aws:
> The fact that we can't always retrieve the bodies of the external functions makes the effects of this PR quite limited, unfortunately. I wonder how the Kani people do, as I thought Kani looked into the bodies of the external definitions (@zhassan-aws could you comment on that?).

In Kani, we run the `kani-compiler` on each of the dependent crates to generate the MIR only. Then, when compiling the target user crate, the `kani-compiler` has access to the MIR of all the dependent crates. This also applies to `std`, but it's handled in a special way.

See [this RFC](https://model-checking.github.io/kani/rfc/rfcs/0001-mir-linker.html) and [one of the PRs that implements it](https://github.com/model-checking/kani/pull/1683) for more information.

Feel free to reach out for further information.

Nadrieril:
Thank you for these links!

sonmarcho:
This is useful documentation, thanks!

sonmarcho (approved):
Looks good: it indeed makes sense to do check once and for all in `translate_body` whether we should extract the body or not.

PR #124, opened 2024-04-15
Fix cleanup of overflow checks

Nadrieril:
#87 revealed a case where we don't correctly clean up overflow checks: in some cases some code is inserted between the overflow check and the division, so we don't catch it. Turns out, `TerminatorKind::Assert` is only used for builtin overflow/div-by-zero/etc checks, and these checks are always generated in the same pattern, so we don't need to check that there is a division after the assert.

There's still a case that isn't detected: https://github.com/AeneasVerif/charon/issues/125. That'll take a bit more effort.

Fixes https://github.com/AeneasVerif/charon/pull/87


sonmarcho (changes requested):
This looks good. I just have two requests.
1. Can you check that the output of Aeneas is unchanged?
2. In `tests/src/no_nested_borrows.rs` there are a few micro tests which contain binary operations and are there only to test the removal of the dynamic checks. Can you move them to `...-binop-simplification.rs`?

sonmarcho (changes requested):
This looks good. I just have two requests.
1. Can you check that the output of Aeneas is unchanged?
2. In `tests/src/no_nested_borrows.rs` there are a few micro tests which contain binary operations and are there only to test the removal of the dynamic checks. Can you move them to `...-binop-simplification.rs`?

Nadrieril:
I checked Aeneas, all good. Nice set of micro-tests, that's a welcome addition.

tests/src/no_nested_borrows.rs
```diff
@@ -43,88 +43,6 @@ pub enum Sum<T1, T2> {
...
-    ((x + y) * (x / y) + (x - (z % y))) % (x + y + z)
-}
-
-/* Checking the simplification of binop operations *inside* global constants.
-
-   In release mode, the Rust compiler inserts additional checks inside constant
-   bodies.
-*/
-pub const CONST0: usize = 1 + 1;
-pub const CONST1: usize = 2 * 2;
```
sonmarcho, line 126: Looks good. Can you create a PR on the side of Aeneas to update the generated test files?

Nadrieril: I'll do it once this is merged

sonmarcho (approved):

Issue #125, opened 2024-04-15
Detect overflow checks properly


Nadrieril:
After https://github.com/AeneasVerif/charon/pull/124, there's still a case where we don't detect overflow checks properly:
```rust
fn div_signed_with_constant() -> i32 {
    const FOO: i32 = 42;
    FOO / 2
}
```
gives:
```
fn test_crate::div_signed_with_constant() -> i32
{
    let @0: i32; // return
    let @1: bool; // anonymous local
    let @2: bool; // anonymous local
    let @3: i32; // anonymous local
    let @4: i32; // anonymous local

    @1 := const (2 : i32) == const (-1 : i32)
    @3 := test_crate::div_signed_with_constant::FOO
    @2 := move (@3) == const (-2147483648 : i32)
    @4 := test_crate::div_signed_with_constant::FOO
    @0 := move (@4) / const (2 : i32)
    return
}
```
because the constants are inserted in the middle of the checks.

sonmarcho:
Actually, there is a way to fix that I think: we simply need to swap the order of the micro-pass which simplify the constants (by inserting intermediate let-bindings) and the micro pass which removes the dynamic checks. One issue is that one of the two operates on ULLBC and the other on LLBC... But it should be easy to update one of the two to work on the proper AST (I guess the constant simplification can be easily moved).

Nadrieril:
Oh! I assumed it was rustc's fault. Yeah sounds good

PR #129, opened 2024-04-17
Rework the removal of dynamic checks

Nadrieril:
To avoid interference with other passes, we do this one earlier on ullbc. This happens to also be a lot nicer to write.

Because `TerminatorKind::Assert` is only used for dynamic checks (listed [here](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.AssertKind.html)), this pass in fact ensures we remove all `Assert`s from ullbc.

Fixes https://github.com/AeneasVerif/charon/issues/125

Based on https://github.com/AeneasVerif/charon/pull/124


Nadrieril:
This is a bit dirty, I've got a better solution in mind

charon/src/driver.rs
```diff
@@ -230,6 +230,13 @@ pub fn translate(
     // we simply apply some micro-passes to make the code cleaner, before
     // serializing the result.
 
+    // # Micro-pass: Remove overflow/div-by-zero/bounds checks since they are already part of the
+    // arithmetic/array operation in the semantics of (U)LLBC.
+    // **WARNING**: this pass uses the fact that the dynamic checks introduced by Rustc use a
+    // special "assert" construct. Because of this, it must happen *before* the
+    // [reconstruct_asserts] pass. See the comments in [crate::remove_dynamic_checks].
```
sonmarcho, line 237: Also explain that the pass must happen before `simplify_constants` because the latter introduces auxiliary let-bindings which mess up with the patterns we're looking for here.

charon/src/expressions.rs
```diff
@@ -141,6 +141,14 @@ pub enum BinOp {
     Sub,
     /// Can overflow
     Mul,
+    /// Returns `(result, did_overflow)`, where `result` is the result of the operation with
+    /// wrapping semantics, and `did_overflow` is a boolean that indicates whether the operation
+    /// overflowed.
```
sonmarcho, line 146: "Note that on its side the unchecked [Add] operation panics in case of under/overflow."

charon/tests/ui/remove-dynamic-checks.out
```diff
@@ -326,17 +339,55 @@ fn test_crate::mix_arith_i32(@1: i32, @2: i32, @3: i32) -> i32
...
+    return
+}
+
 global test_crate::CONST0  {
     let @0: usize; // return
+    let @1: (usize, bool); // anonymous local
 
-    @0 := const (1 : usize) + const (1 : usize)
+    @1 := const (1 : usize) checked.+ const (1 : usize)
+    assert(move ((@1).1) == false)
```
sonmarcho, line 379: As discussed, we won't touch this for now. But could you open an issue to simplify the constants?

charon/src/transform/remove_dynamic_checks.rs
```diff
@@ -1,346 +1,154 @@
...
+            content: RawStatement::Assign(x, _),
+            ..
+        }, Statement {
+            content:
+                RawStatement::Assign(
+                    has_overflow,
+                    Rvalue::BinaryOp(BinOp::Lt, Operand::Copy(lt_op2), Operand::Const(..)),
+                ),
+            ..
+        }] if lt_op2 == x && cond == has_overflow && *expected == true => rest,
```
sonmarcho, line 87: There is nothing for right shifts?

charon/src/transform/remove_dynamic_checks.rs
```diff
@@ -1,346 +1,154 @@
...
 
-    fn merge(&mut self) {}
+        // Overflow checks for addition/subtraction/multiplication. They look like:
+        //   r := x checked.+ y;
+        //   assert(move r.1 == false);
+        // They only happen in constants because we compile with `-C opt-level=3`. They're tricky
+        // to remove so we leave them.
```
sonmarcho, line 93:
```suggestion
        // to remove so we leave them for now.
```

sonmarcho (approved):
This is a lot cleaner, thanks!
Can you:
- open an issue for the constants case?
- locally check that the code generated by Aeneas is unchanged?
After this, you're free to merge.

charon/src/transform/remove_dynamic_checks.rs
Nadrieril: Nope, they can't overflow

Nadrieril: And no one cares whether they underflow apparently

charon/tests/ui/remove-dynamic-checks.out
Nadrieril: Do you want to simplify constants eventually?

sonmarcho: Not necessarily, but it's good if it is documented.

charon/src/transform/remove_dynamic_checks.rs
Nadrieril: I have to check that more thoroughly though, rust documentation isn't as clear as I'd like

charon/tests/ui/remove-dynamic-checks.out
Nadrieril: I think issues are not appropriate for documentation. Should we have a markdown file somewhere about supported/unsupported features?

Nadrieril: Done: https://github.com/AeneasVerif/charon/issues/142

charon/src/transform/remove_dynamic_checks.rs
Nadrieril: I was wrong: somehow I thought I had seen shl overflow checks and no shr overflow checks, but rust emits checks for both when not in release mode, and it's not about overflow. Since we use release mode we don't see them at all.

Nadrieril:
I opened https://github.com/AeneasVerif/charon/issues/142 and checked aeneas locally.

Nadrieril:
This may be breaking Eurydice but I haven't had time to investigate

sonmarcho:
You can merge when you want.
@protz: this may break Eurydice.

Nadrieril:
This also breaks `const MONTGOMERY_R: i32 = 1 << MONTGOMERY_SHIFT;`. I'll leave this as draft for now

Nadrieril:
Alright, this is blocked until I can figure out how to implement the new `Checked*` binops for all backends

Issue #91, opened 2024-03-22
Support enum-to-discriminant casting


zhassan-aws:
For the following program:
```rust
enum Foo {
    A,
    B,
}

fn main() {
    let x = Foo::A;
    let _y = x as isize;
}
```
Charon errors out:
```
$ charon
   Compiling discr v0.1.0 (/home/ubuntu/examples/aeneas/discr)
error: Unsupported statement kind: intrinsic
 --> src/main.rs:8:14
  |
8 |     let _y = x as isize;
  |              ^^^^^^^^^^

[ INFO charon_driver::export:101] [gexport]: Generated the partial (because we encountered errors) file: /home/ubuntu/examples/aeneas/discr/discr.llbc
[ ERROR charon_driver:231] The extraction encountered 1 errors
error: could not compile `discr` (bin "discr") due to previous error
```

PR #132, opened 2024-04-18
Allow discriminant reads not followed by `SwitchInt`

### LINKED ISSUE #120: Support `Rvalue::Discriminant` not followed by `SwitchInt`

Nadrieril:
We replace `_x = Discr(_y);` not followed by a `SwitchInt` with
```rust
match _y {
    0 => {
        _x := const (0 : isize)
    },
    1 => {
        _x := const (1 : isize)
    }
    ...
}
```

Fixes #120.


Nadrieril:
We can use this to make https://github.com/AeneasVerif/charon/issues/91 work too, as soon as hax forwards the right information to us.

charon/tests/ui/issue-91-enum-to-discriminant-cast.out
```diff
@@ -0,0 +1,58 @@
...
+
+fn test_crate::main()
+
+fn core::clone::Clone::clone<'_0, Self>(@1: &'_0 (Self)) -> Self
+
+
+
+error: aborting due to previous error
+
+[ ERROR charon_driver:264] The extraction encountered 1 errors
```
sonmarcho, line 58: Why does this one fail? I thought this PR would fix it.

charon/src/values_utils.rs
```diff
@@ -218,6 +218,9 @@ impl ScalarValue {
             ScalarValue::I128(v) => v as u128,
         }
     }
+    pub fn from_bits(ty: IntegerTy, bits: u128) -> Self {
```
sonmarcho, line 221: Nit: could you add a space above?

charon/src/remove_read_discriminant.rs
```diff
@@ -102,53 +69,106 @@ impl<'a, 'tcx, 'ctx> Visitor<'a, 'tcx, 'ctx> {
...
+                            };
+                            new_sequence(switch, st3).content
+                        } else {
+                            switch
+                        };
+                    }
+                    Err(st2) => {
+                        // The discriminant read is not followed by a `SwitchInt`. We replace `_x =
+                        // Discr(_y)` with `match _y { 0 => { _x = 0 }, 1 => { _x = 1; }, .. }`.
+                        let targets = variants
```
sonmarcho, line 145: Just add a comment which states that it happens when we do, for instance: `y as usize`.

sonmarcho (changes requested):

charon/tests/ui/issue-91-enum-to-discriminant-cast.out
Nadrieril: `as` casts use a different MIR statement, and hax currently doesn't support it

charon/src/remove_read_discriminant.rs
Nadrieril: That's not the case. As far as I know only optimizing a `match` can make that happen. I'll note that

charon/src/values_utils.rs
Nadrieril: FTR I like grouping together functions that belong together

sonmarcho: I like having spaces, I think it's cleaner :)

charon/src/remove_read_discriminant.rs
sonmarcho: Ok, then just add a comment which explains when this can happen :)

sonmarcho (approved):

sonmarcho:
You're free to mark the conversation as resolved and merge when you consider they are.

PR #137, opened 2024-04-19
Fix linking on Mac

Nadrieril:
I accidentally set a linux-only toolchain in #133. This should fix `make test` on Mac


sonmarcho (commented):
Looks good to me. @R1kM I'll let you approve this and merge it if it works on your side.

msprotz:
Works for me

msprotz (approved):

PR #138, opened 2024-04-22
Fix the linking situation

Nadrieril:
We have three sets of users to support:
- Those who build the nix flake;
- Those who build the project using rustup;
- Those who build the project by hand using a nix-provided toolchain.

We should now correctly support all three, with error messages when we're in an unexpected situation. The setup is now as follows: `charon-driver` is never called directly; instead, call `charon --no-cargo`. In general, `charon` is the entrypoint that makes sure we have the right toolchain around.

### DIFF
--- a/charon/Cargo.lock
+++ b/charon/Cargo.lock
@@ -197,6 +197,7 @@ dependencies = [
  "tempfile",
  "toml",
  "walkdir",
+ "which",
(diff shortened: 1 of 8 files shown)


Nadrieril:
Yes, that's why I left it as draft

PR #139, opened 2024-04-22
experiment: Use github-hosted runner

Nadrieril:


Nadrieril:
Good to know: self-hosted runner takes ~2min30 to rebuild charon, github-hosted runner takes ~5min, in large part because it's not as good with deduplicating nix things.

PR #140, opened 2024-04-22
nix: Improve charon building

Nadrieril:
This removes an old workaround, and ensures we don't include irrelevant inputs when building charon via nix.


sonmarcho (approved):
Thanks for cleaning this mess (though I don't understand how you managed to make it work by doing the "naive thing" - but I don't understand much to nix anyway).

PR #141, opened 2024-04-22
Support default systems (including aarch64-darwin) in flake.nix

zgrannan:
These changes enable MacOs (and potentially other platforms) to build Charon via nix (I need to use this workflow because my Ocaml installation is horribly broken somehow). 

I tested this on MacOS and it seems to work, although the UI tests fail because rustc generates different `DefId`s, it seems. I haven't tested this on x86-64 linux however.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,7 @@
   };
 
   outputs = { self, flake-utils, nixpkgs, rust-overlay, crane, flake-compat }:
-    flake-utils.lib.eachSystem [ "x86_64-linux" ] (system:
+    flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs {
           inherit system;
@@ -40,6 +40,7 @@
         };
 
         charon = craneLib.buildPackage (craneArgs // {
+          buildInputs = [ pkgs.zlib ];
           # It's important to pass the same `RUSTFLAGS` to dependencies otherwise we'll have to rebuild them.
           cargoArtifacts = craneLib.buildDepsOnly craneArgs;
           # Check the `ui_llbc` files are correct instead of overwriting them.


flake.nix
```diff
@@ -55,9 +55,14 @@
               # Don't include the `target` directory in the nix inputs.
               filter = path: type: !pkgs.lib.hasPrefix (toString ./target) path;
             };
+            # If the zlib package is not present on `aarch64-darwin`, a linker error
+            # will occur
+            buildInputs = if system == "aarch64-darwin" then [ pkgs.zlib ] else [ ];
```
Nadrieril, line 60: It probably doesn't hurt to add zlib as input for all platforms

flake.nix
```diff
@@ -55,9 +55,14 @@
...
             inherit cargoArtifacts;
             # Check the `ui_llbc` files are correct instead of overwriting them.
             cargoTestCommand = "IN_CI=1 cargo test --profile release";
+            # UI tests seem to fail on darwin, seemingly because rustc generates different DefIDs
+            doCheck = system != "aarch64-darwin";
```
Nadrieril, line 65: Do we print defids in the output? That's not great, I'd rather change that because they're not supposed to be stable. Can you share test outputs?

zgrannan: Here's the relevant snippet:
```
test unsupported/option-is_some.rs                 ... FAILED
test issue-4-traits.rs                             ... ok
test unsupported/result-unwrap.rs                  ... ok
test issue-114-opaque-bodies.rs                    ... ok
test unsupported/issue-79-bound-regions.rs         ... ok
test binop-simplifications.rs                      ... ok
test unsupported/slice_index.rs                    ... ok
test closures.rs                                   ... ok
test issue-45-misc.rs                              ... ok

failures:

---- unsupported/option-is_some.rs ----

---- expected: tests/ui/unsupported/option-is_some.out
++++ actual:   charon output
   1    1 | error: A discriminant read must be followed by a `SwitchInt`
   2    2 |  --> /rustc/d59363ad0b6391b7fc5bbb02c9ccf9300eef3753/library/core/src/option.rs:598:5
   3    3 |
   4    4 | [ INFO charon_lib::driver:332] [translate]: # Final LLBC before serialization:
   5    5 |
          ⋮
  29   29 |     return
  30   30 | }
  31   31 |
  32   32 |
  33   33 |
  34      - error: The external definition DefId(2:7981 ~ core[6c80]::option::{impl#0}::is_some) triggered errors. It is (transitively) used at the following location(s):
       34 + error: The external definition DefId(2:7988 ~ core[02ea]::option::{impl#0}::is_some) triggered errors. It is (transitively) used at the following location(s):
  35   35 |  --> tests/ui/unsupported/option-is_some.rs:6:5
  36   36 |   |
  37   37 | 6 |     opt.is_some()
  38   38 |   |     ^^^^^^^^^^^^^
  39   39 |
  40   40 | error: aborting due to 2 previous errors
  41   41 |
  42   42 | [ ERROR charon_driver:180] The extraction encountered 1 errors
  ```

flake.nix
zgrannan: Updated the PR

flake.nix
sonmarcho: When there is an error, it indeed happens that I directly print the `DefId` of the faulty definition. @Nadrieril do you know a better way?

Nadrieril: There's probably a way to print the path without the ids, I don't think we can do anything with these ids anyway

Nadrieril: I've submitted https://github.com/AeneasVerif/charon/pull/143 that does that.

Nadrieril:
@zgrannan if you rebase on main the tests should pass now

zgrannan:
Seems to work :)

Nadrieril (approved):

sonmarcho (approved):

PR #150, opened 2024-04-25
Fix linking under nix on Mac (again)

Nadrieril:
Turns out cargo doesn't setup dynamic links the same way under mac and linux so we have to override by hand on mac.


sonmarcho (approved):
Gosh, this is so annoying. Thanks for looking into this!

PR #155, opened 2024-04-29
CI: build Eurydice & Kyber

pnmadelaine:


Nadrieril:
Ah. I think the "check if duplicate job" job thinks this is a redundant job. Somehow it didn't notice that the ci yamls changed :sweat:

pnmadelaine:
> @Nadrieril any idea why all the jobs are skipped?

It seems the skip logic checks for the content of the repo and not just the commit sha, this can be problematic for jobs like these one that have effects. Can this be set up?

Anyway, for an example of the workflows created by this PR, see https://github.com/AeneasVerif/charon/actions/runs/8880663304

pnmadelaine:
> for jobs like these ones that have effects

Well they don't exactly have effects, but they are dependant on the state of outside dependencies, and we could definitely want to re-run them against a new version of these dependencies!

Nadrieril:
Oh I'm silly, it's just that you ran these tests on your own branch earlier. All good then.

I don't particularly care to re-run them if the dependencies change; this is a best-effort check to make sure changes to charon don't break the dependencies. For a more thorough whole-toolchain check, I think @R1kM has plans.

.github/workflows/ci.yml
```diff
@@ -33,3 +33,38 @@ jobs:
...
+    steps:
+      - name: checkout libcrux
+        uses: actions/checkout@v4
+        with:
+          repository: cryspen/libcrux
+      - name: checkout eurydice
+        uses: actions/checkout@v4
+        with:
+          repository: aeneasverif/eurydice
+          path: eurydice
```
Nadrieril, line 57: Could you add a comment explaining that `kyber` expects a checkout of eurydice and not just the binary?

Nadrieril (approved):

.github/workflows/ci.yml
pnmadelaine: done!

R1kM (approved):

PR #156, opened 2024-04-29
Fix mac linking properly

Nadrieril:
In https://github.com/AeneasVerif/charon/pull/150 I added a quick workaround for building charon with nix on mac. This is a cleaner fix. I tested it with the GitHub CI runners.

@R1kM could you try running the rustc tests with this branch?


R1kM:
This seems to work on my laptop. Thanks for the quick fix!

R1kM (approved):

Issue #159, opened 2024-04-30
Support for mutually recursive traits and functions


msprotz:
I am getting this error:

```
   Compiling libcrux-ml-kem v0.0.2-pre.2 (/Users/jonathan/Code/libcrux/libcrux-ml-kem)
warning: Could not find clauses for trait obligations:core::ops::function::FnMut<Fold, (Acc, core::iter::adapters::enumerate::{impl core::iter::traits::iterator::Iterator for @Adt15<I>#1}<I>[@TraitClause0]::Item)>

         Available clauses:
         [@TraitClause0]: core::iter::traits::iterator::Iterator<I>
         [@TraitClause1]: core::ops::function::FnMut<Fold, (Acc, (usize, @TraitClause0::Item))>
         [@TraitClause2]: @TraitDecl21<R>
         [Self]: core::iter::traits::iterator::Iterator<@Adt15<I>>
         - context: core::iter::adapters::enumerate::{impl#1}::try_fold
 --> /rustc/d59363ad0b6391b7fc5bbb02c9ccf9300eef3753/library/core/src/iter/adapters/enumerate.rs:73:5

warning: Could not find clauses for trait obligations:core::ops::function::FnMut<Fold, (Acc, core::iter::adapters::enumerate::{impl core::iter::traits::iterator::Iterator for @Adt15<I>#1}<I>[@TraitClause0]::Item)>

         Available clauses:
         [@TraitClause0]: core::iter::traits::iterator::Iterator<I>
         [@TraitClause1]: core::ops::function::FnMut<Fold, (Acc, (usize, @TraitClause0::Item))>
         [Self]: core::iter::traits::iterator::Iterator<@Adt15<I>>
         - context: core::iter::adapters::enumerate::{impl#1}::fold
 --> /rustc/d59363ad0b6391b7fc5bbb02c9ccf9300eef3753/library/core/src/iter/adapters/enumerate.rs:96:5

thread 'rustc' panicked at 'Invalid scc:
libcrux_ml_kem::simd::portable::{impl libcrux_ml_kem::simd::simd_trait::Operations for libcrux_ml_kem::simd::portable::PortableVector}
libcrux_ml_kem::simd::portable::ntt_multiply
libcrux_ml_kem::simd::portable::deserialize_1
libcrux_ml_kem::simd::portable::deserialize_4
libcrux_ml_kem::simd::portable::deserialize_5
libcrux_ml_kem::simd::portable::deserialize_10
[...]

PR #160, opened 2024-05-01
Support mutually recursive traits

### LINKED ISSUE #94: Add support for mutually recursive trait definitions
Mutually recursive traits happen in practice, for example [here](https://github.com/arkworks-rs/algebra/blob/bb663bc27e776af81e66c94d9016bde2dbba9968/ff/src/fields/prime.rs#L27) (mutual recursion between `PrimeField` and `Field`):
```rust
pub trait Trait1 {
    type T: Trait2;
}

pub trait Trait2: Trait1 {}
```

It would be good to handle those, or at least print a nice error message. For now, Charon panics with:
```
thread 'rustc' panicked at 'Invalid trait decl group:
tests4::Trait1
tests4::Trait2', src/reorder_decls.rs:116:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: could not compile `tests4` (bin "tests4")
```

R1kM:
This PR extends support of Charon for mutually recursive trait declarations, and changes the charon-ml GAst and serialization accordingly.

In addition to the Charon support, it also adds one small example to traits.rs, and moves the unsupported recursive traits issue to supported tests.

Fixes #94


charon/tests/ui/issue-94-recursive-trait-defns.rs
```diff
@@ -1,4 +1,3 @@
-//@ known-panic
 pub trait Trait1 {
     type T: Trait2;
 }
```
Nadrieril, line 3: Could you add more examples of ways for traits to be mutually recursive? E.g. what you added to the other file, mutual reference via method signatures, via method calls, implicit reference due to parent traits, etc, as well as examples that could look recursive but aren't.

charon/src/reorder_decls.rs
```diff
@@ -101,27 +101,22 @@ impl DeclarationGroup {
...
         gr: impl Iterator<Item = TraitDeclId::Id>,
     ) -> Self {
         let gr: Vec<_> = gr.collect();
-        let ctx = ctx.into_fmt();
         // Trait declarations often refer to `Self`, like below,
         // which means they are often considered as recursive by our
         // analysis. TODO: do something more precise. What is important
         // is that we never use the "whole" self clause as argument,
         // but rather projections over the self clause (like `<Self as Foo>::u`,
         // in the declaration for `Foo`).
```
Nadrieril, line 114: If you understand what this TODO refers to, I'd like us to have a test for a case like that

Nadrieril: This TODO is exactly the error we got here: https://github.com/AeneasVerif/charon/issues/159

charon/tests/ui/issue-94-recursive-trait-defns.rs
R1kM: I moved the tests to charon/tests/ui (as they are unsupported by Aeneas) and added a few extra tests.
As `dyn` is currently unsupported, I do not think I can add tests using method signatures, but please feel free to expand if you have other ideas of patterns to test.

Nadrieril:
Looks good. Could you squash your commits together? Then we can merge

PR #169, opened 2024-05-03
Run aeneas in CI

Nadrieril:
This is a straightforward extension of https://github.com/AeneasVerif/charon/pull/155. I thought I had to clean up aeneas tests before we could do that but turns out that no.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,13 @@ jobs:
       - run: nix build -L .#charon-ml
       - run: nix flake check -L
 
+  aeneas:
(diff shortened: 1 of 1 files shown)


sonmarcho:
What's the workflow now? If by updating Charon I break Aeneas, how do I update Aeneas so that the CI becomes green for Charon?

Nadrieril:
You fix aeneas normally, and the next charon PR should use that. Unfortunately because of the `skip_duplicate_run` thing we can't re-run the check by hand I think. It would be nice if we could tell GitHub that this job isn't important

PR #173, opened 2024-05-06
Remove leftover arithmetic overflow checks

Nadrieril:
In https://github.com/AeneasVerif/charon/pull/129 we changed how we removed overflow checks and friends, but there remained some checks in constants that we decided to ignore. This reuses the old code before https://github.com/AeneasVerif/charon/pull/129 to handle these checks too.


Nadrieril:
This unbreaks kyber

PR #192, opened 2024-05-22
Don't run charon on build scripts

### LINKED ISSUE #190: Handle build.rs
Charon can't handle crates with a `build.rs` right now. It generates a `build_script_build.llbc` instead and then fails with

```
error: failed to run custom build command for `charon-build-rs v0.1.0`

Caused by:
  could not execute process `target/debug/build/charon-build-rs-28ab3466db3344d5/build-script-build` (never executed)

Caused by:
  No such file or directory (os error 2)
```

To reproduce create a new crate and give it a `build.rs` file. It only needs an empty `main` function.

Nadrieril:
By default, cargo calls our driver on build scripts. This obviously fails to compile since we don't generate a binary. To avoid that, I copied [what Miri does](https://github.com/rust-lang/miri/blob/53481c468db62d3c305499dcab213eacef6979eb/cargo-miri/src/phases.rs#L245-L257), which involves careful use of the `--target` flag. This allows us to detect when we're called on a build script and run rustc normally instead of the charon driver.

Fixes #190


sonmarcho (approved):
This looks good to me, and I think doing the same as MIRI is safe.
Is there any subtle point I should be aware of and that I may have overlooked?

Nadrieril:
I don't think so, let's merge it

PR #197, opened 2024-05-24
Use printf instead of echo -n in Makefile to avoid portability issues

R1kM:
As described in the title.
I've recently been having issues when regenerating CharonVersion.ml through `echo -n` on MacOS, where the echo version used does not seem to recognize the `-n` option, leading to printing it in the file and hence to syntax errors.

Following this [StackOverflow post](https://stackoverflow.com/questions/11193466/echo-n-prints-n), this PR replaces the `echo -n` call by `printf`, which apparently has more consistent behavior across platforms.


sonmarcho:
@Nadrieril the CI failure is dubious. It seems to be failing because of the PR which adds the `Unsolved` trait instance.

R1kM:
I reran the CI, I think this hit just between the merge of the PRs related to `Unsolved` on the Charon and Aeneas sides.

Nadrieril (approved):

PR #201, opened 2024-05-27
Add support for `Charon.toml`

Nadrieril:
With this PR, users can set some charon options in `crate/Charon.toml`.


sonmarcho (approved):
This is a very good idea!

sonmarcho:
This got me thinking: maybe we also want an `Aeneas.toml` file? (I can open an issue)

Issue #112, opened 2024-04-10
Add `StatementKind::Error`


Nadrieril:
For cases like in https://github.com/AeneasVerif/charon/pull/107 where a pass fails.

PR #208, opened 2024-05-29
Added Error to RawStatement

EschericHya:
Fixes https://github.com/AeneasVerif/charon/issues/112


charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -1014,6 +1014,7 @@ impl llbc::Statement {
                     tab
                 )
             }
+            RawStatement::Error => format!("{tab}panic"),
```
Nadrieril, line 1017: `Error` isn't a runtime panic in the emitted code, it's a compile-time error in Charon. I think we can indicate that like:
```suggestion
            RawStatement::Error => format!("{tab}/* charon error */"),
```

Nadrieril:
Looks good, we just need a test now. @sonmarcho what kind of error should we use this for?

sonmarcho:
> Looks good, we just need a test now. @sonmarcho what kind of error should we use this for?

If you look carefully we do not actually introduce the `Error` kind anywhere at this point, so we can't test it: we need to update the translation from Hax to ULLBC first. However, it might break some micro-passes (not sure, though) so I suggest doing that in a subsequent PR where we will add tests. For the missing features, anything will do and if you need inspiration just look at the `translate_statement` function (we can use intrinsics, some unsafe features, async/await, yield, etc. - I leave it to you guys :) ).

sonmarcho:
On a different topic: the CI failure is bizarre, right?

charon/src/ast/llbc_ast.rs
```diff
@@ -59,6 +59,7 @@ pub enum RawStatement {
     Sequence(Box<Statement>, Box<Statement>),
     Switch(Switch),
     Loop(Box<Statement>),
+    Error,
```
sonmarcho, line 62: I suggest adding a string with the unsupported statement: it can be useful for debugging (Lucas does this a lot in hax for instance).

sonmarcho (changes requested):

Nadrieril:
> On a different topic: the CI failure is bizarre, right?

It happens because we changed the `Cargo.toml`, which updates `Cargo.lock` on the next run of `cargo`, but the `Cargo.lock` was not commited.

EschericHya:
I will commit it with the change suggested by @sonmarcho

EschericHya:
Raw_statement_to_json and charon-ml related changes still need to be propagated

charon-ml/src/LlbcAstUtils.ml
```diff
@@ -37,7 +37,7 @@ let rec chain_statements (st1 : statement) (st2 : statement) : statement =
       (* Simply create a sequence *)
       mk_sequence st1 st2
   | Nop -> (* Ignore the nop *) st2
-  | Break _ | Continue _ | Panic | Return ->
+  | Break _ | Continue _ | Panic | Return | Error _ ->
```
sonmarcho, line 40: Move the `Error` case with `SetDiscriminant`, `Assert`, etc.

charon-ml/src/PrintLlbcAst.ml
```diff
@@ -107,6 +107,7 @@ module Ast = struct
         indent ^ "loop {\n"
         ^ statement_to_string env (indent ^ indent_incr) indent_incr loop_st
         ^ "\n" ^ indent ^ "}"
+    | Error s -> indent ^ "error " ^ s
```
sonmarcho, line 110:
```suggestion
    | Error s -> indent ^ "ERROR(' " ^ s ^"')"
```

charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -873,6 +873,9 @@ impl ullbc::Statement {
             RawStatement::Deinit(place) => {
                 format!("@deinit({})", place.fmt_with_ctx(ctx))
             }
+            RawStatement::Error(s) => {
+                format!("@Error{}", s)
```
sonmarcho, line 877:
```suggestion
                format!("@ERROR({})", s)
```

charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -1014,6 +1017,7 @@ impl llbc::Statement {
                     tab
                 )
             }
+            RawStatement::Error(s) => format!("{tab}{}", s),
```
sonmarcho, line 1020:
```suggestion
            RawStatement::Error(s) => format!("{tab}@ERROR({})", s),
```

sonmarcho (changes requested):

sonmarcho (approved):
Looks good, thanks!

Nadrieril:
@msprotz this adds a new variant, eurydice will need to be updated

Nadrieril:
once again foiled by the lazy CI, I forgot to ping you beforehand sorry

PR #209, opened 2024-05-29
Allow slice patterns for destructuring arrays & slices

### LINKED ISSUE #203: Failure when destructuring arrays
Charon fails with the error `Unexpected ProjectionElem::Subslice` on each function of the following code snippet:

```rust
fn f(x: [(); 1]) {
    let [_named] = x;
}

fn g(x: &[(); 1]) {
    let [_named] = x;
}

fn h(x: [(); 1]) {
    match x {
        [_named] => (),
    }
}

fn i(x: &[()]) {
    match x {
        [_named] => (),
        _ => (),
    }
}
```

Note that the failure is present only when a binding is introduced, e.g. the following doesn't fail:
```rust
fn f(x: [(); 1]) {
    let [_] = x;
}
```

W95Psp:
This PR changes the `ProjectionElem::Index` variant so that it takes an `Operand` instead of a `VarId`. It then changes the `index_to_function_calls` pass to create a local when needed.

I'm using `Operand` as a way of encoding a `Either<VarId, ConstantExpr>`, I'm not sure that's entirely satisfactory. The hax -> ULLBC pass will never produce the variant `Operand::Move`, whence the `panic!` I introduced in `index_to_function_calls`.

Fixes #203


charon/src/transform/index_to_function_calls.rs
```diff
@@ -36,9 +36,26 @@ impl<'a> Transform<'a> {
...
+                        };
+                        let content = RawStatement::Assign(place, Rvalue::Use(operand));
+                        let span = self.span.unwrap();
+                        self.statements.push(Statement { span, content });
+                        var_id
+                    }
+                    Operand::Move(_) => {
+                        panic!("Unexpected `Oparand::Move` on a `ProjectionElem::Index`")
+                    }
+                };
```
Nadrieril, line 58: `index_var_id` is only used below as `let arg_index = Operand::Copy(Place::new(index_var_id))`. You don't need to add an `Assign` at all, you can simply do `let arg_index = operand;`.

charon/src/translate/translate_functions_to_ullbc.rs
```diff
@@ -439,16 +439,33 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
...
+                        from_end: false,
+                    } => {
+                        let ty = Ty::Literal(LiteralTy::Integer(IntegerTy::Usize));
+                        let value =
+                            RawConstantExpr::Literal(Literal::Scalar(ScalarValue::Usize(*offset)));
+                        let offset = Operand::Const(ConstantExpr { value, ty });
+                        projection.push(ProjectionElem::Index(offset, current_ty));
+                    }
+                    hax::ProjectionElem::ConstantIndex { .. } => {
+                        // Those don't seem to occur in MIR built
```
Nadrieril, line 465: I doubt this comment is true. Could you add some tests with some variable-length slice patterns to test that?

Nadrieril:
Since you changed the charon AST, you'll have to update the corresponding types on the `charon-ml` side. In this case it's `projection_elem` in `charon-ml/src/Expressions.ml`. Please also increment the version in `Cargo.toml` to indicate an ast change, and run `make test` in the crate root to ensure everything works.

charon/src/transform/index_to_function_calls.rs
W95Psp: Oh, right, fantastic! Thanks!

W95Psp:
Thanks for the review! :)

Yeah, I saw the CI was not happy, but I don't understand why the ML parsers are failing.
The only thing that changed in the AST is `ProjectionElem::Index`. This seems not handled at all currently in the parser `projection_elem_of_json`.
Also, those should be gone at LLBC, if I understand correctly?

charon/tests/ui/arrays.out
```diff
@@ -1747,19 +1737,17 @@ fn test_crate::slice_pattern_4<'_0>(@1: &'_0 (Slice<()>))
...
     @3 := const (1 : usize)
     @4 := move (@2) == move (@3)
     if move (@4) {
-        @6 := const (0 : usize)
-        @7 := &*(x@1)
-        @8 := @SliceIndexShared<'_, ()>(move (@7), copy (@6))
-        _named@5 := &*(@8)
+        @6 := &*(x@1)
+        @7 := @SliceIndexShared<'_, ()>(move (@6), const (0 : usize))
+        _named@5 := &*(@7)
```
Nadrieril: Ah, we have a `simplify_constants` pass that moves constant operands into separate locals but unfortunately it runs before `index_to_function_calls`. That's why there's a change here; that's not great. Unfortunately `simplify_constants` operates on ullbc and `index_to_function_calls` operates on llbc so we can't easily reorder them... We might have to rewrite `simplify_constants` to operate on llbc, it's not the first time I've wanted that.

Nadrieril:
Ah, the CI failure is because of a `RValue::Len`. These should have all been removed by our micro-passes, I don't understand how this PR could have changed that. We should give a better error.

charon/tests/ui/arrays.out
Nadrieril: Oops I'm pointing at the wrong thing

charon/tests/ui/arrays.out
```diff
@@ -1655,6 +1655,109 @@ fn test_crate::sum_mut_slice<'_0>(@1: &'_0 mut (Slice<u32>)) -> u32
...
+    let @6: &'_ mut (Array<&'_ mut (T), 3 : usize>); // anonymous local
+    let @7: &'_ mut (&'_ mut (T)); // anonymous local
+    let @8: &'_ mut (Array<&'_ mut (T), 3 : usize>); // anonymous local
+    let @9: &'_ mut (&'_ mut (T)); // anonymous local
+    let @10: &'_ mut (Array<&'_ mut (T), 3 : usize>); // anonymous local
+    let @11: &'_ mut (&'_ mut (T)); // anonymous local
+
+    @fake_read(x@1)
+    @10 := &mut x@1
+    @11 := @ArrayIndexMut<'_, &'_ mut (T), 3 : usize>(move (@10), const (0 : usize))
```
Nadrieril, line 1695: We have a `simplify_constants` pass that moves constant operands into separate locals but unfortunately it runs before `index_to_function_calls`. I would want it to apply here, so we don't get a `const` as function argument. Unfortunately `simplify_constants` operates on ullbc and `index_to_function_calls` operates on llbc so we can't easily reorder them... We might have to rewrite `simplify_constants` to operate on llbc, it's not the first time I've wanted that.

charon/tests/ui/arrays.out
```diff
@@ -1655,6 +1655,109 @@ fn test_crate::sum_mut_slice<'_0>(@1: &'_0 mut (Slice<u32>)) -> u32
...
+    let x@1: &'_ (Slice<()>); // arg #1
+    let @2: usize; // anonymous local
+    let @3: usize; // anonymous local
+    let @4: bool; // anonymous local
+    let _named@5: &'_ (()); // local
+    let @6: &'_ (Slice<()>); // anonymous local
+    let @7: &'_ (()); // anonymous local
+
+    @fake_read(x@1)
+    @2 := len(*(x@1))
```
Nadrieril, line 1744: Here's the `RValue::Len` that caused the CI failure. I guess we just have to parse it correctly, because it's important t have it here.

charon/tests/ui/arrays.out
Nadrieril: We can do this later, I'll ask Son about it

Nadrieril: Son says it's fine

sonmarcho: Clarifying here: `simplify_constants` it transforms the AST to remove some cases of constants altogether (for instance: constant adts) so that they don't duplicate other operand cases, and it does so by moving some constants to auxiliary operands (among other things).

sonmarcho (approved):
Looks good, thanks! But we need to propagate the changes to Charon-ML *and* Aeneas before merging.

sonmarcho:
I'm merging because @protz really needs it. Something is broken on the side of Aeneas however.

PR #216, opened 2024-05-31
Export type aliases

### LINKED ISSUE #134: Export type aliases in LLBC
This is particularly useful for Eurydice. I have a mechanism for picking user-provided type abbreviations, but instead end up generating things like this https://github.com/cryspen/hacl-packages/pull/457/files#diff-a94673428367ed62df2e7dd7d29ef4b6576ee2bd5e9cf33beed83e3f452e3602R106-R115 because I no longer have the information that the user had provided a custom name (much shorter) for these types.

This doesn't need to change anything else in Charon, the type abbreviations can still be inlined, it's just a matter of keeping the type alias also in scope for me to do something useful with it. Thanks!

Nadrieril:
This adds a new kind of type declaration to represent type aliases, and exports all crate-local type aliases. This is purely cosmetic, as type aliases are inlined everywhere. Because of this inlining we also can't do our usual reachability analysis to also export relevant type aliases from other crates.

Fixes #134.


Nadrieril:
@sonmarcho I did it like this because it was easiest, but I think the clean way to do this would be to add a new kind of item instead, since aliases can't actually show up in types. What do you think?

msprotz:
I think it should, let me give it a go today

Nadrieril:
@msprotz I may still change the representation, I don't want you to waste effort. I'll let you know when I'm happy with the representation

Nadrieril:
I think the better long-term solution is to define a new kind of item, so this doesn't mislead consumers of charon. Right now adding a new kind of item is a bit of a pain though.

msprotz:
FYI your current implementation seems 100% fine to me, and downstream consumers are welcome to ignore aliases if they want to, so adding a new kind of item seems a little more work for you and clients, actually

so in other words if you wanted to merge this as-is, I would be in favor

sonmarcho (approved):
Looks good to me.
For the changes on the side of Aeneas, we should simply implement a pre-pass (see https://github.com/AeneasVerif/aeneas/blob/main/compiler/PrePasses.ml) which removes the type aliases from the crate, and everything should be fine.

Nadrieril:
Ok, let's do it like this then. I'll get the aeneas change ready

Nadrieril:
The aeneas change is ready. @msprotz Let me know when you're ready on the eurydice side and I'll merge this.

sonmarcho (approved):

msprotz:
Actually, scrap that -- I just committed a proper fix in branch protz_type_aliases of Eurydice. I was able to confirm that not only does mlkem still compile, we're already witnessing code quality improvements such as this:

```diff
-libcrux_ml_kem_types_MlKemCiphertext____1088size_t
+libcrux_ml_kem_mlkem768_MlKem768Ciphertext
 libcrux_ml_kem_types___core__convert__From__Array_u8__SIZE___for_libcrux_ml_kem__types__MlKemCiphertext_SIZE___2__from___1088size_t(
     uint8_t value[1088U]);
     ```
     
 So this is great

msprotz:
So just merge the Eurydice branch at the same time as this branch and we'll be good. Thanks again

PR #218, opened 2024-05-31
Various bits of cleanup

Nadrieril:
- I moved some modules around, making a bit of progress towards #178;
- I grouped `fmt_with_ctx` into a trait which may one day be useful. That was supposed to be useful to simplify `Formatter` but I hit a trait solving edge case (pretty sure it's coinduction :smiling_face_with_tear:);
- Other details.


charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -14,11 +14,41 @@ use crate::{
...
+        // By default use no tab.
+        self.fmt_with_ctx_and_indent("", ctx)
+    }
+
+    fn fmt_with_ctx_and_indent(&self, _tab: &str, _ctx: &C) -> String {
+        panic!("This type does not know how to be formatted with indent")
+    }
+
+    /// Returns a struct that implements `Display`. This allows the following:
+    ///     println!("{}", self.with_ctx(ctx));
```
sonmarcho, line 29:
```suggestion
    /// ```text
    ///     println!("{}", self.with_ctx(ctx));
    /// ```
```

sonmarcho (approved):
Looks good, thanks!

PR #222, opened 2024-06-03
Update hax and rustc

Nadrieril:
This is the companion PR to https://github.com/hacspec/hax/pull/690


Nadrieril:
@msprotz This is a PR that updates the rustc version. Rustc changed some of its code generation as you can see from the test differences. This apparently breaks kyber, do you understand why? The error output is:
```
[ INFO charon_lib::export:111] [GCrateData<_, _>::serialize_to_file]: Generated the file: /run/github-runner/charon-ci/charon/charon/kyber-crate/libcrux_kyber.llbc
warning: `libcrux_kyber` (lib) generated 2 warnings (run `cargo fix --lib -p libcrux_kyber` to apply 2 suggestions)
    Finished dev [unoptimized + debuginfo] target(s) in 26.62s
Running eurydice ...
Cannot re-check libcrux_kyber.sampling.sample_from_uniform_distribution_next as valid Low* and will not extract it.  If libcrux_kyber.sampling.sample_from_uniform_distribution_next is not meant to be extracted, consider marking it as Ghost, noextract, or using a bundle. If it is meant to be extracted, use -dast for further debugging.

1️⃣ LLBC ➡️  AST
2️⃣ Cleanup
Warning 4: in the then branch, after the definition of uu____981, in the sequence statement at index 0, after the definition of iter, in top-level declaration libcrux_kyber.sampling.sample_from_uniform_distribution_next, in file libcrux_kyber_common: Malformed input:
subtype mismatch, () (a.k.a. ()) vs bool (a.k.a. bool)
LowStar.Ignore.ignore doesn't go anywhere
File order after topological sort: LowStar_Ignore, core, libcrux_digest, libcrux_kyber_common, libcrux_kyber768
Error: Process completed with exit code 1.
```

msprotz:
Could you try passing `--log '*'` to Eurydice and seeing where it differs?

Nadrieril:
kyber main is broken on charon+eurydice main right now so we'll have to debug that first.

Issue #206, opened 2024-05-29
Add an attribute to mark definitions as `opaque`


sonmarcho:
It is currently possible to pass an option to Charon to mark some modules as opaque, so that Charon will not look inside those definitions (it treats them similarly to external definitions). This option is a bit coarse grained: it would be good to be able to individually mark the functions with a `charon::opaque` attribute. In order not to confuse the Aeneas users, I suggest introducing a `aeneas::opaque` attribute which does the same, and generally speaking introducing both a `charon::...` and an `aeneas::...` version of every new attribute.

Nadrieril:
Note: rustc will complain about such an attribute because it will try to find a macro with that name. Adding
```rust
#![feature(register_tool)]
#![register_tool(charon)]
```
at the top of each file that uses the attributes should solve the problem I hope. We can think of a cleaner solution afterwards. (See [here](https://github.com/rust-lang/rust/issues/66079) for details about this feature).

sonmarcho:
Ok, thanks for looking into this. My understanding is that there are clean ways of doing this, but I'm indeed not aware of the details. It might be worth looking at other verification projects which use attributes (e.g., Verus or Creusot).

Nadrieril:
Creusot defines a proc-maro for their `#[ensures]`/`#[invariant]` etc attributes. That's one of the options available (and you started doing something like that with the `attributes` crate).

sonmarcho:
What do you recommend doing?
I initially thought that was the only way of doing, but I would prefer a way which doesn't require adding a dependency to Charon directly in the Rust code.

Nadrieril:
`register_tool` is the simplest. I've heard discussions on the rustc side about what the best design for this feature would be, so I'd suggest using it for now and they'll likely figure out something more practical, e.g. a line to add in your `Cargo.toml`.

zhassan-aws:
In case it helps, in Kani, we pass `feature(register_tool)` as part of the Rust flags:

https://github.com/model-checking/kani/blob/4a889397bba73fb76f0a21aad66171daad0b2b31/kani-driver/src/call_single_file.rs#L175-L178

Nadrieril:
https://github.com/AeneasVerif/charon/pull/227 allows the attribute on definitions; it would be nice to also allow the attribute on modules to apply to the whole module.

Nadrieril:
These attributes are supported on modules since one of the recent PRs that overhauled opaque handling.

PR #227, opened 2024-06-04
Escherichia/opaque

EschericHya:
Fixes https://github.com/AeneasVerif/charon/issues/206


charon/src/translate/translate_ctx.rs
```diff
@@ -452,6 +452,9 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
             attributes: self.translate_attributes_from_rid(def_id),
             inline: self.translate_inline_from_rid(def_id),
             public,
+            opaque: self
+                .translate_attributes_from_rid(def_id)
+                .contains(&(String::from("charon::opaque"))),
```
sonmarcho, line 457: As per the description of the original issue (https://github.com/AeneasVerif/charon/issues/206) we also want to check for "aeneas::opaque"

charon/src/translate/translate_functions_to_ullbc.rs
```diff
@@ -1833,7 +1833,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
         trace!("Translating function signature");
         let signature = bt_ctx.translate_function_signature(rust_id)?;
 
-        let body = if !is_trait_method_decl {
+        let body = if !is_trait_method_decl && !item_meta.opaque {
```
sonmarcho, line 1836: In order to keep the code consistent, this check should not be here, but in `translate_body`. To make this work we should add a reference to an ItemMeta as input to `translate_body`.

charon/src/translate/translate_functions_to_ullbc.rs
```diff
@@ -1911,12 +1911,15 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
 
         // Translate its body like the body of a function. This returns `None` if we can't/decide
         // not to translate this body.
-        let body = match bt_ctx.translate_body(rust_id, 0) {
-            Ok(body) => body,
-            // Error case: we could have a specific variant
-            Err(_) => None,
+        let body = if item_meta.opaque {
```
sonmarcho, line 1914: This check should be in `translate_body`

charon/src/translate/translate_traits.rs
```diff
@@ -246,9 +246,12 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
         trace!("About to translate trait decl:\n{:?}", rust_id);
 
         let def_id = self.register_trait_decl_id(&None, rust_id)?;
+
+        let item_meta = self.translate_item_meta_from_rid(rust_id);
+
         // We may need to ignore the trait (happens if the trait is a marker
         // trait like [core::marker::Sized]
-        if def_id.is_none() {
+        if def_id.is_none() || item_meta.opaque {
```
sonmarcho, line 254: This check ignores the trait altogether: you should remove the `item_meta.opaque` from here (we do this for very specific traits that we completely ignore so that they do not appear anywhere in the crate).

We actually have no way of completely ignoring trait declarations at this point: if `item_meta.opaque` is `true` you should print a warning stating that this attribute has no effect on trait declarations and that we will thus ignore (with span, name of the declaration, etc.).

charon/src/translate/translate_traits.rs
```diff
@@ -432,17 +435,17 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
         trace!("About to translate trait impl:\n{:?}", rust_id);
 
         let def_id = self.register_trait_impl_id(&None, rust_id)?;
+        let tcx = self.tcx;
+        let span = tcx.def_span(rust_id);
+        let mut bt_ctx = BodyTransCtx::new(rust_id, self);
+        let item_meta = bt_ctx.t_ctx.translate_item_meta_from_rid(rust_id);
         // We may need to ignore the trait
-        if def_id.is_none() {
+        if def_id.is_none() || item_meta.opaque {
```
sonmarcho, line 443: Same comment as for trait decls above

charon/src/translate/translate_traits.rs
```diff
@@ -409,7 +412,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
             def_id,
             is_local: rust_id.is_local(),
             name,
-            item_meta: self.translate_item_meta_from_rid(rust_id),
+            item_meta: item_meta, // self.translate_item_meta_from_rid(rust_id),
```
sonmarcho, line 415:
```suggestion
            item_meta,
```

charon/src/translate/translate_traits.rs
```diff
@@ -622,7 +625,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
             def_id,
             is_local: rust_id.is_local(),
             name,
-            item_meta: bt_ctx.t_ctx.translate_item_meta_from_rid(rust_id),
+            item_meta: item_meta,
```
sonmarcho, line 628:
```suggestion
            item_meta,
```

sonmarcho (changes requested):

sonmarcho:
I'm asking for a few modifications, but overall it works.

charon/src/translate/translate_functions_to_ullbc.rs
Nadrieril: Should we? I feel like `translate_body` should be called when we want to translate a body.

charon/src/translate/translate_traits.rs
Nadrieril: We could translate the trait name but not translate its methods. That way it can be used as a bound.

sonmarcho: Correct, but we often lookup trait methods in other places, so we have to pay attention not to trigger bugs by doing something like this.

Nadrieril: Fair, we can do that later if we have a use for it

charon/src/translate/translate_ctx.rs
```diff
@@ -452,6 +452,10 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
             attributes: self.translate_attributes_from_rid(def_id),
             inline: self.translate_inline_from_rid(def_id),
             public,
+            opaque: self
+                .translate_attributes_from_rid(def_id)
```
Nadrieril, line 456: Please compute `translate_attributes_from_rid` only once; it is already computed a few lines above.

charon/src/translate/translate_functions_to_ullbc.rs
```diff
@@ -1471,16 +1471,21 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
         mut self,
         rust_id: DefId,
         arg_count: usize,
+        item_meta: &crate::meta::ItemMeta,
```
Nadrieril, line 1474:
```suggestion
        item_meta: &ItemMeta,
```
and add a `use crate::meta::ItemMeta` at the start of the file.

charon/src/translate/translate_traits.rs
```diff
@@ -401,15 +404,18 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
                 generic_clauses
             );
         }
-
+        if item_meta.opaque {
+            let ctx = bt_ctx.into_fmt();
+            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The aeneas::opaque or charon::opaque currently has no effect on trait declaration and will be ignored. \n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```
Nadrieril, line 409:
```suggestion
            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The `aeneas::opaque` or `charon::opaque` attribute currently has no effect on trait declarations and will be ignored. \n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```

charon/src/translate/translate_traits.rs
```diff
@@ -618,11 +624,16 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
             }
         }
 
+        if item_meta.opaque {
+            let ctx = bt_ctx.into_fmt();
+            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The aeneas::opaque or charon::opaque currently has no effect on trait declaration and will be ignored. \n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```
Nadrieril, line 629:
```suggestion
            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The `aeneas::opaque` or `charon::opaque` attribute currently has no effect on trait implementations and will be ignored. \n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```

Nadrieril:
Could you add a test for all the new cases you cover? That means adding `charon/tests/ui/opaque_attribute.rs` that contains a few cases like the atribute on a function, on a trait etc to record the output.

charon/src/translate/translate_ctx.rs
```diff
@@ -447,13 +447,13 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+        let attributes = self.translate_attributes_from_rid(def_id);
         ItemMeta {
             span,
-            attributes: self.translate_attributes_from_rid(def_id),
+            attributes: attributes.clone(),
             inline: self.translate_inline_from_rid(def_id),
             public,
-            opaque: self
-                .translate_attributes_from_rid(def_id)
+            opaque: attributes
```
Nadrieril, line 456: You can compute `opaque` before the `ItemMeta { .. }` to avoid needing to clone `attributes`.

charon/src/translate/translate_functions_to_ullbc.rs
```diff
@@ -1462,16 +1463,21 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
...
         // Stopgap measure because there are still many panics in charon and hax.
-        let mut this = panic::AssertUnwindSafe(&mut self);
-        let res = panic::catch_unwind(move || this.translate_body_aux(rust_id, arg_count));
-        match res {
-            Ok(Ok(body)) => Ok(body),
-            Ok(Err(e)) => Err(e),
-            Err(_) => {
-                let span = self.t_ctx.tcx.def_span(rust_id);
-                error_or_panic!(self, span, "Thread panicked when extracting body.");
+        if item_meta.opaque {
```
sonmarcho, line 1469: I wrote we should put this with the other checks for opacity in `translate_body` but it was actually `translate_body_aux` (see the checks starting at line 1493).

charon/src/translate/translate_functions_to_ullbc.rs
```diff
@@ -1462,16 +1463,21 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
         mut self,
         rust_id: DefId,
         arg_count: usize,
+        item_meta: &ItemMeta,
     ) -> Result<Option<ExprBody>, Error> {
         // Stopgap measure because there are still many panics in charon and hax.
```
sonmarcho, line 1468: Please put attention to the comments when you update the code. This comment is about the `panic::catch_unwind` and should have been moved with it: it has nothing to do with the check `if item_meta.opaque`.

charon/src/translate/translate_traits.rs
```diff
@@ -389,15 +389,19 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
                 generic_clauses
             );
         }
-
+        let item_meta = bt_ctx.t_ctx.translate_item_meta_from_rid(rust_id);
+        if item_meta.opaque {
+            let ctx = bt_ctx.into_fmt();
+            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The `aeneas::opaque` or `charon::opaque` attribute currently has no effect on trait declarations and will be ignored. \n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```
sonmarcho, line 395:
```suggestion
            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The `aeneas::opaque` or `charon::opaque` attribute currently has no effect on trait declarations and will be ignored.\n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```
This is a super long line, could you break it?
(note that I'm surprised that the formatter doesn't do it on itself)

charon/src/translate/translate_traits.rs
```diff
@@ -594,12 +598,17 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
                 }
             }
         }
+        let item_meta = bt_ctx.t_ctx.translate_item_meta_from_rid(rust_id);
+        if item_meta.opaque {
+            let ctx = bt_ctx.into_fmt();
+            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The `aeneas::opaque` or `charon::opaque` attribute currently has no effect on trait implementations and will be ignored. \n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```
sonmarcho, line 604:
```suggestion
            bt_ctx.t_ctx.errors.session.span_warn(item_meta.span, format!("The `aeneas::opaque` or `charon::opaque` attribute currently has no effect on trait implementations and will be ignored.\n- Declaration name: {}\n", name.fmt_with_ctx(&ctx)))
```
break line

sonmarcho (changes requested):
Still some nitpicks, but the modifications should be super quick to to.
And, yes, we should add some tests.

Nadrieril:
Hm, something is weird: the changes to charon-ml are missing but the CI is passing :thinking:

Nadrieril:
Oh sorry we skip serializing the field so there's no charon-ml change to do.

charon/src/bin/charon-driver/main.rs
```diff
@@ -122,6 +122,18 @@ fn main() {
...
+    // ```
+    // #![feature(register_tool)]
+    // #![register_tool(charon)]
+    // ```
+    compiler_args.push("-Z".to_string());
+    compiler_args.push("crate-attr=feature(register_tool)".to_string());
+    compiler_args.push("-Z".to_string());
+    compiler_args.push("crate-attr=register_tool(charon)".to_string());
+    compiler_args.push("-Z".to_string());
+    compiler_args.push("crate-attr=register_tool(aeneas)".to_string());
```
Nadrieril, line 136: Can you remove the `register_rool` clauses from the top of the new test file as well?

Nadrieril: Ah, we can't register the same tool twice. I'm afraid this means we have to let our users set this themselves

sonmarcho: Yes.

Issue #207, opened 2024-05-29
Add an attribute to rename definitions in the backend


sonmarcho:
We should introduce an attribute `charon::name(...)` (and `aeneas::name(...)`) to tweak the name of the code extracted by Charon (and Aeneas). This is to address https://github.com/AeneasVerif/aeneas/issues/211
This attribute should be applicable to: types, variants, fields, traits, functions, globals.

PR #237, opened 2024-06-07
Add an attribute to rename definitions in the backend

EschericHya:
Fixes https://github.com/AeneasVerif/charon/issues/207

Still need to write some tests I guess


charon/src/translate/translate_ctx.rs
```diff
@@ -451,12 +451,39 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
             .any(|attr| attr == "charon::opaque" || attr == "aeneas::opaque");
+        let rename = {
+            let pos = attributes.iter().position(|str| {
+                str.starts_with("charon::rename(") || str.starts_with("aeneas::rename(")
+            });
+            if pos.is_none() {
+                None
+            } else {
+                let str = match attributes
+                    .get(pos.expect("Should not be None, it was checked right before"))
```
sonmarcho, line 462: You should simply use `Option::unwrap`

sonmarcho (changes requested):

charon/src/translate/translate_ctx.rs
```diff
@@ -452,28 +452,43 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
             .iter()
             .any(|attr| attr == "charon::opaque" || attr == "aeneas::opaque");
         let rename = {
-            let pos = attributes.iter().position(|str| {
+            let str = attributes.iter().find(|str| {
                 str.starts_with("charon::rename(") || str.starts_with("aeneas::rename(")
             });
-            if pos.is_none() {
+            if str.is_none() {
```
sonmarcho, line 458: You can use the `if let ...` notation (you use it below actuallly) instead of `str.is_none()` then `str.unwrap()`.

charon/src/translate/translate_ctx.rs
```diff
@@ -452,28 +452,43 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+                    if !str.is_empty() {
+                        let first_char_alphabetic = str
+                            .chars()
+                            .nth(0)
+                            .expect("Attribute `rename` should not be empty")
+                            .is_alphabetic();
+                        let is_identifier = first_char_alphabetic
+                            && str.chars().all(|c| c.is_alphanumeric() || c == '_');
+                        if !is_identifier {
+                            self.span_err(span, "Attribute `rename` should not be empty or should only contains alphanumeric characters and `_` or should start with a letter");
```
sonmarcho, line 477: "should not be empty": you already have a message for this above
"**or** should start": it is actually an **and**

charon/tests/ui/rename_attribute.rs
```diff
@@ -37,11 +37,16 @@ enum SimpleEnum {
...
+#[charon::rename(StructTest)]
+struct Foo {
+    #[charon::rename(FieldTest)]
+    field1: u32,
+}
 
-#[aeneas::rename(TypeAeneas)]
-type Test2 = u32;
+#[charon::rename(Const_Test)]
+const SIX_SIX_SIX: u32 = 600 + 60 + 6;
```
sonmarcho, line 47:
```suggestion
const C: u32 = 100 + 10 + 1;
```

charon/tests/ui/unsupported/rename_attribute_failure.rs
```diff
@@ -0,0 +1,14 @@
+//@ known-failure
+//@ no-check-output
```
sonmarcho, line 2: We actually want to check the output, to make sure the error messages are correctly maintained

sonmarcho (changes requested):

sonmarcho:
Put aside the cleanup above, I would like to replace the `Span` field of `Field` with `ItemMeta` so that we can collect the attributes given to structure fields (and potentially rename fields).

Nadrieril:
hm, if we use `ItemMeta` for things that aren't items then we shouldn't call it that

R1kM:
I'd be in favour of wrapping up and merging this, and discussing the further extension to fields (and the precise name) in a further issue/PR. Wdyt?

charon/src/translate/translate_ctx.rs
```diff
@@ -451,12 +451,55 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+                            .chars()
+                            .nth(0)
+                            .expect("Attribute `rename` should not be empty")
+                            .is_alphabetic();
+                        let is_identifier = first_char_alphabetic
+                            && str
+                                .chars()
+                                .all(|c| c.is_alphanumeric() || c == '_' || c == '-');
+                        if !is_identifier {
+                            self.span_err(span, "Attribute `rename` should only contains alphanumeric characters and `_` or `-` and should start with a letter");
```
sonmarcho, line 476:
```suggestion
                            self.span_err(span, "Attribute `rename` should only contains alphanumeric character, `_` and `-`, and should start with a letter");
```

charon/src/translate/translate_types.rs
```diff
@@ -601,12 +603,9 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
                     }
                 };
 
-                // Translate the span information
-                let span = self.translate_span_from_rspan(field_def.span);
-
                 // Store the field
                 let field = Field {
-                    span,
+                    item_meta: item_meta.clone(),
```
sonmarcho, line 608: You're using the meta information of the *type* here, but you should use the type information from the *field*, which is different.
Ex.:
```rust
#[rename("Type")]
struct S {
  #[rename("y")]
  x : u32
}
```

sonmarcho (changes requested):

charon/src/translate/translate_ctx.rs
```diff
@@ -441,12 +441,55 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
         let opaque = attributes
             .iter()
             .any(|attr| attr == "charon::opaque" || attr == "aeneas::opaque");
+        let rename = {
+            let str = attributes.iter().find(|str| {
+                str.starts_with("charon::rename(") || str.starts_with("aeneas::rename(")
```
Nadrieril, line 446: `starts_with` is redundant with the `strip_prefix` that follows

Nadrieril: To make this code more robust, it would also be nice to have a `parse_attribute(&str) -> Option<Attribute>` function, with
```rust
enum Attribute {
  Opaque,
  Rename(String),
}
```

charon/src/ast/types.rs
```diff
@@ -370,7 +370,7 @@ pub struct Variant {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Field {
-    pub span: Span,
+    pub item_meta: ItemMeta,
```
Nadrieril, line 373: Please add a comment like `FIXME: define a more appropriate container for attribute and visibility information.`

charon/src/ast/meta.rs
```diff
@@ -135,6 +135,7 @@ pub struct ItemMeta {
     /// API (this is called "pub-in-priv" items). With or without the `pub use`, we set `public =
     /// true`; computing item reachability is harder.
     pub public: bool,
+    pub rename: Option<String>,
```
Nadrieril, line 138: Please add a comment explaining what this field means

charon/src/translate/translate_types.rs
```diff
@@ -584,7 +582,9 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
             for (j, field_def) in var_def.fields.into_iter().enumerate() {
                 trace!("variant {i}: field {j}: {field_def:?}");
                 let field_span = field_def.span.rust_span_data.unwrap().span();
-
+                let item_meta = self
```
sonmarcho, line 585:
```suggestion
                let field_meta = self
```

charon/src/translate/translate_types.rs
```diff
@@ -605,7 +605,7 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
 
                 // Store the field
                 let field = Field {
-                    item_meta: item_meta.clone(),
+                    item_meta,
```
sonmarcho, line 608:
```suggestion
                    item_meta : field_meta,
```

sonmarcho (changes requested):

sonmarcho:
I just realized that we should also add some `ItemMeta` data to enumeration variants.

Nadrieril:
Beware: another PR updated the `Cargo.toml` version, so when you merge `main` you'll have to update it again

charon/src/translate/translate_ctx.rs
```diff
@@ -437,12 +437,64 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+                                .chars()
+                                .all(|c| c.is_alphanumeric() || c == '_' || c == '-');
+                        if !is_identifier {
+                            self.span_err(span, "Attribute `rename` should only contains alphanumeric characters, `_` and `-` and should start with a letter");
+                            None
+                        } else {
+                            Some(rename.unwrap().to_string())
+                        }
+                    } else {
+                        self.span_err(span, "Attribute `rename` should not be empty");
```
Nadrieril, line 471: Don't use `span_err` directly, we have macros that do the right thing and register errors. Probably you want to return a `Result<.., Error>` and use the macro `register_error_or_panic`

Nadrieril: actually no, it would be silly to have to skip the whole item if an attribute is malformed. returning `ItemMeta` is fine, bug we need to register the error so charon knows there was one

Nadrieril: oh, `span_err` does that already, nevermind

charon/src/translate/translate_ctx.rs
```diff
@@ -437,12 +437,64 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+            } else if filter_attribute.len() > 1 {
+                self.span_err(
+                    span,
+                    "There are too many `rename` attributes, please use only one",
+                );
+                None
+            } else {
+                None
+            }
+        };
```
Nadrieril, line 490: This whole thing is getting complex and hard to follow. Could you move it to another function, with early `return None` instead of nested `if`s? And please add some comments to explain the different phases of what you're doing.

charon/tests/crate_data.rs
```diff
@@ -189,3 +189,127 @@ fn discriminants() -> Result<(), Box<dyn Error>> {
...
+
+        #[aeneas::rename("Type-Aeneas36")]
+        type Test2 = u32;
+        "#,
+    )?;
+
+    assert_eq!(
+        crate_data.trait_decls[0].item_meta.rename,
+        Some("BoolTest".to_string())
+    );
```
Nadrieril, line 254:
```suggestion
    assert_eq!(
        crate_data.trait_decls[0].item_meta.rename.as_deref(),
        Some("BoolTest")
    );
```

charon/tests/crate_data.rs
```diff
@@ -189,3 +189,127 @@ fn discriminants() -> Result<(), Box<dyn Error>> {
...
+        Some("Const_Test".to_string())
+    );
+
+    assert_eq!(
+        crate_data.types[3].item_meta.rename,
+        Some("Type-Aeneas36".to_string())
+    );
+
+    let fieldvec = crate_data.types[2].kind.as_struct();
+    let field = fieldvec.get(0.into());
```
Nadrieril, line 309:
```suggestion
    let field = fieldvec[0];
```

charon/tests/crate_data.rs
```diff
@@ -189,3 +189,127 @@ fn discriminants() -> Result<(), Box<dyn Error>> {
...
+    assert_eq!(
+        field.unwrap().item_meta.rename,
+        Some("FieldTest".to_string())
+    );
```
Nadrieril, line 313:
```suggestion
    let field = &crate_data.types[2].kind.as_struct()[0];
    assert_eq!(
        field.item_meta.rename.as_deref(),
        Some("FieldTest")
    );
```

charon/tests/ui/unsupported/rename_attribute_failure.rs
```diff
@@ -0,0 +1,17 @@
+//@ known-failure
```
Nadrieril, line 1: The `unsupported` folder is for things we don't support yet but want to support eventually. Could you move this test out of the folder?

charon/src/translate/translate_ctx.rs
```diff
@@ -426,6 +426,62 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+                        .expect("Attribute `rename` should not be empty")
+                        .is_alphabetic();
+                    let is_identifier = first_char_alphabetic
+                        && str
+                            .chars()
+                            .all(|c| c.is_alphanumeric() || c == '_' || c == '-');
+                    if !is_identifier {
+                        self.span_err(span, "Attribute `rename` should only contains alphanumeric characters, `_` and `-` and should start with a letter");
+                        return None;
+                    } else {
```
Nadrieril, line 461: The point of asking you to use `return None;` is to avoid these nested `else` cases. The intent is that we can follow the code linearly where errors exit early and success continues down the flow of the function.

charon/src/translate/translate_ctx.rs
```diff
@@ -399,8 +399,8 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
     fn register_file(&mut self, filename: FileName) -> FileId {
         // Lookup the file if it was already registered
         match self.translated.file_to_id.get(&filename) {
-            Some(id) => *id,
-            None => {
+            Option::Some(id) => *id,
+            Option::None => {
```
Nadrieril, line 403: Please don't re-add these `Option::`; they were removed in a recent PR

charon/src/translate/translate_ctx.rs
```diff
@@ -426,6 +426,59 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+                );
+                return None;
+            }
+        // if there are more than one rename attribute, we tell the user there should only be one
+        } else if filter_attribute.len() > 1 {
+            self.span_err(
+                span,
+                "There are too many `rename` attributes, please use only one",
+            );
+            return None;
```
Nadrieril, line 476: This could be an early return

charon/src/translate/translate_ctx.rs
```diff
@@ -426,6 +426,59 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+                    self.span_err(span, "Attribute `rename` should only contains alphanumeric characters, `_` and `-` and should start with a letter");
+                    return None;
+                }
+                return Some(rename.unwrap().to_string());
+            } else {
+                self.span_err(
+                    span,
+                    "Attribute `rename` should be of the shape `rename(\"...\")`",
+                );
+                return None;
```
Nadrieril, line 468: This could be an early return

sonmarcho (approved):

PR #240, opened 2024-06-11
Use `Vector`s instead of `Generator`+`Map` pairs

Nadrieril:
This is a small API improvement: instead of generating ids with `Generator` and storing them in a `Map`, this uses `Vector` to do both at once. I tweaked `Vector` to be able to reserve slots in advance (by storing `None`) and fill them later.

Long-term I'd like to ensure we always store a value for each id (potentially an explicit error value) to avoid surprises.


charon/src/translate/translate_predicates.rs
```diff
@@ -380,7 +380,8 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
         hspan: &hax::Span,
         trait_pred: &hax::TraitPredicate,
     ) -> Result<Option<TraitClauseId>, Error> {
-        // FIXME: once `clause` can't be `None`, use a less error-prone `Vector` API.
+        // FIXME: once `clause` can't be `None`, use |Vector::reserve_slot` to be sure we don't use
```
sonmarcho, line 383: |Vector::reserve_slot`

Nadrieril: oops

PR #242, opened 2024-06-11
Globally index function and global bodies

### LINKED ISSUE #217: Remove excessive type genericity
I propose that instead of using type generics (gast, `GExprBody`, etc) to reuse code between ullbc and llbc, we replace `GExprBody` with:
```rust
enum BodyKind {
  Unstructured(Vector<BlockId, ullbc_ast::BlockData>),
  Structured(llbc_ast::Statement),
}
pub struct ExprBody {
    pub span: Span,
    // etc
    pub body: BodyKind,
}
```
Then ullbc->llbc conversion changes the enum variant from `Unstructured` to `Structured` and we can remove a ton of generic types. This has the additional benefit that if we accidentally run a pass on ullbc _after_ control flow restructuring we get a panic instead of silently modifying bodies we don't care about anymore.

Nadrieril:
Bodies are a big source of duplication between functions and globals. This PR does the following:
- Each body is now indexed by a `BodyId` and stored separately. `FunDecl` and `GlobalDecl` store a `BodyId` instead of a body;
- Because of that, `FunDecl` and `GlobalDecl` aren't type-generic anymore. There's a single definition used by both llbc and ullbc;
- The bodies are stored as `Vector<BodyId, Body>`, where `Body` is an enum with two variants: one for ullbc, one for llbc. When we reconstruct control flow, we change this variant.

All of this removes a decent amount of duplication and paves the way for more.

Fixes #217

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.7"
+let supported_charon_version = "0.1.8"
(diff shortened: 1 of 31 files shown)


Nadrieril:
Importantly, this is invisible for users of `charon-ml`

PR #247, opened 2024-06-12
fix(charon): remove useless dep. to `hax-frontend-exporter-options`

W95Psp:
This dependency is useless, and is blocking for
https://github.com/hacspec/hax/pull/713, which removes the crate `hax-frontend-exporter-options`. This commit thus just drops that dependency and cleans up the commented similar dependencies below.

### DIFF
--- a/charon/Cargo.toml
+++ b/charon/Cargo.toml
@@ -55,9 +55,7 @@ walkdir = "2.3.2"
 which = "6.0.1"
 
 hax-frontend-exporter = { git = "https://github.com/hacspec/hax", branch = "main" }
-hax-frontend-exporter-options = { git = "https://github.com/hacspec/hax", branch = "main" }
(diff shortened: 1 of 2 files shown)


charon/Cargo.toml
```diff
@@ -55,9 +55,6 @@ walkdir = "2.3.2"
 which = "6.0.1"
 
 hax-frontend-exporter = { git = "https://github.com/hacspec/hax", branch = "main" }
-hax-frontend-exporter-options = { git = "https://github.com/hacspec/hax", branch = "main" }
-# hax-frontend-exporter = { path = "../../hax/frontend/exporter" }
```
Nadrieril, line 59: Can you keep this one? I use it all the time to use my local hax instead of the locked one

W95Psp: Oh, sorry, yes sure! :)

Nadrieril (approved):

Issue #122, opened 2024-04-12
Use a `Vec` in `RawStatement::Sequence`


Nadrieril:
Currently it takes a pair of statements, which isn't very practical to pattern-match on. If it was a `Vec`, we could use slice patterns which should be a lot more convenient.

The drawback is we can't move out of a slice pattern, so we'll either have to clone or `insert`/`remove` on the `Vec`.

PR #260, opened 2024-06-17
Use the `derive-visitor` crate to replace our hand-written visitors

### LINKED ISSUE #249: Rework visitors

Nadrieril:
This uses the clever [derive_visitor](https://docs.rs/derive-visitor/latest/derive_visitor/index.html) crate instead of our handwritten visitors. I managed to replace all of our visitor needs with it.

This PR is more readable commit-by-commit. Fixes #249.


Nadrieril:
I'm a bit worried about the risk of stack overflow given that `derive-visitor` doesn't use `ensure_sufficient_stack`. I'm working on https://github.com/AeneasVerif/charon/issues/122 to remove the main source of runaway recursion.

sonmarcho (approved):
That's great!
If we have a stack-overflow issue with one instantiation of the visitors, we can manually override some of its (well-chosen) functions to use `ensure_sufficient_stack`, right?

Nadrieril:
Oh indeed, you're right! Let's merge then

PR #261, opened 2024-06-18
Make `RawStatement::Sequence` contain a `Vec`

### LINKED ISSUE #122: Use a `Vec` in `RawStatement::Sequence`

Nadrieril:
This is more convenient to use from rustc and avoids the high risk of overflow whenever we traverse a function body. It's likely a lot faster too. `charon-ml` users won't see the difference.

One surprising consequence is that this changes some loop spans in Aeneas tests. For a reason I don't understand, Aeneas usually had loop spans that pointed to the whole function instead of just the loop. Now it uses the span that is supposed to cover just the loop. Unfortunately spans are hard so they tend to be wrong anyway.

Fixes #122

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.10"
+let supported_charon_version = "0.1.11"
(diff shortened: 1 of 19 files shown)


sonmarcho (approved):
This looks good. I think I will update the Charon-ML to match this representation at some point.

Issue #127, opened 2024-04-16
Hide associated types equality constraints


Nadrieril:
The following is unsupported because we don't handle the equality constraint:
```rust
trait Iterator {
    type Item;
}

fn f<I, J>(x: I::Item) -> J::Item
where
    I: Iterator,
    J: Iterator<Item=I::Item>,
{ x }
```

The solution proposed by @sonmarcho would be to transform associated types into type parameters. Type equalities would then only apply to free type variables, which we can handle by merging any two variables that are equated. We would therefore generate:
```rust
trait Iterator<Item> {}

fn f<I, J, Item>(x: Item) -> Item
where
    I: Iterator<Item>,
    J: Iterator<Item>,
{ x }
```

sonmarcho:
Rephrasing a bit: we can extract the code above with Charon, and it's actually handled by Aeneas (which has a normalization mechanism) but the generated pure model doesn't type-check.

PR #262, opened 2024-06-19
Various improvements to trait clauses

Nadrieril:
This lays some groundwork towards #127. There's a few tweaks but mainly this PR:
- Removes some more leftover code from trait solving;
- Tracks for each `TraitClause` where it came from (function generics, associated type, etc).

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -335,6 +335,7 @@ generate_index_type!(TraitClauseId, "TraitClause");
 generate_index_type!(TraitDeclId, "TraitDecl");
 generate_index_type!(TraitImplId, "TraitImpl");
 
+/// A predicate of the form `Type: Trait<Args>`.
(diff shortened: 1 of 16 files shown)


charon/src/ast/types.rs
```diff
@@ -352,13 +345,62 @@ pub struct TraitClause {
...
+    /// The generics applied to the trait. Note: this includes the `Self` type.
     /// Remark: the trait refs list in the [generics] field should be empty.
     pub generics: GenericArgs,
 }
 
 impl Eq for TraitClause {}
 
+/// Where a given predicate came from.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Derivative, Drive, DriveMut)]
+pub enum PredicateOrigin {
```
sonmarcho, line 362: This is just additional useful information, correct?
Or do you need this somewhere?

Nadrieril: I don't need it now, but I expect to need it later

Issue #67, opened 2023-12-20
Add support for quantified where clauses (`for <'a>`)


sonmarcho:
For now, quantified where clauses like below are not supported:
```rust
where for<'a> Foo : Bar<'a>
```

PR #264, opened 2024-06-21
Merge `Predicates` into `GenericParams`

Nadrieril:
This PR merges two structs together. This reduces code duplication, as these two always came together in the AST. This is correct to do because predicates can appear exactly in the same places that clauses can, so it makes sense to group them together. This is part of my attempts to clarify how we deal with generics in preparation of #127 and #67 among others.

I will now prepare the changes to aeneas aud eurydice.

### DIFF
--- a/charon-ml/Makefile
+++ b/charon-ml/Makefile
@@ -11,7 +11,7 @@ build-dev:
 # You can use the environment variable "CHARON_LOG" to activate the log.
 # For instance: `CHARON_LOG=1 make tests`.
 .PHONY: tests
-tests: build copy-tests
+tests: copy-tests
 	dune test
 
 # Reformat the code
--- a/charon-ml/src/PrintUtils.ml
+++ b/charon-ml/src/PrintUtils.ml
@@ -26,27 +26,16 @@ type ('fun_body, 'global_body) fmt_env = {
           indices, i.e., the region var of index 0 doesn't have to be at index 0,
           etc. We lookup the variables by checking their id, not their position.
       *)
-  types : type_var list;
-  const_generics : const_generic_var list;
-  trait_clauses : trait_clause list;
-  preds : predicates;
+  generics : generic_params;
+      (* We ignore `generics.regions` as they are tracked in `regions` already *)
   locals : (VarId.id * string option) list;
       (** The local variables don't need to be ordered (same as the generics) *)
 }
 
 let fmt_env_update_generics_and_preds (env : ('a, 'b) fmt_env)
-    (generics : generic_params) (preds : predicates) : ('a, 'b) fmt_env =
-  let { regions; types; const_generics; trait_clauses } : generic_params =
-    generics
-  in
-  {
-    env with
-    regions = regions :: env.regions;
-    types;
-    const_generics;
-    trait_clauses;
-    preds;
-  }
+    (generics : generic_params) : ('a, 'b) fmt_env =
+  let { regions; _ } : generic_params = generics in
+  { env with regions = regions :: env.regions; generics }
 
 let fmt_env_push_regions (env : ('a, 'b) fmt_env) (regions : region_var list) :
     ('a, 'b) fmt_env =
--- a/charon-ml/src/Substitute.ml
+++ b/charon-ml/src/Substitute.ml
@@ -49,17 +49,7 @@ let st_substitute_visitor (subst : subst) =
       TArrow (regions, inputs, output)
 
     method! visit_TVar (subst : subst) id = subst.ty_subst id
-
-    method! visit_type_var_id _ _ =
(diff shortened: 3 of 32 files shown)


charon-ml/Makefile
```diff
@@ -11,7 +11,7 @@ build-dev:
 # You can use the environment variable "CHARON_LOG" to activate the log.
 # For instance: `CHARON_LOG=1 make tests`.
 .PHONY: tests
-tests: build copy-tests
```
Nadrieril, line 14: This `build` requirement meant we built the project twice every time

charon-ml/src/PrintUtils.ml
```diff
@@ -26,27 +26,16 @@ type ('fun_body, 'global_body) fmt_env = {
           indices, i.e., the region var of index 0 doesn't have to be at index 0,
           etc. We lookup the variables by checking their id, not their position.
       *)
-  types : type_var list;
-  const_generics : const_generic_var list;
-  trait_clauses : trait_clause list;
-  preds : predicates;
+  generics : generic_params;
+      (* We ignore `generics.regions` as they are tracked in `regions` already *)
```
Nadrieril, line 30: I'm tempted to also remove `regions` and store a stack of `generic_params` instead. This might be a bit too general but I've heard rumors of the new trait solver being able to support `for<T: Clone> X: Trait<Item = T>` kinds of bounds, which would require an actual stack of `generic_params`.

charon-ml/src/Substitute.ml
```diff
@@ -49,17 +49,7 @@ let st_substitute_visitor (subst : subst) =
...
-    method! visit_type_var_id _ _ =
-      (* We should never get here because we reimplemented [visit_TypeVar] *)
-      raise (Failure "Unexpected")
-
     method! visit_CgVar _ id = subst.cg_subst id
-
-    method! visit_const_generic_var_id _ _ =
-      (* We should never get here because we reimplemented [visit_Var] *)
-      raise (Failure "Unexpected")
-
```
Nadrieril, line 62: I had to remove these because `generic_params` stores some `type_var`s which aren't inside `Ty::TVar` and thus they fail here. I'm not sure if we want to substitute on a whole `generic_params` and what the output of that should be. I'm thinking to turn `GenericParams` into a list of an enum, and then we can restrict substitution to only predicates again.

charon/tests/ui/issue-45-misc.out
```diff
@@ -144,6 +144,8 @@ trait core::iter::traits::iterator::Iterator<Self>
 }
 
 trait core::iter::traits::collect::IntoIterator<Self>
+where
+    (Self::IntoIter::[@TraitClause0])::Item = Self::Item,
```
Nadrieril, line 148: This change is because we used to do `ctx.get_predicates()` before exploring trait associated items, and we never tracked item-specific predicates so we forgot this one.

Nadrieril:
The aeneas and eurydice PRs are ready

charon-ml/src/PrintUtils.ml
sonmarcho: It would make sense I think.

charon-ml/src/Substitute.ml
sonmarcho: I don't get this: why did the behavior change (how come there are type vars which are not insdie `Ty::TVar`?
On the side of Aeneas, I always pay attention to substituting a whole `generic_params` at once. If we get there it is a *bug* (we should put this kind of checks back).

charon-ml/src/PrintUtils.ml
Nadrieril: I beg to differ, any simplification of this whole thing is a huge win at this point

charon-ml/src/Substitute.ml
Nadrieril: There are type vars in `GenericParams.types`

Nadrieril: And it would make no sense to substitute them, unless we were turning this into a `GenericArgs` (but we'd lose the predicates).

charon-ml/src/PrintUtils.ml
Nadrieril: I'm about to add methods to match up and substitute `GenericParams` and `GenericArgs`. If we unify here we'd be able to reuse all these methods and checks.

charon-ml/src/Substitute.ml
Nadrieril: I've added those checks back in #271

charon-ml/src/PrintUtils.ml
Nadrieril: On further thought, I think we will need this to handle GATs too.

PR #266, opened 2024-06-24
Cleanup `ItemMeta` and attributes

Nadrieril:
https://github.com/AeneasVerif/charon/pull/237 introduced a new attribute and reused `ItemMeta` in fields and variants. This Pr splits off `ItemMeta` so that fields and variants can have attribute information and items can have item-specific information in `ItemMeta`. I then move some shared item fields to `ItemMeta`. This also parses attributes into a high-level representation.

### DIFF
--- a/charon/src/ast/meta.rs
+++ b/charon/src/ast/meta.rs
@@ -130,14 +130,21 @@ pub enum Attribute {
     Unknown(String),
 }
 
-/// Meta information about an item (function, trait decl, trait impl, type decl, global).
+/// Information about the attributes and visibility of an item, field or variant..
 #[derive(Debug, Clone, Serialize, Deserialize, Drive, DriveMut)]
-pub struct ItemMeta {
-    pub span: Span,
+pub struct AttrInfo {
     /// Attributes (`#[...]`).
     pub attributes: Vec<Attribute>,
     /// Inline hints (on functions only).
     pub inline: Option<InlineAttr>,
+    /// Attribute `charon::rename("...")` provide a custom name that can be used by consumers of
+    /// llbc. E.g. Aeneas uses this to rename definitions in the extracted code.
+    pub rename: Option<String>,
+    /// Whether there was a `#[charon::opaque]` annotation on this item. This is different from
+    /// other reasons this may be opaque.
+    #[serde(skip_serializing)]
+    #[serde(default)]
+    pub opaque: bool,
     /// Whether this item is declared public. Impl blocks and closures don't have visibility
     /// modifiers; we arbitrarily set this to `false` for them.
     ///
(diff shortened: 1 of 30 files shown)


Nadrieril:
cc @sonmarcho @msprotz This changes where we store the name of an item: `decl.name` has to be replaced with `decl.item_meta.name` now. (I'll do the necessary changes in Aeneas and Eurydice of course).

charon/src/ast/meta.rs
```diff
@@ -113,14 +111,41 @@ pub enum InlineAttr {
...
+pub struct AttrInfo {
     /// Attributes (`#[...]`).
     pub attributes: Vec<Attribute>,
     /// Inline hints (on functions only).
     pub inline: Option<InlineAttr>,
+    /// Attribute `charon::rename("...")` provide a custom name that can be used by consumers of
+    /// llbc. E.g. Aeneas uses this to rename definitions in the extracted code.
+    pub rename: Option<String>,
+    /// Whether there was a `#[charon::opaque]` annotation on this item. This is different from
+    /// other reasons this may be opaque.
```
sonmarcho, line 145: I don't understand the second sentence.

charon/src/ast/meta.rs
```diff
@@ -113,14 +111,41 @@ pub enum InlineAttr {
...
+/// Information about the attributes and visibility of an item, field or variant..
 #[derive(Debug, Clone, Serialize, Deserialize, Drive, DriveMut)]
-pub struct ItemMeta {
-    pub span: Span,
+pub struct AttrInfo {
     /// Attributes (`#[...]`).
     pub attributes: Vec<Attribute>,
     /// Inline hints (on functions only).
     pub inline: Option<InlineAttr>,
+    /// Attribute `charon::rename("...")` provide a custom name that can be used by consumers of
```
sonmarcho, line 141:
```suggestion
    /// Attribute `charon::rename("...")` provides a custom name that can be used by consumers of
```

charon/src/ast/meta.rs
```diff
@@ -140,12 +165,17 @@ pub struct ItemMeta {
...
-    #[serde(skip_serializing)]
-    #[serde(default)]
-    pub opaque: bool,
+}
+
+/// Meta information about an item (function, trait decl, trait impl, type decl, global).
+#[derive(Debug, Clone, Serialize, Deserialize, Drive, DriveMut)]
+pub struct ItemMeta {
+    pub span: Span,
+    pub name: Name,
```
sonmarcho, line 174: Do you really want to put this here? I understand that we may want to factor the name out. But is it really "meta" information?

Nadrieril: :shrug: clonse enough for me, I don't think this introduces ambiguity. The name feels definitely more meta than the signature or the body.

charon/src/ast/meta.rs
Nadrieril: Should I say "an item may be opaque for other reasons than this attribute, e.g. if specified on the command line"?

PR #272, opened 2024-06-26
Mad rustc update

### LINKED ISSUE #182: Update rustc version
This mostly depends on https://github.com/hacspec/hax/issues/224.

Nadrieril:
Companion PR to https://github.com/hacspec/hax/pull/735. This updates the rustc pin by about 8 months.

Closes #182.

### DIFF
--- a/charon/tests/ui/issue-120-bare-discriminant-read.out
+++ b/charon/tests/ui/issue-120-bare-discriminant-read.out
@@ -9,21 +9,17 @@ fn test_crate::call_is_some<T>(@1: core::option::Option<T>) -> bool
 {
     let @0: bool; // return
     let opt@1: core::option::Option<T>; // arg #1
-    let self@2: &'_ (core::option::Option<T>); // local
-    let @3: isize; // anonymous local
+    let @2: isize; // anonymous local
 
-    self@2 := &opt@1
     match opt@1 {
-        0 => {
-            @3 := const (0 : isize)
-        },
         1 => {
-            @3 := const (1 : isize)
+            @0 := const (true)
+        },
+        0 => {
+            @0 := const (false)
         }
     }
-    @0 := copy (@3) == const (1 : isize)
-    drop @3
-    drop self@2
+    drop @2
     drop opt@1
     return
 }
--- a/charon/tests/ui/matches.out
+++ b/charon/tests/ui/matches.out
@@ -13,14 +13,19 @@ fn test_crate::test1(@1: test_crate::E1) -> bool
 
     @fake_read(x@1)
     match x@1 {
-        0 => {
+        0 | 1 => {
             @0 := const (true)
         },
-        1 => {
-            @0 := const (true)
-        },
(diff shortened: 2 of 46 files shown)


charon/tests/ui/matches.out
```diff
@@ -13,14 +13,19 @@ fn test_crate::test1(@1: test_crate::E1) -> bool
...
+        _ => {
+            match x@1 {
+                2 => {
+                    @0 := const (false)
+                },
+                _ => {
+                    @fake_read(x@1)
+                    undefined_behavior
+                }
+            }
```
Nadrieril, line 28: Interesting, not sure what changed here. I did make _some_ changes to the MIR-lowering of or-patterns but they shouldn't have affected a simple case like this.

charon/tests/ui/constants.out
```diff
@@ -223,8 +223,8 @@ fn test_crate::get_z2() -> i32
     let @0: i32; // return
     let @1: i32; // anonymous local
     let @2: i32; // anonymous local
-    let @3: i32; // anonymous local
-    let @4: i32; // anonymous local
+    let @3: fn() -> i32; // anonymous local
+    let @4: fn() -> i32; // anonymous local
```
Nadrieril, line 227: I have _no idea_ what's happening here. It seems that every local that is assigned a constant value now things it's a function type?

Nadrieril: Oh lord, this type is the same type as the enclosing function (reproducibly), and comes from `ConstantExpr.ty`

charon/tests/ui/matches.out
Nadrieril: This was caused by a bug in MIR generation in rustc. The fix is here: https://github.com/rust-lang/rust/pull/127028.

charon/tests/ui/issue-120-bare-discriminant-read.out
```diff
@@ -9,21 +9,17 @@ fn test_crate::call_is_some<T>(@1: core::option::Option<T>) -> bool
...
-        0 => {
-            @3 := const (0 : isize)
-        },
         1 => {
-            @3 := const (1 : isize)
+            @0 := const (true)
+        },
+        0 => {
+            @0 := const (false)
         }
```
sonmarcho, line 20: I'm surprised that the order changed here (case 1 then case 0)

Nadrieril: I think it changed in the MIR (and this is likely my doing again).

Nadrieril: Oh wait, that's code that we generate ourselves. We generate it in the order of the variants in the type definition, maybe that changed? That _is_ weird.

sonmarcho: I opened an issue https://github.com/AeneasVerif/charon/issues/278

PR #277, opened 2024-06-30
Fix an issue with inline_local_panic_functions

sonmarcho:
The micro-pass which inlines the panic functions did not filter those functions from the declaration groups.


Nadrieril:
We could have also run `reorder_decl` after that pass instead of before

PR #281, opened 2024-07-02
Use upstream libcrux flake in CI

Nadrieril:
Now that libcrux has proper flake support, we can use their flake in CI instead of our outdated copy-paste of `c.sh`.


Nadrieril:
Thanks @pnmadelaine for that work!

PR #282, opened 2024-07-03
Run `reorder_decls` after the micro-passes

Nadrieril:
This makes `reorder_decls` more robust to missing items, which makes it possible to run after our micro-passes. This makes our graph computation reflect more accurately the final structure of the code. This removes the need for post-hoc alterations of the declaration groups like in https://github.com/AeneasVerif/charon/pull/277.


sonmarcho:
Looks good! And I actually hesitated to do the same.

sonmarcho (approved):

sonmarcho:
However it seems that it reordered some definitions, leading to some diff in the files generated by Aeneas: this is slightly suspicious.

PR #283, opened 2024-07-03
Generate `GAstOfJson.ml` from the rust definitions

Nadrieril:
This PR adds a rust test that runs charon on itself and uses the resulting llbc to generate most of the `*_of_json` functions in `GAstOfJson.ml`. Eventually we should also generate the other files, including the type definitions, but this is a good start. I aimed to make it change the existing file as little as possible.

### DIFF
--- a/charon/src/lib.rs
+++ b/charon/src/lib.rs
@@ -24,6 +24,9 @@
 #![feature(let_chains)]
 #![feature(lint_reasons)]
 #![feature(trait_alias)]
+#![feature(register_tool)]
+// For when we use charon on itself :3
+#![register_tool(charon)]
 
 extern crate rustc_ast;
 extern crate rustc_ast_pretty;
--- a/charon/tests/generate-ml.rs
+++ b/charon/tests/generate-ml.rs
@@ -1,0 +1,467 @@
+#![cfg(test)]
+//! Generate ocaml deserialization code for our types.
+//!
+//! This test runs charon on itself and generates the appropriate `<type>_of_json` functions for
+//! our types. The generated functions are inserted into `./generated-ml/GAstOfJson.template.ml` to
+//! construct the final `GAstOfJson.ml`.
+//!
+//! Note: this test is ignored by default because running charon on itself is slow. To run it, call
+//! `cargo test --test generate-ml -- --ignored`. It is also run by `make generate-ml` in the crate
+//! root. Don't forget to format the output code after regenerating.
+use anyhow::{bail, Result};
+use assert_cmd::cargo::CommandCargoExt;
+use charon_lib::export::CrateData;
+use charon_lib::meta::ItemMeta;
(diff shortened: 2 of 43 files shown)


sonmarcho:
For information, the command which regenerates the ML files is `make generate-ml`.

charon/src/lib.rs
```diff
@@ -24,6 +24,9 @@
 #![feature(let_chains)]
 #![feature(lint_reasons)]
 #![feature(trait_alias)]
+#![feature(register_tool)]
+// For when we use charon on itself :3
```
sonmarcho, line 28: I prefer avoiding smileys in the code... But smileys in the github conversations are of course ok. ;)

charon/tests/generated-ml/GAstOfJson.template.ml
```diff
@@ -0,0 +1,343 @@
...
+    `GAstOfJson` by hand. *)
+
+(** Functions to load (U)LLBC ASTs from json.
+
+    Initially, we used [ppx_derive_yojson] to automate this.
+    However, [ppx_derive_yojson] expects formatting to be slightly
+    different from what [serde_rs] generates (because it uses [Yojson.Safe.t]
+    and not [Yojson.Basic.t]).
+
+    TODO: we should check all that the integer values are in the proper range
```
sonmarcho, line 15: This TODO is obsolete

charon/tests/generated-ml/GAstOfJson.template.ml
```diff
@@ -0,0 +1,343 @@
+(** WARNING: this file is partially auto-generated. `GAstOfJson.template.ml`
```
sonmarcho, line 1: This comment is slightly confusing. My understanding is that it should be read from the `GAstOfJson.ml` file, but when we read it in the `.template` file it is not clear whether it is ok to modify this file or not.

charon/tests/generate-ml.rs
```diff
@@ -0,0 +1,467 @@
...
+        if long_name.starts_with("charon_lib") {
+            let short_name = ty.item_meta.name.name.last().unwrap().as_ident().0.clone();
+            name_to_type.insert(short_name, ty);
+        }
+        name_to_type.insert(long_name, ty);
+    }
+    let raw_span = name_to_type.get("RawSpan").unwrap();
+    ctx.contains_raw_span = contains_id(&ctx.crate_data, raw_span.def_id);
+
+    let names: &[&[&str]] = &[
```
sonmarcho, line 375: Would be good to add a comment here to explain that we will extract those groups of definitions in the order listed below.

charon/tests/generate-ml.rs
```diff
@@ -0,0 +1,467 @@
+#![cfg(test)]
```
sonmarcho, line 1: From my point of view this file is not really a test (though it also acts as a test).

PR #285, opened 2024-07-05
Cleanup handling of trait generics

Nadrieril:
This PR cleanups our handling of generics somewhat:
- Adds a pass that ensures all `GenericArgs` match the corresponding `GenericParams`;
- Moves the generic arguments in `TraitRef` to be close to the id they apply to (also avoids creating empty `GenericArgs` in cases where non is needed);
- Uses `TraitDeclRef` anywhere a trait is referenced alongside generic arguments;
- The `trait_instance_id.TraitRef` variant on the ml side was a hack and is no longer needed.

In theory, all occurrences of `TraitDeclId` in the ast could come with appropriate generic arguments, but this is not yet needed and hax does not give us the necessary information yet (the tricky place is in nested trait references, e.g. parent clauses of a where-clause in scope).

### DIFF
--- a/charon-ml/src/Types.ml
+++ b/charon-ml/src/Types.ml
@@ -320,18 +320,6 @@ and trait_instance_id =
   | ParentClause of trait_instance_id * trait_decl_id * trait_clause_id
   | ItemClause of
       trait_instance_id * trait_decl_id * trait_item_name * trait_clause_id
-  | TraitRef of trait_ref
-      (** Not present in the Rust version of Charon.
-
-          We need this case for instantiations: when calling a function which has
-          trait clauses, for instance, we substitute the clauses refernced in the
-          [Clause] and [Self] case with trait references.
-
-          Remark: something potentially confusing is that [trait_clause_id] is used for
-          different purposes. In the [Clause] case, a trait clause id identifies a local
-          trait clause (which can thus be substituted). In the other cases, it references
-          a sub-clause relative to a trait instance id.
-       *)
   | FnPointer of ty
   | Closure of fun_decl_id * generic_args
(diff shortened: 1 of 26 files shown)


charon-ml/src/Types.ml
```diff
@@ -315,27 +314,15 @@ and generic_args = {
 and trait_instance_id =
```
sonmarcho, line 314: I think we should keep the names consistent between the Rust code and the ML code: we should rename `trait_instance_id` (shouldn't be more than a `sed` right?).

charon/src/ast/types_utils.rs
```diff
@@ -102,6 +102,14 @@ impl GenericArgs {
         }
     }
 
+    /// Check whether this matches the given `GenericParams`.
+    pub fn matches(&self, params: &GenericParams) -> bool {
```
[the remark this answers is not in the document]
Nadrieril: The intent was to add better checks as time goes, e.g. to check that trait refs are for the appropriate trait. I'll add a `// TODO` next time I touch this

charon-ml/src/Types.ml
Nadrieril: I'm waiting for the dust to settle first: as you pointed out this may not be the best name, and I expect other naming will have to change.

Nadrieril: Opened https://github.com/AeneasVerif/charon/issues/287

charon/src/ast/types_utils.rs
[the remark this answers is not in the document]
sonmarcho: Makes sense

PR #289, opened 2024-07-16
Adds small helpers in llbcastutils and document raw_statement a bit

EschericHya:


charon/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "charon"
-version = "0.1.20"
+version = "0.1.21"
```
Nadrieril, line 3: This is not necessary here. The CI check will complain, but it's a false positive since adding helpers cannot break users of charon-ml.

charon-ml/src/LlbcAstUtils.ml
```diff
@@ -1,6 +1,18 @@
...
+let fun_decl_list_from_crate (crate : crate) : fun_decl list =
+  snd (List.split (FunDeclId.Map.bindings crate.fun_decls))
+
+let get_fun_args (fun_decl : fun_decl) : var list =
+  match fun_decl.body with
+  | Some body ->
+      let input_number = body.arg_count in
+      let input_list = snd (List.split_at body.locals 1) in
+      fst (List.split_at input_list input_number)
+  | None -> []
```
Nadrieril, line 15: This is misleading: a function without a body is not the same thing as a function that takes no arguments; this could cause issues. Please either return `var list option` or operate on a `fun_body`.

charon-ml/src/LlbcAstUtils.ml
```diff
@@ -1,6 +1,18 @@
 include GAstUtils
 open LlbcAst
 open Utils
+open Collections
+
+let fun_decl_list_from_crate (crate : crate) : fun_decl list =
+  snd (List.split (FunDeclId.Map.bindings crate.fun_decls))
+
+let get_fun_args (fun_decl : fun_decl) : var list =
```
Nadrieril, line 9: Please document what this function does

Nadrieril: Could you also clarify the docs of `GExprBody` so we know in which order the arguments go in `locals`?

Nadrieril (changes requested):

charon/src/ast/llbc_ast.rs
```diff
@@ -27,8 +27,14 @@ pub struct Assert {
     Debug, Clone, EnumIsA, EnumToGetters, EnumAsGetters, Serialize, Deserialize, Drive, DriveMut,
 )]
 pub enum RawStatement {
+    /// Assigns an `Rvalue` to a `Place`. e.g. `let y = x` becomes
+    /// `y := copy x` which looks like
+    /// `"Assign":(("var_id":2,"projection":[]),("Use":("Copy":("var_id":1,
+    /// "projection":[]))))` in llbc
```
Nadrieril, line 33:
```suggestion
    /// Assigns an `Rvalue` to a `Place`. e.g. `let y = x;` could become
    /// `y := move x` which is represented as `Assign(y, Rvalue::Use(Operand::Move(x)))`.
```

charon/src/ast/llbc_ast.rs
```diff
@@ -79,7 +87,10 @@ pub struct Statement {
     VariantIndexArity,
 )]
 pub enum Switch {
-    /// Gives the `if` block and the `else` block
+    /// Gives the `if` block and the `else` block. The `Operand` is the "body" of the `if`, e.g. `if (y == 0)` becomes
+    /// `v@3 := copy y^1;
+    ///  v@2 := move v@3 == (0: u32 : u32);
+    ///  if (move v@2) {`
```
Nadrieril, line 93:
```suggestion
    /// Gives the `if` block and the `else` block. The `Operand` is the condition of the `if`, e.g. `if (y == 0)` could become
    /// ```
    /// v@3 := copy y; // Represented as `Assign(v@3, Use(Copy(y))`
    /// v@2 := move v@3 == 0; // Represented as `Assign(v@2, BinOp(BinOp::Eq, Move(y), Const(0)))`
    /// if (move v@2) { // Represented as `If(Move(v@2), <then branch>, <else branch>)`
    /// ```
```

charon/src/ast/llbc_ast.rs
```diff
@@ -79,7 +87,12 @@ pub struct Statement {
     VariantIndexArity,
 )]
 pub enum Switch {
-    /// Gives the `if` block and the `else` block
+    /// Gives the `if` block and the `else` block. The `Operand` is the condition of the `if`, e.g. `if (y == 0)` could become
+    /// ```
```
Nadrieril, line 91:
```suggestion
    /// ```rust,ignore
```

Nadrieril (approved):

PR #290, opened 2024-07-16
Lift clauses on associated types to be clauses on the whole trait

Nadrieril:
Dinstinguishing trait clauses defined on an associated type versus those defined as a where clause on the whole trait does not seem useful. What's meaningful is distinguishing clauses that are required to mention the trait versus clauses that are implied by the trait. We already make this distinction: required clauses are in the `TraitDecl.generics` field, implied clauses are in `TraitDecl.parent_clauses`.

This PR moves all the `type Item: Trait` clauses to be `where Self::Item: Trait` parent clauses on the trait to simplify clause management. This is done as a post-processing pass. I would have liked to to this when we translate clauses, but I'll leave that for later when generics are easier to follow.

### DIFF
--- a/charon-ml/src/GAst.ml
+++ b/charon-ml/src/GAst.ml
@@ -159,7 +159,7 @@ type trait_decl = {
   generics : generic_params;
   parent_clauses : trait_clause list;
   consts : (trait_item_name * (ty * global_decl_id option)) list;
-  types : (trait_item_name * (trait_clause list * ty option)) list;
+  types : (trait_item_name * ty option) list;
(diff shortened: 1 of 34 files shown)


Nadrieril:
For the CI checks:
- check-version-number failure is a false positive because the json format hasn't changed;
- aeneas is indeed failing, which is fixed in [the companion PR](https://github.com/AeneasVerif/aeneas/pull/288);
- kyber has a build issue since the last eurydice update, which is independent from charon. This *might* be hiding breakage caused by this PR, but if so it should only be some easy-to-fix name changes.

PR #291, opened 2024-07-16
Fix CI flake overrides

Nadrieril:
We were victims of the non-transitivity of flake overrides. This should work better.


Nadrieril:
It appears that libcrux has copy-pasted some karamel files into its repo, hence using the eurydice-compatible version of karamel is not enough. I'll let them sort that out.

PR #294, opened 2024-07-22
Store `TraitImplId` in `Name` instead of reproducing generics by hand

Nadrieril:
This is important for #127, as otherwise we don't have enough information in the names to ensure they call traits with the appropriate number of parameters. This also removes one source of generics complexity.

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_ctx.rs
+++ b/charon/src/bin/charon-driver/translate/translate_ctx.rs
@@ -319,61 +319,60 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
                     name.push(PathElem::Ident(crate_name.clone(), disambiguator));
                 }
                 DefPathData::Impl => {
-                    // We need to convert the type, which may contain quantified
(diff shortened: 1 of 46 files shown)


charon/src/bin/charon-driver/translate/translate_ctx.rs
```diff
@@ -319,61 +319,60 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
-                        Some(trait_ref) => {
-                            // Trait implementation
-                            let trait_ref = trait_ref.sinto(&bt_ctx.hax_state);
-                            let erase_regions = false;
-                            let trait_ref =
-                                bt_ctx.translate_trait_decl_ref(span, erase_regions, &trait_ref)?;
-                            match trait_ref {
+                        Some(_) => {
+                            let impl_id = self.register_trait_impl_id(&None, cur_id)?;
+                            // // Trait implementation
```
sonmarcho, line 361: Clean this up

Nadrieril: Oops

PR #295, opened 2024-07-24
Cleanup predicate handling again

Nadrieril:
Simplify predicate handling again. `ClauseTransCtx` was quite confusing, so I replaced it with a simpler enum passed in function arguments and straightforward vectors of clauses in `BodyTransCtx`. I also discovered that we no longer need to use `predicates_of` to get some of the predicates we care about: `predicates_defined_on` now returns all the predicates we need.


sonmarcho:
Following a conversation on Zulip, I think `predicates_defined_of` *doesn't* give us all the predicates we need, right?

PR #296, opened 2024-07-25
Upstream many queries to hax

Nadrieril:
Companion PR to https://github.com/hacspec/hax/pull/803. This uses the new `hax::FullDef` to greatly simplify our query usage.


Nadrieril:
The kyber failure is independent. It passed with this PR quite recently; the changes to this PR since then are unlikelyto have broken it.

Issue #215, opened 2024-05-30
Automatically generate json parsing


Nadrieril:
It's tedious to write them by hand. I found https://github.com/serde-ml/serde but it requires ocaml `>= 5.1.1`. We could write our own generator from rust possibly.

sonmarcho:
Does it print nice error messages when it fails?

Nadrieril:
I don't know I couldn't try because of ocaml version mismatches

Nadrieril:
First big chunk was done here: https://github.com/AeneasVerif/charon/pull/283

Nadrieril:
This is now done

PR #302, opened 2024-08-06
Generate some ml type definitions

Nadrieril:
Progress on #215. This auto-generates a number of ml definitions from the rust ones. I've left the visitor definitions as-is; that feels incomplete, I hope to find a cleaner solution when more of the types are auto-generated.

Nice bonus: this copies the doc-comments present on the rust side.


charon-ml/src/Expressions.ml
```diff
@@ -1,27 +1,75 @@
+(** WARNING: this file is partially auto-generated. Do not edit `src/Expressions.ml`
+    by hand. Edit `templaces/Expressions.ml` instead, or improve the code
```
sonmarcho, line 2: "templaces"

charon-ml/src/Expressions.ml
```diff
@@ -1,27 +1,75 @@
...
 type var_id = VarId.id [@@deriving show, ord]
 
+(** An assumed function identifier, identifying a function coming from a
+ standard library. *)
 type assumed_fun_id =
-  | BoxNew
+  | BoxNew  (** `alloc::boxed::Box::new` *)
   | BoxFree
+      (** `alloc::alloc::box_free`
+ This is actually an unsafe function, but the rust compiler sometimes
```
sonmarcho, line 29: We should add indentation in multi-line comments

sonmarcho:
I hadn't thought of the fact that doing this would allow to port the documentation from the Rust side to the OCaml side: this is extremely useful. The only minor change I would like to do is a better indentation of the multi-line comments (as you can see, having no indentation is a bit ugly).

charon-ml/src/Expressions.ml
Nadrieril: The `--wrap-comments` ocamlformat option doesn't work on doc-comments and destroys the formatting of e.g. code blocks

Nadrieril: I can still guess the indent level, that'd be better than the current state.

Nadrieril: This worked great (https://github.com/AeneasVerif/charon/pull/307)

PR #310, opened 2024-08-14
Cache `Name` computation

Nadrieril:
We compute names _a lot_ (e.g. every time we translate a type, to check if it refers to a built-in). This makes a significant difference iun debug mode (~30% speedup when running `make generate-ml`), not so much when charon is compiled in release mode.


sonmarcho:
Oh wow, I hadn't realized it was that expensive :scream:

PR #311, opened 2024-08-14
Ask rustc to recognize built-ins

Nadrieril:
Rustc has a mechanism to recognize specific items: lang items and diagnostic items. For most of our built-ins we can use these mechanisms instead of matching on names.


sonmarcho:
That's really cool!

PR #313, opened 2024-08-14
Reuse `TranslatedCrate` in json export

Nadrieril:
This reduces duplication, allows consumers to make use of convenient `TranslatedCrate` APIs, and makes the version detection more robust.


charon/src/ast/krate.rs
```diff
@@ -230,3 +233,26 @@ impl<'tcx, 'ctx, 'a> IntoFormatter for &'a TranslatedCrate {
...
+            }
+        }
+    };
+}
+mk_index_impls!(TranslatedCrate.type_decls[TypeDeclId]: TypeDecl);
+mk_index_impls!(TranslatedCrate.fun_decls[FunDeclId]: FunDecl);
+mk_index_impls!(TranslatedCrate.global_decls[GlobalDeclId]: GlobalDecl);
+mk_index_impls!(TranslatedCrate.bodies[BodyId]: Body);
+mk_index_impls!(TranslatedCrate.trait_decls[TraitDeclId]: TraitDecl);
+mk_index_impls!(TranslatedCrate.trait_impls[TraitImplId]: TraitImpl);
```
sonmarcho, line 258: This is fun :)

Issue #214, opened 2024-05-30
Support extracting specific foreign items


Nadrieril:
`--extract-opaque-bodies` extracts all the foreign definitions it can. It would be useful to be able to specify only specific ones. E.g. in `Charon.toml`:
```toml
[charon]
extract = [
  "trait core::convert::Into",
  "fn core::convert::{impl core::convert::Into<@U> for @T}::into",
]
```
This would extract the `Into` trait and the blanket `From => Into` implementation, which could be useful if the crate defines any `From` impls.

msprotz:
FYI I tried this just now and I got this:
```
[Charon__GAstOfJson.type_decl_kind_of_json] failed on: `Assoc ([("Error",
          `String ("Unsupported type: foreign type: core::fmt::rt::{extern#0}::Opaque"))
          ])
```

to reproduce, add `--extract-opaque-bodies` to $(CHARON) in eurydice/Makefile, then run `make test-where_clauses_closure`.

Nadrieril:
With https://github.com/AeneasVerif/charon/pull/316, `charon --include "core::convert::{impl core::convert::Into<_> for _}::into` will translate precisely the requested method. This avoids the issue of `--extract-opaque-bodies` extracting too much.

PR #316, opened 2024-08-19
Add precise filtering of which items to translate

Nadrieril:
This PR adds a rust implementation of something like the `NameMatcher`. This makes it possible to precisely filter items precisely, e.g. translate exactly one impl from the standard library. This fixes the issue of `--extract-opaque-bodies` triggering a slurry of errors. This is still a bit rough but already useful.

Fixes https://github.com/AeneasVerif/charon/issues/214


charon-ml/src/LlbcOfJson.ml
```diff
@@ -181,43 +181,42 @@ let split_global (gid_conv : global_id_converter) global :
   (global_decl, fun_decl)
 
 let crate_of_json (js : json) : (crate, string) result =
-  combine_error_msgs js __FUNCTION__
```
sonmarcho, line 184: The `combine_error_msgs` is not used anymore?

charon/src/name_matcher/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+enum PatElem {
+    /// An identifier, optionally with generic arguments. E.g. `std` or `Box<_>`.
+    Ident {
+        name: String,
+        generics: Vec<PatTy>,
+        /// For pretty-printing only: whether this is the name of a trait.
+        is_trait: bool,
```
sonmarcho, line 24: Is this only used for pretty printing, or also for matching?
For the Charon-ML side: we should have the same, but should also use it for matching (if `true`, we can only match a trait impl).

sonmarcho:
It's super nice to finally have this, thanks :)

charon-ml/src/LlbcOfJson.ml
Nadrieril: This is only for the version mismatch error: we don't want to dump the parsed AST in that case.

charon/src/name_matcher/mod.rs
Nadrieril: I'm not sure how useful that is: there cannot be a trait with the same name as a type. The fact that we ignore this means we can support both the `impl Trait<..> for Self` syntax and the `Trait<Self, ..>` syntaxes.

Nadrieril: Well, I guess with wildcards there can be :thinking:

sonmarcho: Actually you're probably right: it may be useless. But I was mostly thinking about consistency/sanity...

PR #318, opened 2024-08-20
Add `--exclude` flag

Nadrieril:
This complements `--include` and `--exclude`: `--exclude` means "translate the signature of the item but not its contents". When even the signature is something we can't handle, we can now use `--hide` to skip its translation entirely.


sonmarcho:
I actually think `--opaque` would make more sense than `--exclude`, and `--exclude` should have the current meaning of `--hide`. What do you think?

PR #327, opened 2024-08-21
Extract the source text of each item

### LINKED ISSUE #317: Store the source text of every item in the llbc output

Nadrieril:
Fixes #317. This also changes item spans to cover the whole item, which is what I had always expected they would do.


sonmarcho:
The problem with the spans have been annoying me for a while :)
And good for the source text! I opened a subsequent issue to use it in the comments of the generated code (I will take care of that).

PR #334, opened 2024-08-27
Add stack space for recursive functions over a body

Nadrieril:
These were discovered to cause stack overflow on functions with complex control flow.


sonmarcho:
I'm not surprised: this is not the first time that I get a stack overflow because of the reconstruction of control-flow. (thanks!)

PR #337, opened 2024-08-28
Support almost all MIR rvalues

### LINKED ISSUE #332: Support raw pointer cast and floats
**Code snippet that is not correctly supported**:

```rust
fn test_cast (x : [u8; 64]) -> u8 {
    x.as_ptr() as u8
}
```

**Current Charon output**:

```
Compilation failed: error: Unsupported cast:
       
       - rvalue: Cast(PointerExposeProvenance, Move(Place { ty: RawPtr(Uint(U8), false), kind: Local(Local(2)) }), Uint(U8))
       
       - src=RawPtr(Literal(Integer(U8)), Shared)
       
       - dst=Literal(Integer(U8))
   --> tests/ui/taint.rs:165:5
    |
    |     x.as_ptr() as u8
    |     ^^^^^^^^^^^^^^^^

```

**Expected behavior**: 
Charon should support cast

**Code snippet that is not correctly supported**:

```rust
fn test_float (x: f64) -> f64 {
    x + 2
}
```

**Current Charon output**:

```
error: Floats are not supported yet
   --> tests/ui/taint.rs:173:1
    |
    | / fn test_float (x: f64) -> f64 {
    | |     x + 2
    | | }
    | |_^

```

**Expected behavior**: 
Charon should support floats

Nadrieril:
This adds support for most MIR rvales in llbc. Fixes #332


charon/src/ast/expressions.rs
```diff
@@ -159,13 +171,15 @@ pub enum CastKind {
...
     Scalar(LiteralTy, LiteralTy),
+    RawPtr(Ty, Ty),
     FnPtr(Ty, Ty),
     /// [Unsize coercion](https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html). This is
     /// either `[T; N]` -> `[T]` or `T: Trait` -> `dyn Trait` coercions, behind a pointer
     /// (reference, `Box`, or other type that implements `CoerceUnsized`).
     ///
     /// The special case of `&[T; N]` -> `&[T]` coercion is caught by `UnOp::ArrayToSlice`.
     Unsize(Ty, Ty),
+    Transmute(Ty, Ty),
```
sonmarcho, line 182: Could we add a comment explaining this one? :)

charon/src/ast/expressions.rs
```diff
@@ -451,13 +465,20 @@ pub struct ConstantExpr {
     Debug, Clone, EnumToGetters, EnumAsGetters, EnumIsA, Serialize, Deserialize, Drive, DriveMut,
 )]
 pub enum Rvalue {
+    /// Returns the operand unchanged.
```
sonmarcho, line 468: I think the real distinction is rather than operands can be used in different situations (for instance as inputs to functions), so this simply lifts operands to rvalues. By contrast, a borrow can not be used directly as input to a function (we have to create the borrow, store it in a temporary value, then move it).

charon/tests/ui/rvalues.rs
```diff
@@ -0,0 +1,77 @@
...
+fn boxes() {
+    let _ = Box::new(42);
+    let _ = vec![0, 1];
+}
+
+// We force stealing to get the optimized_mir of the function. Optimizing the whole file would
+// remove most of our (indeed useless) assignments.
+static STEAL: [(); transmute([1, 0]) as usize] = [(); 1];
+const fn transmute(x: [u32; 2]) -> u64 {
+    // When optimized, this becomes a uilt-in cast. Otherwise this is just a call to `transmute`.
```
sonmarcho, line 57:
```suggestion
    // When optimized, this becomes a built-in cast. Otherwise this is just a call to `transmute`.
```

PR #344, opened 2024-09-03
Improve logging

Nadrieril:
This improves the logging setup to in particular include timing information for functions annotated with `#[tracing::instrument]`. To get these logs, call charon with `RUSTC_LOG=charon`.

### DIFF
--- a/charon/Cargo.toml
+++ b/charon/Cargo.toml
@@ -69,7 +69,8 @@ stacker = "0.1"
 take_mut = "0.2.2"
 toml = { version = "0.8", features = ["parse"] }
 tracing-subscriber = { version = "0.3", features = [ "env-filter", "std", "fmt" ] }
-tracing-tree = "0.3"
+tracing-tree = { git = "https://github.com/Nadrieril/tracing-tree", features = [ "time" ] } # Fork with improved formating and timing info.
+# tracing-tree = { path = "../../tracing-tree", features = [ "time" ] }
 tracing = { version = "0.1", features = [ "max_level_trace", "release_max_level_warn" ] }
 which = "6.0.1"
 
--- a/charon/src/logger.rs
+++ b/charon/src/logger.rs
@@ -1,7 +1,6 @@
 extern crate env_logger;
 
-/// Initialize the logger. We use a custom initialization to add some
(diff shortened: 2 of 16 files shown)


charon/Cargo.toml
```diff
@@ -69,7 +69,8 @@ stacker = "0.1"
 take_mut = "0.2.2"
 toml = { version = "0.8", features = ["parse"] }
 tracing-subscriber = { version = "0.3", features = [ "env-filter", "std", "fmt" ] }
-tracing-tree = "0.3"
+tracing-tree = { git = "https://github.com/Nadrieril/tracing-tree", features = [ "time" ] } # Fork with improved formating and timing info.
```
sonmarcho, line 72: What is your plan for this fork?
If we keep this change, on the longer term I think we should make the fork live in the AeneasVerif repo.
What do you think?

charon/src/logger.rs
```diff
@@ -11,130 +10,86 @@ pub fn initialize_logger() {
...
-            tracing_tree::HierarchicalLayer::new(2)
+            tracing_tree::HierarchicalLayer::new(1)
                 .with_ansi(true)
-                .with_indent_amount(1)
-                .with_indent_lines(true),
-        );
-    tracing::subscriber::set_global_default(subscriber).unwrap();
+                .with_indent_lines(true)
+                .with_bracketed_fields(true)
+                .with_timer(tracing_tree::time::Uptime::default()),
```
sonmarcho, line 22: Do we print the local time of the machine, or the time elapsed since the start of the execution?

Nadrieril: Each function call prints the time elapsed during that function call

charon/Cargo.toml
Nadrieril: The goal would be to upstream my changes to the main `tracing-tree` project. I'll experiment more in charon first though.

Importantly, note that this is a drop-in replacement for the upstream version. So we could revert to the upstream version and only use my fork locally when debugging, if using my fork like this is undesireable.

sonmarcho: Upstreaming the changes to the main `tracing-tree` is probably the best thing to do.

PR #346, opened 2024-09-04
Merge linear control-flow before reconstruction

Nadrieril:
This PR does several cfg-related things:
- Makes some ullbc terminators be statements instead; this reduces terminators to be the bare minimum needed for control flow: goto, switch, return and abort;
- Adds a pass that merges linear `goto` chains into a single block. This happens particularly in bodies with lots of function calls; this lightens the work for the control-flow reconstruction pass;
- ~~These two transformations made the llbc output slightly worse, so this PR also adds a pass that improves the cfg.~~

### DIFF
--- a/charon/Makefile
+++ b/charon/Makefile
@@ -3,7 +3,7 @@ all: build
 
 .PHONY: build
 build: format
-	cargo build --release
+	CARGO_PROFILE_RELEASE_DEBUG=true cargo build --release
 
 .PHONY: format
 format:
--- a/charon/src/transform/prettify_cfg.rs
+++ b/charon/src/transform/prettify_cfg.rs
@@ -1,0 +1,46 @@
+use crate::ids::Vector;
+use crate::llbc_ast::*;
+use crate::transform::TransformCtx;
+
+use super::ctx::LlbcPass;
+
+pub struct Transform;
+
+impl Transform {
+    fn update_statements(locals: &Vector<VarId, Var>, seq: &mut [Statement]) -> Vec<Statement> {
+        // Remove double aborts. This can happen when a function call is turned into an `Abort` by
+        // `inline_local_panic_functions`.
+        if let [Statement {
+            content: RawStatement::Abort(_),
+            ..
+        }, Statement {
+            content: second_abort @ RawStatement::Abort(_),
+            ..
+        }, ..] = seq
+        {
+            *second_abort = RawStatement::Nop;
+            return Vec::new();
+        }
+        if let [Statement {
+            content: RawStatement::Call(call),
+            ..
+        }, Statement {
+            content: second_abort @ RawStatement::Abort(_),
+            ..
+        }, ..] = seq
+            && locals[call.dest.var_id].ty.is_never()
+        {
+            *second_abort = RawStatement::Nop;
+            return Vec::new();
+        }
+
+        Vec::new()
+    }
+}
+
+impl LlbcPass for Transform {
+    fn transform_body(&self, _ctx: &mut TransformCtx<'_>, b: &mut ExprBody) {
+        b.body
+            .transform_sequences(&mut |seq| Transform::update_statements(&b.locals, seq))
+    }
+}
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.37"
+let supported_charon_version = "0.1.38"
--- a/charon-ml/src/PrintUllbcAst.ml
+++ b/charon-ml/src/PrintUllbcAst.ml
@@ -31,7 +31,9 @@ module Ast = struct
     | StorageDead var_id ->
         indent ^ "storage_dead " ^ var_id_to_string env var_id
     | Deinit p -> indent ^ "deinit " ^ place_to_string env p
-    | StAssert a -> assertion_to_string env indent a
+    | Assert a -> assertion_to_string env indent a
+    | Drop p -> indent ^ "drop " ^ place_to_string env p
+    | Call call -> call_to_string env indent call
 
   let switch_to_string (indent : string) (tgt : switch) : string =
     match tgt with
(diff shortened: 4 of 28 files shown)


charon/Makefile
```diff
@@ -3,7 +3,7 @@ all: build
 
 .PHONY: build
 build: format
-	cargo build --release
+	CARGO_PROFILE_RELEASE_DEBUG=true cargo build --release
```
sonmarcho, line 6: What does this do?

charon/src/ast/llbc_ast_utils.rs
```diff
@@ -38,29 +38,38 @@ pub fn combine_switch_targets_span(targets: &Switch) -> Span {
...
+            Switch::If(_, exp1, exp2) => Either::Left([&**exp1, &**exp2].into_iter()),
+            Switch::SwitchInt(_, _, targets, otherwise) => Either::Right(Either::Left(
+                targets.iter().map(|(_, tgt)| tgt).chain([&**otherwise]),
+            )),
+            Switch::Match(_, targets, otherwise) => Either::Right(Either::Right(
+                targets
+                    .iter()
+                    .map(|(_, tgt)| tgt)
+                    .chain(otherwise.as_deref()),
+            )),
```
sonmarcho, line 53: Why do you use `Either` in this code?

charon/src/transform/prettify_cfg.rs
```diff
@@ -0,0 +1,105 @@
...
+                            // anything.
+                            None
+                        }
+                    }
+                    None => {
+                        drop(nondivergents);
+                        // All the branches diverge (this is often the case in desugared loops). We
+                        // pick a branch that diverges closest (e.g. we prefer picking `break` over
+                        // `panic`). In case of a tie, we pick the longest branch.
+                        switch
```
sonmarcho, line 67: Does this case really improve the code?

charon/src/ast/llbc_ast_utils.rs
```diff
@@ -130,6 +139,49 @@ impl Statement {
...
+                .and_then(|st| st.diverges()),
+            RawStatement::Switch(switch) => {
+                let mut min_div = (4, 0, 0);
+                for st in switch.iter_targets() {
+                    match st.diverges() {
+                        None => return None,
+                        Some(div) => min_div = std::cmp::min(div, min_div),
+                    }
+                }
+                Some(min_div)
```
sonmarcho, line 179: I'm not sure that I agree with this case.
If we have:
```
if b {
  st
} else {
  switch {
     ...
  }
}
```
and all the branches in the `switch` break the control-flow, do we really want to move `st`?
(I don't understand when this case happens)

charon/Makefile
Nadrieril: That's a drive-by change that keeps more debuginfo in release mode, so that we get useful backtraces and flamegraphs. Otherwise a lot of info is lost and the flamegraphs are useless

charon/src/ast/llbc_ast_utils.rs
Nadrieril: Because each branch returns a different iterator type, so if I didn't use `Either` I'd get a type error. `Either` merges two possible iterator types into one

charon/src/ast/llbc_ast_utils.rs
```diff
@@ -130,6 +139,49 @@ impl Statement {
...
+            RawStatement::Switch(switch) => {
+                1 + switch.iter_targets().map(|st| st.len()).sum::<usize>()
+            }
+            RawStatement::Loop(st) => 1 + st.len(),
+            RawStatement::Nop => 0,
+            _ => 1,
+        }
+    }
+
+    /// Computes whether this statement diverges, and if yes returns a value indicating how far it
```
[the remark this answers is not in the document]
Nadrieril: "diverges" is the term that the rust typechecker uses though

charon/src/transform/prettify_cfg.rs
Nadrieril: It does because we often have `if foo { break 0 } else { statements...; continue 0 }`, where both branches jump but it would be more readable with the `continue` branch outside the `if`

charon/src/ast/llbc_ast_utils.rs
Nadrieril: Same as above, happens if all the branches end with `break` or `continue`, e.g. [here](https://github.com/AeneasVerif/charon/pull/346/files#diff-8473fe4a377959b679ef87dddfe430b803ee027fe69bdc77546f20fcf6de9fbdL302-L338)

Nadrieril:
There's too high a risk of breaking libcrux because eurydice relies on specific control-flow shapes to detect loops. I'll remove the last pass added in this PR and keep it for when libcrux is fixed.

charon/src/ast/llbc_ast_utils.rs
sonmarcho: Indeed, thanks!

charon/src/ast/llbc_ast_utils.rs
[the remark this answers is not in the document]
sonmarcho: Ok, let's keep it that way then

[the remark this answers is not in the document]
sonmarcho: But maybe precise that here by diverge we're using the rustc terminology and it means "jump"? (I'm not the only one who will bug on this comment, I think :) )

charon/src/transform/prettify_cfg.rs
sonmarcho: Indeed, it makes sense.

sonmarcho:
> There's too high a risk of breaking libcrux because eurydice relies on specific control-flow shapes to detect loops. I'll remove the last pass added in this PR and keep it for when libcrux is fixed.

I guess we should open an issue for this, then?
For the loops: you mean to generate `for` and `while` loops, right?

charon/src/ast/llbc_ast_utils.rs
sonmarcho: The link doesn't point to something valid it seems

Nadrieril:
> I guess we should open an issue for this, then?

I mean I've got the branch ready for whenever it's fixed

> For libcrux: you mean to generate `for` and `while` loops, right?

yes

charon/src/ast/llbc_ast_utils.rs
Nadrieril: Ah yeah, because I changed the code since I guess. We'll discuss again when I revive that change

sonmarcho:
> > I guess we should open an issue for this, then?
> 
> I mean, I've got the branch ready for whenever it's fixed, no need

As long as we don't forget it, I'm fine with that.

PR #348, opened 2024-09-05
Add support for float literals

R1kM:
Companion PR to https://github.com/hacspec/hax/pull/880/

Propagates support for float literals such as `1.52f32` or simply `2.0`


charon/src/ast/values.rs
```diff
@@ -81,3 +82,30 @@ pub enum ScalarValue {
...
+    Ord,
+    Drive,
+    DriveMut,
+)]
+pub enum FloatValue {
+    F16(u128),
+    F32(u128),
+    F64(u128),
+    F128(u128),
+}
```
sonmarcho, line 111: I think it's worth adding a comment explaining why we store the float values as `u128`

sonmarcho (approved):
This looks good to me, thanks!

charon/src/ast/values.rs
R1kM: As discussed in person, the comment (hidden behind the stack of Derives) is fine

charon/src/lib.rs
```diff
@@ -24,6 +24,7 @@
 // For when we use charon on itself :3
 #![register_tool(charon)]
 
+extern crate rustc_apfloat;
```
Nadrieril, line 27: The main charon library should use the strict minimum amount of rustc dependencies, in the end I'd like to have 0 rustc dependencies, so I'm not a fan of adding one.

charon/src/ast/values.rs
Nadrieril, line 111: Instead of an enum, could this be
```rust
struct FloatValue {
    bits: u128,
    ty: FloatTy,
}
```
That's more convenient to manipulate most of the time.

charon/src/ast/values_utils.rs
```diff
@@ -303,3 +303,68 @@ impl<'de> Deserialize<'de> for ScalarValue {
...
+                self,
+                mut map: A,
+            ) -> Result<Self::Value, A::Error> {
+                use serde::de::Error;
+                let (k, v): (String, String) = map.next_entry()?.expect("Malformed FloatValue");
+                Ok(match k.as_str() {
+                    "F16" => FloatValue::F16(v.parse().unwrap()),
+                    "F32" => FloatValue::F32(v.parse().unwrap()),
+                    "F64" => FloatValue::F64(v.parse().unwrap()),
+                    "F128" => FloatValue::F128(v.parse().unwrap()),
```
Nadrieril, line 359: This will try to parse a float value into a `u128`, it won't work

charon/src/bin/charon-driver/translate/translate_constants.rs
```diff
@@ -42,6 +42,16 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
                     }
                 };
                 Literal::Scalar(scalar)
+            },
+            hax::ConstantLiteral::Float(bits, float_type) => {
+                use hax::FloatTy;
+                let value = match float_type {
+                    FloatTy::F16 => FloatValue::F16(*bits as u128),
```
Nadrieril, line 49: why do you need the `as u128`? isn't `*bits` already a `u128`?

charon/src/ast/values.rs
Nadrieril: Do we even need the bit-value? We could store a string representation

charon-ml/src/GAstOfJson.ml
```diff
@@ -912,6 +912,9 @@ and literal_of_json (js : json) : (literal, string) result =
     | `Assoc [ ("Scalar", scalar) ] ->
         let* scalar = scalar_value_of_json scalar in
         Ok (VScalar scalar)
+    | `Assoc [ ("Float", float) ] ->
+        let* float = float_value_of_json float in
```
Nadrieril, line 916: This won't work: you serialized the float value to a string, this expects a json float

charon-ml/src/GAstOfJson.ml
```diff
@@ -912,6 +912,9 @@ and literal_of_json (js : json) : (literal, string) result =
     | `Assoc [ ("Scalar", scalar) ] ->
```
Nadrieril, line 912: As indicated at the top of this file, it should not be modified by hand. Edit `charon/src/bin/generate-ml/templates/GAstOfJson.ml` instead and call `make generate-ml` to propagate your changes to this file. In this particular case the definition is auto-generated so you shouldn't have anything to do.

Nadrieril:
Since you modified the json output, please increment the version in `Cargo.toml` and re-run `make test`

charon/src/lib.rs
Nadrieril: Even better: we could use `rustv_apfloat` from `crates.io` here, since we don't need to pass the resulting values to rustc

charon-ml/src/GAstOfJson.ml
```diff
@@ -912,6 +912,9 @@ and literal_of_json (js : json) : (literal, string) result =
...
         let* scalar = scalar_value_of_json scalar in
         Ok (VScalar scalar)
+    | `Assoc [ ("Float", float) ] ->
+        let* float = float_value_of_json float in
+        Ok (VFloat float)
```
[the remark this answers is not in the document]
Nadrieril: Also, is `float` a reserved keyword in ocaml? If so, could you add it to the list of reserved keywords in `generate-ml/main.rs` in `make_ocaml_ident`?

charon/src/ast/values.rs
R1kM: Done in https://github.com/AeneasVerif/charon/pull/348/commits/7e49300f449f6cc9bf27d6f09f756d07f5c0083b

charon/src/lib.rs
R1kM: Done in https://github.com/AeneasVerif/charon/pull/348/commits/1bf4d29b4d8183b33cf881ae192358ac0dc4e399

charon/src/bin/charon-driver/translate/translate_constants.rs
R1kM: Not applicable anymore as the representation changed

charon/src/ast/values_utils.rs
R1kM: We're now using automatically derived (de)serializers since we are not risking integer overflows anymore with the new float value representation

Nadrieril:
Could you add `FloatValue` to the lists of types in `generate-ml/main.rs`? You'll have to add it twice: once to generate the type definition, and once to generate the json translation. In both cases you can likely add it just below `"Literal"`. And don't forget to increment the version in `Cargo.toml`

Nadrieril (approved):
Looks good, could you please squash (e.g. `git reset --soft main; git commit`)?

PR #352, opened 2024-09-10
Simplify code generation and visitors

Nadrieril:
This PR greatly cleans up `charon-ml`:
- we now generate all the boilerplate visitors code;
- we no longer have to list which types to generate json code for (we still do for type declarations because of visitors);
- this removes many unneeded intermediate visitors;
- overall the template files are much more legible now.

There are visitors I would like to merge but cannot because of limitations in the `visitors` library. This is an issue for `dyn` support, as this requires `GenericParams` and `GenericArgs` to be in the same group of mutual recursion.

### DIFF
--- a/charon-ml/src/Expressions.ml
+++ b/charon-ml/src/Expressions.ml
@@ -72,24 +72,18 @@ and abort_kind =
        *)
 [@@deriving show, ord]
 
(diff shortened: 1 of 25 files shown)


sonmarcho:
It's indeed a lot better now!

PR #356, opened 2024-09-12
Distinguish statements and sequences of statements in rust llbc ast

Nadrieril:
This removes `RawStatement::Sequence` and instead correctly tracks where lists of statements are allowed, namely in loops, switch branches, and function bodies. This makes operating on statements much easier. This does not affect the caml side.


sonmarcho:
Actually I think it makes sense to do the same on the Caml side: I opened an issue

PR #385, opened 2024-09-30
Add an option to not merge the chains of gotos

sonmarcho:


Nadrieril:
What's the usecase for this feature?

Nadrieril:
What's the usecase for this feature?

sonmarcho:
Just noticed I answered this question on the Zulip but not on github.
This option deactivates the micro-pass which merges basic blocks in ULLBC when they form a "chain" of gotos, prior to control-flow reconstruction. I'm using it in the port of Rudra to use Charon: Rudra uses a taint analysis which loses precision if several function calls are grouped into a single basic block, as can happen in ULLBC if we merge the goto chains.

PR #389, opened 2024-10-01
Improve comments heuristic

### LINKED ISSUE #387: Comments trim all of leading whitespace
For a comment of the form:

```rust
// A
//   B
let x = ...
```

what I receive in the `comments_before` of the `let` node is `"A"` and `"B"`. I expected `"A"` and `"  B"` so as to preserve the intended alignment / indentation.

### LINKED ISSUE #388: Suboptimal placement of comment
Consider the following function (from mlkem):

```rust
pub(crate) fn barrett_reduce_element(value: FieldElement) -> FieldElement {
    // hax_debug_assert!(
    //     i32::from(value) > -BARRETT_R && i32::from(value) < BARRETT_R,
    //     "value is {value}"
    // );

    let t = (i32::from(value) * BARRETT_MULTIPLIER) + (BARRETT_R >> 1);
    ...
```

If I dump what I receive from Charon, the comment is attached to the intermediary (unnamed) value that MIR introduces for `i32::from(value)` (which internally is just `value`), or more specifically:

```
// pseudo-MIR
fn barrett_...(...) {
... lots of variable declarations
i32 t; // this is the user-provided name
... lots of variable declarations
i32 tmp; // internal MIR name
... lots more variable declarations
<<comment attached to the assignment below>>
tmp := value;
```

I would expect the comment to be attached either to
i) the declaration `i32 t`; or ii) the assignment into `t` that comes further down (not modeled in the snippet above).

Right now, Eurydice tracks the location of the comment all the way down to C, re-inlines the temporary computations, and produces something unsightly:

```c
  int32_t
  t =
    (int32_t)/*
    hax_debug_assert!(
    i32::from(value) > -BARRETT_R && i32::from(value) < BARRETT_R,
    "value is {value}"
    );
    */
      value
```

notice how the comment is attached to the inner `value`, per the original LLBC received from Charon. Thank you!

Nadrieril:
Fixes #388.

### DIFF
--- a/charon/Cargo.lock
+++ b/charon/Cargo.lock
@@ -166,6 +166,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "by_address"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "64fa3c856b712db6612c019f14756e64e4bcea13337a6b33b696333a9eaa2d06"
+
+[[package]]
 name = "cc"
 version = "1.0.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -183,6 +189,7 @@ version = "0.1.44"
 dependencies = [
  "anyhow",
  "assert_cmd",
+ "by_address",
  "clap",
  "colored",
  "convert_case 0.6.0",
--- a/charon/Cargo.toml
+++ b/charon/Cargo.toml
@@ -77,6 +77,7 @@ which = "6.0.1"
 
 hax-frontend-exporter = { git = "https://github.com/hacspec/hax", branch = "main", optional = true }
 macros = { path = "./macros" }
+by_address = "1.2.1"
 
 [features]
 default = ["rustc"]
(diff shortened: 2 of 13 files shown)


msprotz:
FYI, just gave this a quick try, and I don't seem to be receiving the comments at all anymore (or at least, the one I noticed had a problem)

Nadrieril:
Yeah I marked it as draft because it doesn't work for complex cases such as destructuring or `assert!` that you test I Eurydice

sonmarcho:
This seems to be rotting. @Nadrieril do you have any plan for this PR?

Nadrieril:
I was hoping to get back to it when higher-priority items are done, but tbf I don't know how to go about this. The few heuristics I tried don't work and the last idea I had required an important change to the visitors library we're using (so we can hold on to borrows while visiting)

msprotz:
Is no one else in the Rust ecosystem doing this? (Retaining inline comments.)

Should this be a discussion with upstream Rustc to add more support in the compiler itself?

Nadrieril:
I don't know of anyone else doing that no

Nadrieril:
This has been in draft for 2 years, I'll come back to it if we find comments to be our priority, but it isn't at the moment.

PR #392, opened 2024-10-01
Add the source files in the translated crate

sonmarcho:
Having the source code in the serialized files is extremely convenient.
Also, I checked the size of the serialized files: the increase in size of the JSON files is generally negligible (a few percent).

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
+++ b/charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
@@ -361,7 +361,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
     }
 
     /// Return a map from source file name to file content.
-    fn read_source_files(&mut self) -> HashMap<FileName, String> {
+    fn read_source_files(&mut self) -> HashMap<FileName, Arc<String>> {
         // Retrieve the source map
         let source_map = self.tcx.sess.source_map();
 
@@ -377,7 +377,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
                 let filename = self.translate_filename(&filename);
 
                 // Retrieve the content
-                file.src.as_ref().map(|src| (filename, src.deref().clone()))
+                file.src.as_ref().map(|src| (filename, src.clone()))
             })
             .collect()
     }
@@ -477,11 +477,16 @@ pub fn translate<'tcx, 'ctx>(options: &CliOpts, tcx: TyCtxt<'tcx>) -> TransformC
 
     // Read the source files
     let source_files = ctx.read_source_files();
+    use std::ops::Deref;
     ctx.translated.file_id_to_content = ctx
         .translated
         .id_to_file
         .iter_indexed()
-        .filter_map(|(id, filename)| (source_files.get(filename).map(|file| (id, file.clone()))))
+        .filter_map(|(id, filename)| {
+            source_files
+                .get(filename)
+                .map(|file| (id, file.deref().clone()))
+        })
         .collect();
 
     // Return the context, dropping the hax state and rustc `tcx`.
--- a/charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
+++ b/charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
@@ -359,28 +359,27 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
         }
         Ok(self.translated.get_item(id).unwrap())
     }
-}
 
-/// Read and return the content of a source file.
-fn read_file_content(filename: &FileName) -> Option<Vec<String>> {
-    use std::fs::File;
-    use std::io::prelude::*;
-    use std::io::BufReader;
-    use FileName::*;
-    match filename {
-        Local(path) => {
-            let file = File::open(path).ok()?;
-            BufReader::new(file)
-                .lines()
-                .into_iter()
-                .collect::<Result<Vec<String>, _>>()
-                .ok()
-        }
-        Virtual(_) => {
-            // TODO: we should be able to retrieve the file content here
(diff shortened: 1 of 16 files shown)


charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
```diff
@@ -360,6 +361,29 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+    use std::io::BufReader;
+    use FileName::*;
+    match filename {
+        Local(path) => {
+            let file = File::open(path).ok()?;
+            BufReader::new(file)
+                .lines()
+                .into_iter()
+                .collect::<Result<Vec<String>, _>>()
+                .ok()
```
Nadrieril, line 377: This may not be exactly the source that was read because of races on the filesystem or similar issues. I recommend we retrieve the source from rustc directly using [`SourceMap::get_source_file`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/source_map/struct.SourceMap.html#method.get_source_file) and `SourceFile.src` instead

sonmarcho: Ok thanks for the remark

Nadrieril:
That's a good idea, it will also help get for getting rid of the rustc dependency for displaying errors (maybe that's already what you had in mind)

charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
sonmarcho: Actually it's a good way of retrieving the source for the virtual files, I think

charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
```diff
@@ -358,6 +359,28 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+
+    /// Return a map from source file name to file content.
+    fn read_source_files(&mut self) -> HashMap<FileName, String> {
+        // Retrieve the source map
+        let source_map = self.tcx.sess.source_map();
+
+        // Read all the files
+        use std::ops::Deref;
+        source_map
+            .files()
```
Nadrieril, line 371: Will that include every file of every dependency? Can you check if the impact on json size is still small? Otherwise we should only add to the map whenever we register a new file id

sonmarcho: Yes, it probably loads all the files. But if you look carefully, you'll notice that when I compute the map from file id to file content, I only lookup the ids which we register in the spans, meaning that only the relevant files are written to the JSON.
Also, I thought of loading the file contents whenever registering a new file id (it seemed the natural way of doing) but actually it is cumbersome because we have access to *hax* information, not rustc information, meaning we actually don't know how to retrieve the file content. As a consequence, it proved more natural to load all the files, and compute an intermediate mapping from (LLBC) file name to file content.

Nadrieril: Then maybe we should let hax do this for us, or at least we could keep the `Arc<String>` instead of cloning everything

Nadrieril: Either way please add a comment explaining this

sonmarcho: Yes, makes sense: I kept the `Arc<String>`

charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
```diff
@@ -477,11 +477,16 @@ pub fn translate<'tcx, 'ctx>(options: &CliOpts, tcx: TyCtxt<'tcx>) -> TransformC
...
+    use std::ops::Deref;
     ctx.translated.file_id_to_content = ctx
         .translated
         .id_to_file
         .iter_indexed()
-        .filter_map(|(id, filename)| (source_files.get(filename).map(|file| (id, file.clone()))))
+        .filter_map(|(id, filename)| {
+            source_files
+                .get(filename)
+                .map(|file| (id, file.deref().clone()))
```
Nadrieril, line 488: Why is the `deref` needed here?

charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
```diff
@@ -358,6 +359,28 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
...
+    /// Return a map from source file name to file content.
+    fn read_source_files(&mut self) -> HashMap<FileName, Arc<String>> {
+        // Retrieve the source map
+        let source_map = self.tcx.sess.source_map();
+
+        // Read all the files
+        use std::ops::Deref;
+        source_map
+            .files()
+            .deref()
```
Nadrieril, line 372: Why is `deref` needed here?

charon/src/bin/charon-driver/translate/translate_crate_to_ullbc.rs
```diff
@@ -358,6 +359,28 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx, 'ctx> {
         }
         Ok(self.translated.get_item(id).unwrap())
     }
+
+    /// Return a map from source file name to file content.
+    fn read_source_files(&mut self) -> HashMap<FileName, Arc<String>> {
+        // Retrieve the source map
+        let source_map = self.tcx.sess.source_map();
+
+        // Read all the files
```
Nadrieril, line 368: Could you mention that this includes all the files of all the dependent crates too and hence we should be careful?

sonmarcho: Yes, good idea

PR #425, opened 2024-10-18
Test charon on popular crates

### LINKED ISSUE #426: test charon on the most popular crates.io crates

Nadrieril:
This adds a disabled-by-default test that downloads the 50 most downloaded crates from crates.io and runs Charon on them. Today 29 of the top 50 pass without crashing (with `--errors-as-warnings`).

Fixes #426.

### DIFF
--- a/charon/Cargo.lock
+++ b/charon/Cargo.lock
@@ -3,6 +3,27 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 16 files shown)


Nadrieril:
This is not run on CI because it takes a while, but we could run it weekly if we wanted

Issue #145, opened 2024-04-23
Tracking issue for passing the rustc test suite


Nadrieril:
In https://github.com/AeneasVerif/charon/pull/136/ we added a script that runs charon on the rustc test suite and displays the results. As we'd like to support a large portion of the rust language, we will track in this issue our results on these tests.

sonmarcho:
I just noticed we almost have 7.2k success now, to be compared with the 4.2k success from when we started tracking this issue :)

Nadrieril:
Big change this week, in https://github.com/AeneasVerif/charon/pull/536 I was finally able to distinguish rustc errors that indicate incorrect code from charon panics, thanks to [catch_fatal_errors](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_driver/fn.catch_fatal_errors.html).

sonmarcho:
What is "success-when-failure-expected"?

Nadrieril:
It's a case where charon successfully compiled despite the fact that the test was supposed to be a failing test. This is most likely because we don't parse most of the test options in files.

Nadrieril:
I was emitting error codes incorrectly, with https://github.com/AeneasVerif/charon/pull/555 we should now get sensible results.

sonmarcho:
This is neat :)

Nadrieril:
I've sorted the unexpected failures due to hax, they're all covered by:
- https://github.com/AeneasVerif/charon/issues/546;
- Support for promoted constants (#543);
- https://github.com/cryspen/hax/issues/1311;
- A bunch of unhandled cases in the const conversion code.

In other words it's all about constants except for https://github.com/AeneasVerif/charon/issues/546.

The panics are all because of the same assertion in `expand_associated_types`, fix incoming.

Nadrieril:
No panics left!

Nadrieril:
Triggering a manual run after https://github.com/AeneasVerif/charon/pull/669

Nadrieril:
(toolchain got bumped)

Nadrieril:
(updated the rustc toolchain this week, looks like most of the diff is due to that)

PR #430, opened 2024-10-18
Set `--errors-as-warnings` by default

### LINKED ISSUE #416: make errors-as-warnings the default

Nadrieril:
This way charon passes on a lot more crates right away. We now have `--error-on-warnings` as a new flag that restores the old behavior.

This should drastically change the behavior on the rustc test suite (#145).

Fixes #416.


Nadrieril:
cc @EschericHya @sonmarcho @msprotz this changes the default behavior of charon to _not_ fail when there are errors.

PR #431, opened 2024-10-18
Remove im dependency

zhassan-aws:
Remove the `im` crate dependency and replace the collections used with ones from the standard library.


Nadrieril (approved):
Looks good, I'll just wait Friday to merge because we're in the midst of preparing Charon for a paper submission

PR #440, opened 2024-10-28
fix(hax api): follow up PR for hacspec/hax#1057

W95Psp:


charon/src/bin/charon-driver/translate/translate_constants.rs
```diff
@@ -145,9 +145,28 @@ impl<'tcx, 'ctx, 'ctx1> BodyTransCtx<'tcx, 'ctx, 'ctx1> {
...
+                mutability: true, ..
+            } => {
+                error_or_panic!(
+                    self,
+                    span,
+                    &format!("Unsupported constant: `ConstantExprKind::RawBorrow {{mutability: true, ..}}`",)
+                )
+            }
+            ConstantExprKind::RawBorrow {
+                mutability: false,
```
Nadrieril, line 165: Should be `true` :D

W95Psp: oops, I pushed an embarrassing commit fixing this embarrassing mistake, thanks

PR #456, opened 2024-11-06
Add a micro-pass to fix borrow-checking issues introduced by `index_to_function_calls`

sonmarcho:
This fixes https://github.com/AeneasVerif/aeneas/issues/295.
The issue is that if you write `x[0] += 1`, the micro-pass `index_to_function_calls` transforms the code to (note that the problem remains if we introduce `src` before `dest`):
```rust
dest = &mut x[0];
src = & x[0]; // Invalidates dest
*dest = copy (*src) + 1; // Can't dereference dest!
```

This PR adds a micro-pass which introduces an intermediate assignment to prevent this issue.
We first do:
```rust
tmp = copy x[0] + 1;
x[0] = move tmp;
```

Which leads to:
```rust
// RHS:
src = & x[0];
tmp = copy (*src) + 1;
// LHS:
dest = &mut x[0];
*dest = move tmp;
```

### DIFF
--- a/charon/src/ast/expressions_utils.rs
+++ b/charon/src/ast/expressions_utils.rs
@@ -1,5 +1,6 @@
 //! This file groups everything which is linked to implementations about [crate::expressions]
 use crate::ast::*;
+use crate::ids::Vector;
 
 impl Place {
     pub fn new(var_id: VarId) -> Place {
@@ -17,5 +18,79 @@ impl BorrowKind {
         } else {
             Self::Shared
         }
+    }
+}
+
+impl Place {
+    /// Compute the type of a place.
+    pub fn compute_projected_type(
+        &self,
+        type_decls: &Vector<TypeDeclId, TypeDecl>,
+        locals: &Vector<VarId, Var>,
+    ) -> Result<Ty, ()> {
+        // Lookup the local
+        let mut cur_ty = locals.get(self.var_id).ok_or(())?.ty.clone();
+
+        // Apply the projection
+        use ProjectionElem::*;
+        for pe in &self.projection {
+            cur_ty = match pe {
+                Deref => {
+                    use TyKind::*;
+                    match cur_ty.kind() {
+                        Ref(_, ty, _) | RawPtr(ty, _) => ty.clone(),
+                        Adt(..) | TypeVar(_) | Literal(_) | Never | TraitType(..) | DynTrait(_)
+                        | Arrow(..) => {
+                            // Unreachable
+                            return Err(());
+                        }
+                    }
+                }
+                Field(pkind, field_id) => {
+                    // Lookup the type decl
+                    use FieldProjKind::*;
+                    match pkind {
+                        Adt(type_decl_id, variant_id) => {
+                            let type_decl = type_decls.get(*type_decl_id).ok_or(())?;
+                            use TypeDeclKind::*;
+                            match &type_decl.kind {
+                                Struct(fields) => {
+                                    if variant_id.is_some() {
+                                        return Err(());
+                                    };
+                                    fields.get(*field_id).ok_or(())?.ty.clone()
+                                }
+                                Enum(variants) => {
+                                    let variant_id = variant_id.ok_or(())?;
+                                    let variant = variants.get(variant_id).ok_or(())?;
+                                    variant.fields.get(*field_id).ok_or(())?.ty.clone()
+                                }
+                                Union(_) | Opaque | Alias(_) | Error(_) => return Err(()),
+                            }
+                        }
+                        Tuple(_) => cur_ty
+                            .as_tuple()
+                            .ok_or(())?
+                            .get(TypeVarId::from(usize::from(*field_id)))
+                            .ok_or(())?
+                            .clone(),
+                        ClosureState => return Err(()),
+                    }
+                }
+                Index { ty, .. } | Subslice { ty, .. } => {
+                    // We could check that that the current type is compatible with
+                    // the type of the
+                    let cur_ty = cur_ty.as_array_or_slice().ok_or(())?;
+                    let ty = ty.as_array_or_slice().ok_or(())?;
+                    if cur_ty == ty {
+                        ty.clone()
+                    } else {
+                        return Err(());
+                    }
+                }
+            };
+        }
+
+        Ok(cur_ty)
     }
 }
--- a/charon/src/ast/expressions_utils.rs
+++ b/charon/src/ast/expressions_utils.rs
@@ -52,18 +52,34 @@ impl Place {
                     match pkind {
                         Adt(type_decl_id, variant_id) => {
                             let type_decl = type_decls.get(*type_decl_id).ok_or(())?;
+                            let (type_id, generics) = cur_ty.as_adt().ok_or(())?;
+                            assert!(TypeId::Adt(*type_decl_id) == type_id);
+                            assert!(generics.regions.len() == type_decl.generics.regions.len());
+                            assert!(generics.types.len() == type_decl.generics.types.len());
+                            assert!(
+                                generics.const_generics.len()
+                                    == type_decl.generics.const_generics.len()
+                            );
+                            assert!(
+                                generics.trait_refs.len() == type_decl.generics.trait_clauses.len()
+                            );
                             use TypeDeclKind::*;
                             match &type_decl.kind {
                                 Struct(fields) => {
                                     if variant_id.is_some() {
                                         return Err(());
                                     };
-                                    fields.get(*field_id).ok_or(())?.ty.clone()
+                                    let mut ty = fields.get(*field_id).ok_or(())?.ty.clone();
+                                    ty.substitute(generics);
+                                    ty
                                 }
                                 Enum(variants) => {
                                     let variant_id = variant_id.ok_or(())?;
                                     let variant = variants.get(variant_id).ok_or(())?;
-                                    variant.fields.get(*field_id).ok_or(())?.ty.clone()
+                                    let mut ty =
+                                        variant.fields.get(*field_id).ok_or(())?.ty.clone();
+                                    ty.substitute(generics);
+                                    ty
                                 }
                                 Union(_) | Opaque | Alias(_) | Error(_) => return Err(()),
                             }
--- a/charon/src/ast/expressions_utils.rs
+++ b/charon/src/ast/expressions_utils.rs
@@ -39,6 +39,9 @@ impl Place {
                     use TyKind::*;
                     match cur_ty.kind() {
                         Ref(_, ty, _) | RawPtr(ty, _) => ty.clone(),
+                        Adt(TypeId::Builtin(BuiltinTy::Box), args) => {
+                            args.types.get(TypeVarId::new(0)).unwrap().clone()
+                        }
                         Adt(..) | TypeVar(_) | Literal(_) | Never | TraitType(..) | DynTrait(_)
                         | Arrow(..) => {
                             // Unreachable
@@ -65,7 +68,7 @@ impl Place {
                             );
                             use TypeDeclKind::*;
                             match &type_decl.kind {
-                                Struct(fields) => {
+                                Struct(fields) | Union(fields) => {
                                     if variant_id.is_some() {
                                         return Err(());
                                     };
@@ -81,7 +84,7 @@ impl Place {
                                     ty.substitute(generics);
                                     ty
                                 }
-                                Union(_) | Opaque | Alias(_) | Error(_) => return Err(()),
+                                Opaque | Alias(_) | Error(_) => return Err(()),
                             }
                         }
                         Tuple(_) => cur_ty
--- a/charon/src/ast/expressions_utils.rs
+++ b/charon/src/ast/expressions_utils.rs
@@ -94,10 +94,9 @@ impl Place {
                     }
                 }
                 Index { ty, .. } | Subslice { ty, .. } => {
-                    // We could check that that the current type is compatible with
-                    // the type of the
                     let cur_ty = cur_ty.as_array_or_slice().ok_or(())?;
                     let ty = ty.as_array_or_slice().ok_or(())?;
+                    // Sanity check: ensure we're using the same types.
                     if cur_ty == ty {
                         ty.clone()
                     } else {
--- a/charon/src/ast/types_utils.rs
+++ b/charon/src/ast/types_utils.rs
@@ -541,7 +541,6 @@ struct Subst<'a> {
 }
 
 impl<'a> Subst<'a> {
-    /// Returns [true] if the item is a binder (we use this to skip some checks)
     fn enter_poly_trait_decl_ref(&mut self, _: &mut PolyTraitDeclRef) {
         self.current_level.index += 1;
     }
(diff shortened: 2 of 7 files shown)


charon/src/ast/expressions_utils.rs
```diff
@@ -19,3 +20,77 @@ impl BorrowKind {
...
     }
 }
+
+impl Place {
+    /// Compute the type of a place.
+    pub fn compute_projected_type(
+        &self,
+        type_decls: &Vector<TypeDeclId, TypeDecl>,
+        locals: &Vector<VarId, Var>,
+    ) -> Result<Ty, ()> {
```
Nadrieril, line 30: That makes me think we should keep the type of places in translation

Nadrieril:
Please add a test for the case you mention.

charon/src/transform/index_intermediate_assigns.rs
```diff
@@ -0,0 +1,149 @@
...
+        index,
+        name: None,
+        ty: lhs_ty,
+    });
+
+    // Update the rhs
+    let tmp_rhs = std::mem::replace(rhs, Rvalue::Use(Operand::Move(Place::new(tmp_var))));
+
+    // Retrieve the "comments before" to move them to the new intermediate statement
+    let comments_before = std::mem::take(comments_before);
```
Nadrieril, line 108: This is not useful, the `comments_before` stays empty until the last pass of them all which adds them back.

sonmarcho: Ok, that's good to know

charon/src/ast/expressions_utils.rs
```diff
@@ -19,3 +20,77 @@ impl BorrowKind {
...
+                                        return Err(());
+                                    };
+                                    fields.get(*field_id).ok_or(())?.ty.clone()
+                                }
+                                Enum(variants) => {
+                                    let variant_id = variant_id.ok_or(())?;
+                                    let variant = variants.get(variant_id).ok_or(())?;
+                                    variant.fields.get(*field_id).ok_or(())?.ty.clone()
+                                }
+                                Union(_) | Opaque | Alias(_) | Error(_) => return Err(()),
```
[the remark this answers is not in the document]
sonmarcho: Yes indeed, I went too fast

sonmarcho:
> Please add a test for the case you mention.

This micro-pass is unfortunately not easily testable on the Charon side as its purpose is to make Aeneas' symbolic execution succeed. I prefer adding a test on Aeneas' side (I'm working on a PR to improve the handling of arrays/slices).

sonmarcho:
> Could we use a two-phase borrow for this? That's exactly what they're for

Following an offline conversation: two-phase borrows don't work here. Two-phase borrows work if we have two statements, and the reserved borrow gets activated between the first statement (which requires a shared borrow) and the second statement (which requires a mutable borrow). Here we have a single statement that we have to split in two, and we choose when we introduce the mutable borrow and the shared borrow (we would need two phase borrows if we had the bad idea of introducing the mutable borrow *before* the shared borrow)..

charon/src/ast/expressions_utils.rs
sonmarcho: Yes, I hesitated to do that a while ago. But note that it is not a bad idea to preserve this function as a helper, or rather an equivalent function which operates on a single projection element (as the function operating on places would be useless once we add the type information).

charon/src/ast/expressions_utils.rs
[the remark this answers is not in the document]
sonmarcho: I pushed something. I'm not very familiar with `DriveMut` and `Visitor` but I think my substitution function works.

Nadrieril:
> This micro-pass is unfortunately not easily testable on the Charon side

I would like at least one case that is changed by this PR, so we don't accidentally regress it.

charon/src/ast/expressions_utils.rs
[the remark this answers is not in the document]
Nadrieril: I have a branch that adds the types to places, it just needs to be propagated to aeneas. I'd rather use that, as there's no risk of getting it wrong. That said having substitution functions is useful, I'd be happy to merge something like that. I'll review on tuesday.

charon/src/ast/types_utils.rs
```diff
@@ -518,3 +525,118 @@ impl<V> std::ops::DerefMut for VisitInsideTy<V> {
...
+impl Ty {
+    pub fn substitute(&mut self, generics: &GenericArgs) {
+        let mut subst = Subst {
+            current_level: 0,
+            region_subst: Vector::from_iter(generics.regions.iter().cloned()),
+            ty_subst: Vector::from_iter(generics.types.iter().cloned()),
+            cg_subst: Vector::from_iter(generics.const_generics.iter().cloned()),
+            trait_subst: Vector::from_iter(generics.trait_refs.iter().cloned()),
+        };
+        self.drive_mut(&mut subst);
```
Nadrieril, line 640: This will not recurse deeply into types, you need to wrap `subst` in `Ty::visit_inside`

charon/src/ast/types_utils.rs
```diff
@@ -518,3 +525,118 @@ impl<V> std::ops::DerefMut for VisitInsideTy<V> {
...
+        // Other cases: we only do something upon entering
+        else if enters {
+            if self.try_subst_in_region(item) {
+            } else if self.try_subst_in_ty(item) {
+            } else if self.try_subst_in_const_generic(item) {
+            } else {
+                let _ = self.try_subst_in_trait_ref(item);
+            }
+        }
+    }
```
Nadrieril, line 628: This won't recurse deeply into `Ty` because the `DriveMut` impl of `Ty` does nothing by default. That's a good thing, otherwise a visitor like this one would not be possible. To make it work you need to call `ty.drive_inner_mut()` whenever you want to recurse. But then you don't need to downcast by hand, use `#[derive(VisitorMut)]` instead, you can see examples in many of the micro-passes.

To avoid double-substitutions, simply substitute on the `Exit` event instead of the `Enter` one.

charon/src/ast/types_utils.rs
```diff
@@ -518,3 +525,118 @@ impl<V> std::ops::DerefMut for VisitInsideTy<V> {
...
+}
+
+impl Ty {
+    pub fn substitute(&mut self, generics: &GenericArgs) {
+        let mut subst = Subst {
+            current_level: 0,
+            region_subst: Vector::from_iter(generics.regions.iter().cloned()),
+            ty_subst: Vector::from_iter(generics.types.iter().cloned()),
+            cg_subst: Vector::from_iter(generics.const_generics.iter().cloned()),
+            trait_subst: Vector::from_iter(generics.trait_refs.iter().cloned()),
```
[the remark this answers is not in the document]
sonmarcho: Oh indeed, in my mind the `GenericArgs` used something other than `Vector<Id, ...>`

charon/src/ast/expressions_utils.rs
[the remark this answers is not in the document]
sonmarcho: Yes, I definitely want to keep the substitution function. Let me insist a bit on keeping the projection function as well. :)

charon/src/ast/types_utils.rs
sonmarcho: Ah good `#[derive(VisitorMut)]` looks like what I'm looking for.
So what's the mental model for `Enter` and `Exit`? Does the visitor do: "call the function on enter", "drive inside the type", "call the function on exit"?

Nadrieril: yep, exactly. also instead of calling `drive_inner_mut`, you can wrap your visitor in `Ty::visit_inside_ty` which will do it for you.

Nadrieril:
> (if we regress on this, Aeneas breaks).

I insist, please add a test case. The correctness of charon should be tested in its own test suite. The fact that Aeneas relies on it is not part of the process of ensuring the correctness of charon, it's only a nice extra check. For example if one day Aeneas stops relying on this, and we haven't removed this micro-pass, I would not have any tests left for it.

charon/src/ast/expressions_utils.rs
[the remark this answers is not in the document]
Nadrieril: Now that you've written it we may as well keep it

charon/src/ast/types_utils.rs
sonmarcho: (I updated the code)

sonmarcho:
> > (if we regress on this, Aeneas breaks).
> 
> I insist, please add a test case. The correctness of charon should be tested in its own test suite. The fact that Aeneas relies on it is not part of the process of ensuring the correctness of charon, it's only a nice extra check. For example if one day Aeneas stops relying on this, and we haven't removed this micro-pass, I would not have any tests left for it.

Nevermind,  I had forgotten we actually check the output LLBC (I think you got my point if that hadn't been the case).

charon/src/ast/expressions_utils.rs
```diff
@@ -19,3 +20,93 @@ impl BorrowKind {
...
+            cur_ty = match pe {
+                Deref => {
+                    use TyKind::*;
+                    match cur_ty.kind() {
+                        Ref(_, ty, _) | RawPtr(ty, _) => ty.clone(),
+                        Adt(..) | TypeVar(_) | Literal(_) | Never | TraitType(..) | DynTrait(_)
+                        | Arrow(..) => {
+                            // Unreachable
+                            return Err(());
+                        }
```
Nadrieril, line 46: To be fully correct we should handle boxes here too

charon/src/ast/expressions_utils.rs
```diff
@@ -19,3 +20,93 @@ impl BorrowKind {
...
+                                }
+                                Enum(variants) => {
+                                    let variant_id = variant_id.ok_or(())?;
+                                    let variant = variants.get(variant_id).ok_or(())?;
+                                    let mut ty =
+                                        variant.fields.get(*field_id).ok_or(())?.ty.clone();
+                                    ty.substitute(generics);
+                                    ty
+                                }
+                                Union(_) | Opaque | Alias(_) | Error(_) => return Err(()),
```
Nadrieril, line 84: We should handle unions too

charon/src/ast/expressions_utils.rs
```diff
@@ -19,3 +20,93 @@ impl BorrowKind {
...
+                            return Err(());
+                        }
+                    }
+                }
+                Field(pkind, field_id) => {
+                    // Lookup the type decl
+                    use FieldProjKind::*;
+                    match pkind {
+                        Adt(type_decl_id, variant_id) => {
+                            let type_decl = type_decls.get(*type_decl_id).ok_or(())?;
```
Nadrieril, line 54: Hm, this case is annoying: the type decl may be missing because it was marked `--exclude`, which means this function can fail for reasons other than us messing up. It's fine for the current usage anyhow

charon/src/ast/expressions_utils.rs
```diff
@@ -19,3 +20,93 @@ impl BorrowKind {
...
+                            .clone(),
+                        ClosureState => return Err(()),
+                    }
+                }
+                Index { ty, .. } | Subslice { ty, .. } => {
+                    // We could check that that the current type is compatible with
+                    // the type of the
+                    let cur_ty = cur_ty.as_array_or_slice().ok_or(())?;
+                    let ty = ty.as_array_or_slice().ok_or(())?;
+                    if cur_ty == ty {
```
Nadrieril, line 101:
```suggestion
                    let cur_ty = cur_ty.as_array_or_slice().ok_or(())?;
                    let ty = ty.as_array_or_slice().ok_or(())?;
                    // Sanity check: ensure we're using the same types.
                    if cur_ty == ty {
```

charon/src/ast/types_utils.rs
```diff
@@ -496,3 +525,71 @@ impl<V> std::ops::DerefMut for VisitInsideTy<V> {
...
+    TraitRef(exit)
+)]
+struct Subst<'a> {
+    // Tracks the current de Bruijn level
+    current_level: DeBruijnId,
+    generics: &'a GenericArgs,
+}
+
+impl<'a> Subst<'a> {
+    /// Returns [true] if the item is a binder (we use this to skip some checks)
```
Nadrieril, line 544: This comment is no longer relevant

charon/src/ast/types_utils.rs
```diff
@@ -496,3 +525,71 @@ impl<V> std::ops::DerefMut for VisitInsideTy<V> {
...
+#[derive(VisitorMut)]
+#[visitor(
+    PolyTraitDeclRef(enter, exit),
+    Region(exit),
+    Ty(exit),
+    ConstGeneric(exit),
+    TraitRef(exit)
+)]
+struct Subst<'a> {
+    // Tracks the current de Bruijn level
```
Nadrieril, line 538: Could you add a note about why we consider only `PolyTraitDeclRef`? I think you had one in a previous commit.

charon/src/ast/types_utils.rs
```diff
@@ -496,3 +525,71 @@ impl<V> std::ops::DerefMut for VisitInsideTy<V> {
...
+            Region::BVar(db, id) => {
+                if *db == self.current_level {
+                    *r = self.generics.regions.get(*id).unwrap().clone()
+                }
+            }
+            _ => (),
+        }
+    }
+
+    /// Returns [true] if the item is a binder (we use this to skip some checks)
```
Nadrieril, line 564:
```suggestion
```

charon/src/transform/index_intermediate_assigns.rs
```diff
@@ -0,0 +1,138 @@
...
+        self.contains_index = self.contains_index || p.contains_index();
+    }
+}
+
+impl Rvalue {
+    fn contains_index(&self) -> bool {
+        let mut v = RvalueVisitor {
+            contains_index: false,
+        };
+        self.drive(&mut v);
```
Nadrieril, line 72: if you wish, you can use `self.drive(&mut visitor_enter_fn(|p: &Place| contains_index |= p.contains_index()))` to avoid the need for a separate visitor struct

charon/src/ast/expressions_utils.rs
Nadrieril: Let's just ignore that for now

Nadrieril:
I'm ok with the general shape of this. I have a bunch of nitpicks around comments and one important comment.

charon/src/ast/expressions_utils.rs
sonmarcho: Yes, that's the reason why I put an `ok_or` rather than an `unwrap`

charon/src/ast/expressions_utils.rs
sonmarcho: I updated the code

charon/src/ast/expressions_utils.rs
sonmarcho: I updated the code

charon/src/ast/types_utils.rs
sonmarcho: I added a comment. PolyTraitDeclRef is the only occurrence of RegionBinder that we can find in a binder so far.
I don't know how (is it possible?) to handle all occurrences of `RegionBinder`- this can be annoying for future updates of the code.

Nadrieril: There isn't really a way, that's one big drawback of the `dyn` visitor approach

Nadrieril: That's not the only place the `dyn Visitor` thing has had limitations, we might add more normal visitors back if that becomes too much of a problem.

PR #458, opened 2024-11-10
Update LlbcAstUtils.chain_statements

sonmarcho:
This is a very minor cleanup.


sonmarcho:
I don't understand the CI failing on Kyber: I think the failure doesn't come from changes on our side.

PR #459, opened 2024-11-11
Switch from `derivative` to `derive-where`

qinheping:
Derivative has not been maintained in 3 years and recently been marked as unmaintained by RUSTSEC (https://rustsec.org/advisories/RUSTSEC-2024-0388).

This PR replaces `derivative` with `derive-where`, a not dead alternative.

### DIFF
--- a/charon/Cargo.lock
+++ b/charon/Cargo.lock
@@ -225,8 +225,8 @@ dependencies = [
  "colored",
  "convert_case 0.6.0",
  "crates_io_api",
- "derivative",
  "derive-visitor",
+ "derive-where",
  "env_logger",
  "flate2",
  "hashlink",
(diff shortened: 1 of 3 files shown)


Nadrieril:
Thank you, this is a welcome change! Could you commit the updated `Cargo.lock` as well?

qinheping:
`Cargo.lock` added in the latest commit.

Nadrieril:
I'm getting "note: an implementation of `PartialEq` might be missing for `ast::types::TraitClause`" errors, something isn't working right.

Nadrieril:
I see, there was a typo. I have fixed this but please run tests before submitting a PR next time.

Nadrieril:
Ended up switching to a manual impl in the end. Should be good, let me know if there are other dependencies that are problematic for kani!

PR #461, opened 2024-11-14
Translate global body as a separate initializer function

Nadrieril:
This changes how we represent globals: instead of carrying a body directly, we instead translate their body as a separate initializer function. The global itself only keeps the id of this function. This reduces duplication in many places.

We were already presenting the llbc in this way on the ocaml side, so this PR doesn't change the LLBC OCaml API; it does change the ULLBC OCaml API however.


charon/src/ast/types.rs
```diff
@@ -883,6 +879,9 @@ pub struct FunSig {
     ///   (the function in which the closure is defined)
     /// - the region variables are local to the closure
     pub is_closure: bool,
+    /// Whether this function is in fact the body of a constant/static that we turned into an
+    /// initializer function.
+    pub is_global_initializer: bool,
```
sonmarcho, line 884: Why is this in the FunSig? It would make more sense to put it in the function itself, as it should be independent of the signature.

Nadrieril: Because it feels like the same kind of thing as "is_closure". I agree that neither really belongs in FnSig tbf

sonmarcho: The `is_closure` parameter here does have an impact on the way we should understand the signature: the way the parameters are quantified is not the same depending on whether the function is a closure or not (this is not clean: I would like to get rid of this, at least through a micro-pass).

Nadrieril: I moved it back to `FunDecl`

PR #463, opened 2024-11-14
Use "builtin" instead of "assumed" in the definition names in Charon-ML

sonmarcho:
Everything is in the title.


Nadrieril:
You have to run `make generate-ml`

PR #469, opened 2024-11-21
Fix issues in the ML substitution functions

sonmarcho:
This is a very minor PR, that I do in preparation of a more important PR.

### DIFF
--- a/charon-ml/src/GAst.ml
+++ b/charon-ml/src/GAst.ml
@@ -51,13 +51,15 @@ and any_decl_id =
 class ['self] iter_statement_base_base =
   object (self : 'self)
     inherit [_] iter_rvalue
-    method visit_abort_kind : 'env -> abort_kind -> unit = fun _ _ -> ()
+    method visit_abort_kind : 'env. 'env -> abort_kind -> unit = fun _ _ -> ()
   end
 
 class ['self] map_statement_base_base =
   object (self : 'self)
     inherit [_] map_rvalue
-    method visit_abort_kind : 'env -> abort_kind -> abort_kind = fun _ x -> x
+
+    method visit_abort_kind : 'env. 'env -> abort_kind -> abort_kind =
+      fun _ x -> x
   end
 
 (** A function operand is used in function calls.
--- a/charon-ml/src/Substitute.ml
+++ b/charon-ml/src/Substitute.ml
@@ -29,13 +29,13 @@ let empty_subst : subst =
     tr_self = Self;
   }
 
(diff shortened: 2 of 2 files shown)


Nadrieril:
Aren't all the other visitors polymorphic in `'env` at the level of the whole visitor? If every method is polymorphic in `'env` that would mean we can't write a visitor that only works with a specific environment. That sounds like a strong restriction, no?

charon-ml/src/GAst.ml
```diff
@@ -51,13 +51,15 @@ and any_decl_id =
 class ['self] iter_statement_base_base =
   object (self : 'self)
     inherit [_] iter_rvalue
-    method visit_abort_kind : 'env -> abort_kind -> unit = fun _ _ -> ()
+    method visit_abort_kind : 'env. 'env -> abort_kind -> unit = fun _ _ -> ()
```
Nadrieril, line 54: You're changing `GAst.ml`, which as explained at the top is automatically-generated. You need to change `charon/src/bin/generate-ml/templates/GAst.ml` instead and run `make generate-ml`.

Nadrieril: Actually this particular visitor is automatically-generated, so you'll have to update the visitor-generating code in `charon/src/bin/generate-ml/main.rs`, specifically `fn generate_visitor_bases`.

charon-ml/src/Substitute.ml
```diff
@@ -29,13 +29,13 @@ let empty_subst : subst =
     tr_self = Self;
   }
 
-let st_substitute_visitor (subst : subst) =
+let st_substitute_visitor =
```
sonmarcho, line 32: This is the problematic definition

charon-ml/src/GAst.ml
sonmarcho: Yes, I'm perfectly aware of that.
I'm currently fighting with a weirdness of OCaml: it seems that because this visitor has only one method, OCaml doesn't manage to see that the `'env` type variable here should be the same as in the methods inherited from the `rvalue` visitors. It leads to an error when type-checking `st_substitute_visitor`.

sonmarcho: The solution is to move the `abort_kind` visit methods to the ULLBC visitors, which actually makes more sense (cf the commit I just pushed).

charon-ml/src/Substitute.ml
sonmarcho: Another solution would be to overwrite the `visit_abort_kind` method.

PR #471, opened 2024-11-22
Change `.ocamlformat` to match aeneas

### LINKED ISSUE #470: Make the .ocamlformat files of Charon and Aeneas agree
The .ocamlformat files in Charon-ML and Aeneas do not agree. The consequence is that when using a symlink to Charon, calling `dune fmt` in Aeneas often reformats the Charon files, leading to unexpected diffs. We should use the same options for both projects (the options to be used are probably the ones from Aeneas).

I believe this is the reason why the `template` files also regularly get reformatted when working locally.

Nadrieril:
Fixes #470.


Nadrieril:
@sonmarcho this is a good time for me to do that, I don't have many wip changes that would be broken by this. Is that good for you too?

sonmarcho:
Yes, that's good for me: I have a PR in the making but it barely touches the Charon-ML.

PR #472, opened 2024-11-22
Change the order of item exploration

Nadrieril:
We now start with the local crate instead of sorting by item type. This means we'll generally translate all items from the current crate before attempting to translate an external item. This should make errors easier to follow.


Nadrieril:
I expected this to change aeneas tests! Lucky.

PR #473, opened 2024-11-22
Improve error reporting on external dependencies

Nadrieril:
When an external item fails to translate we indicate which local items (transitively) caused the external item to attempt to be translated. Until now, we listed that at the end of translation, i.e. after all errors. This PR makes it so we indicate this right below the first error coming from the failed translated item.

This makes errors a lot easier to follow and avoids having to scroll through a big wall of "The external definition `...` triggered errors" at the end of the charon output.


sonmarcho:
Thanks for this PR: it is most useful!

PR #476, opened 2024-11-27
Use de Bruijn indices for all type-level variables

### LINKED ISSUE #361: Use DeBruijn indices for all type-level variables
A few features involve nested binders:
- `for<'a>` clauses and function pointer types;
- GATs;
- `dyn Trait`, which I would like to encode as `exists<T> T: Trait<Vars..>`;
- closures, where I'd like to have one binder level per parent (to handle nested closures gracefully);
- trait methods, where I would like to explicitly separate the parameters from the parents from those of the method; this is ilkely necessary for provided methods.

Today we use binder groups + DeBruijn indices (just like rustc) for lifetime parameters. To support all these features cleanly I would like to do the same for type, const and trait clause parameters.

Nadrieril:
Fixes #361.

### DIFF
--- a/charon-ml/src/Expressions.ml
+++ b/charon-ml/src/Expressions.ml
@@ -278,7 +278,7 @@ and raw_constant_expr =
           Remark: trait constants can not be used in types, they are necessarily
           values.
        *)
-  | CVar of const_generic_var_id  (** A const generic var *)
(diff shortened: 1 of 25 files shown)


Nadrieril:
This was done differently in other PRs

PR #479, opened 2024-11-29
Add some features

sonmarcho:
I need this for a PR on Aeneas' side

### DIFF
--- a/charon/src/bin/generate-ml/main.rs
+++ b/charon/src/bin/generate-ml/main.rs
@@ -1395,9 +1395,9 @@ fn generate_ml(
                 ]),
                 (GenerationKind::TypeDecl(Some(DeriveVisitors {
                     name: "type_decl",
-                    ancestor: Some("type_decl_base"),
+                    ancestor: Some("generic_params"),
                     reduce: false,
-                    extra_types: &[],
+                    extra_types: &["attr_info", "item_meta"],
                 })), &[
                     "Field",
                     "Variant",
--- a/charon/src/bin/generate-ml/main.rs
+++ b/charon/src/bin/generate-ml/main.rs
@@ -1202,31 +1202,54 @@ fn generate_ml(
             template: template_dir.join("GAst.ml"),
             target: output_dir.join("GAst.ml"),
             markers: ctx.markers_from_names(&[
+                (GenerationKind::TypeDecl(None), &["FunDeclId"]),
                 (GenerationKind::TypeDecl(Some(DeriveVisitors {
-                    name: "statement_base",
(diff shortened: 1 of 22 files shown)


charon/src/bin/generate-ml/templates/Types.ml
```diff
@@ -167,6 +167,22 @@ class virtual ['self] map_ty_base_base =
...
+    inherit [_] iter_generic_params
+    method visit_attr_info : 'env -> attr_info -> unit = fun _ _ -> ()
+    method visit_item_meta : 'env -> item_meta -> unit = fun _ _ -> ()
+  end
+
+class ['self] map_type_decl_base =
+  object (self : 'self)
+    inherit [_] map_generic_params
+    method visit_attr_info : 'env -> attr_info -> attr_info = fun _ x -> x
+    method visit_item_meta : 'env -> item_meta -> item_meta = fun _ x -> x
```
Nadrieril, line 181: I think you can get those to be auto-generated by adding `extra_types: &["attr_info", "item_meta"]` at the right spot

charon/src/bin/generate-ml/main.rs
```diff
@@ -1369,6 +1392,13 @@ fn generate_ml(
                     "PathElem",
                     "Name",
                     "ItemMeta",
+                ]),
+                (GenerationKind::TypeDecl(Some(DeriveVisitors {
+                    name: "type_decl",
+                    ancestor: Some("type_decl_base"),
+                    reduce: false,
+                    extra_types: &[],
```
Nadrieril, line 1400: I think this would avoid the need for the manual intermediate visitor:
```suggestion
                    ancestor: Some("generic_params"),
                    reduce: false,
                    extra_types: &["attr_info", "item_meta"]
```

sonmarcho:
I had to introduce several auxiliary visitors for definitions like `global_decl`, `trait_decl`, etc. because of a limitation of the visitor library: we can't derive visitors for a group of definitions containing structures with fields which have the same name (for instance `global_decl` and `trait_decl` both have a field named `def_id`, and if we try to derive one visitor for both definitions at once, we get a type error).

charon/src/bin/generate-ml/main.rs
sonmarcho: Ah yes indeed

sonmarcho:
Also, this PR introduces an `LlbcAstUtils.iter_crate_with_span` visitor which keeps track of the declaration id and of the last (most precise) encountered span when visiting declarations.

Nadrieril (approved):

Issue #287, opened 2024-07-09
Name things consistently and idiomatically


Nadrieril:
Naming could be improved across the project. Among other things:
- "assumed" should be renamed to "built-in";
- `Id` is a confusing suffix for a rust user: it normally designates integer identifiers only. Similarly, `Raw` is used in places where a rust user would naturally use a `Kind` suffix, e.g. `TerminatorKind`;
- Traits and clauses and predicates are subtle things, I think better names would help follow what's happening. E.g. `TraitInstanceRef` instead of `TraitRef`, `TraitClause` or `TraitBound` instead of `TraitDeclRef`, something else instead of `TraitClause`.
- Names don't always match between the rust and ocaml sides. Once we're happy with the rust side we should propagate the changes.

Nadrieril:
My current list:
- [x] `Var` -> `Local`: https://github.com/AeneasVerif/charon/pull/645
- [x] `RawStatement` -> `StatementKind` https://github.com/AeneasVerif/charon/pull/805
- [x] `RawTerminator` -> `TerminatorKind` https://github.com/AeneasVerif/charon/pull/805
- [x] `RawConstantExpr` -> `ConstantExprKind` https://github.com/AeneasVerif/charon/pull/805
- [x] `Statement.content` -> `Statement.kind`, `Terminator.content -> Terminator.kind`, `ConstantExpr.value` -> `ConstantExpr.kind`
- [x] `FunIdOrTraitMethodRef` -> `FnPtrKind`, `FnPtr.func` -> `FnPtr.kind`
- [x] `RawSpan` -> `SpanData`, `Span.span` -> `Span.data`
- [x] `TypeVar`/`RegionVar`/`ConstGenericVar` -> `TypeParam`/`RegionParam`/`ConstGenericParam`
- [x] `TraitClause` -> `TraitParam`
- [x] `AnyTransId` -> `ItemId`, `AnyTransItem` -> `ItemRef`, `AnyTransItemMut` -> `ItemRefMut`
- [ ] `DeBruijnVar` -> `Var`? (this one is likely to cause clashes in aeneas)
- [x] `ItemKind` -> `ItemSource`, `FunDecl.kind` -> `FunDecl.src`, `GlobalDecl.kind` -> `GlobalDecl.src`
- [x] `TraitDecl.parent_clauses` -> `TraitDecl.implied_clauses`, `TraitImpl.parent_trait_refs` -> `TraitImpl.implied_trait_refs`
[...]

PR #486, opened 2024-12-06
More preparations for #361

Nadrieril:


charon/src/ast/types/vars.rs
```diff
@@ -0,0 +1,172 @@
...
+    Ord,
+    Serialize,
+    Deserialize,
+    Drive,
+    DriveMut,
+)]
+#[serde(transparent)]
+pub struct DeBruijnId {
+    pub index: usize,
+}
```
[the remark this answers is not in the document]
Nadrieril: In my mind the `generate_index_type!` types are meant for indexing into a `Vector` we never manipulate the underlying integer, they may as well be opaque. Here I keep manipulating the integer value as an integer, so it feels different. Note in particular that there's no `impl Idx for DeBruijnId`, which is the main purpose of `generate_index_type!`.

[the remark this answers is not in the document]
sonmarcho: Yes, that makes sense.

charon/src/ast/types/vars.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub span: Option<Span>,
+    /// Where the predicate was written, relative to the item that requires it.
+    #[charon::opaque]
+    pub origin: PredicateOrigin,
+    /// The trait that is implemented.
+    #[charon::rename("trait")]
+    pub trait_: PolyTraitDeclRef,
+}
+
+pub type RegionDbVar = DeBruijnVar<BoundRegionId, FreeRegionId>;
```
sonmarcho, line 116: I just realized I'm not super happy with the name `DeBruijnVar`: to me a de Bruijn var is necessarily a bound var, while here it actually refers to a bound var or a free var. I'm aware the name `Var` is already taken. Maybe we should use `VarDecl` (for the current `Var` type) and `Var` (for the current `DeBruijnVar`)?

charon/src/ast/types/vars.rs
```diff
@@ -0,0 +1,172 @@
...
+///                                   Var id: 1
+/// ```
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum DeBruijnVar<Bound, Free> {
+    /// A variable attached to the nth binder, counting from the inside.
+    Bound(DeBruijnId, Bound),
+    /// A variable attached to an implicit binder outside all other binders. This is not present in
+    /// translated code, and only provided as a convenience for convenient variable manipulation.
+    Free(Free),
+}
```
sonmarcho, line 57: I just realized I'm not super happy with the name `DeBruijnVar`: to me a de Bruijn var is necessarily a bound var, while here it actually refers to a bound var or a free var. I'm aware the name `Var` is already taken. Maybe we should use `VarDecl` (for the current `Var` type) and `Var` (for the current `DeBruijnVar`)?

sonmarcho: (and yes, it does make sense to see a free var as bound outside all binders: it's just not the standard interpretation I believe)

Nadrieril: I was thiking the same. The current `Var` should be called `Local` anyway, I thought I had renamed it already.

Nadrieril: I'll add it to the [list of renames](https://github.com/AeneasVerif/charon/issues/287)

PR #494, opened 2024-12-11
Translate correct closure signature

Nadrieril:
This fixes the signature we use for closures. This also fixes a crash when attempting to print crates using the charon-ml pretty-printer.


sonmarcho:
Thanks! This has been much needed.

PR #497, opened 2024-12-12
Some cleanups

Nadrieril:
Some cleanups after the various changes related to #361.

### DIFF
--- a/charon-ml/src/Substitute.ml
+++ b/charon-ml/src/Substitute.ml
@@ -7,6 +7,9 @@ open TypesUtils
 open Expressions
 open LlbcAst
 
+(* TODO: use Core.Fn.compose *)
+let compose f g x = f (g x)
+
 (* A substitution that takes a full variable as input *)
 type subst = {
   r_subst : RegionId.id de_bruijn_var -> region;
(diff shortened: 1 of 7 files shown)


charon-ml/src/Substitute.ml
```diff
@@ -69,15 +101,37 @@ let subst_at_binder_zero (subst : single_binder_subst) : subst =
     tr_self = subst.tr_sb_self;
   }
 
+(** Substitute the variables bound by the current (level 0) binder, and shift variables to remove the current binder level. *)
+let subst_remove_binder_zero (subst : single_binder_subst) : subst =
+  let subst_remove_zero subst nosubst = function
+    | Bound (0, id) -> subst id
+    | Bound (dbid, varid) when dbid > 0 -> nosubst (Bound (dbid - 1, varid))
```
sonmarcho, line 108: Nit: you actually don't need the `when dbid > 0`

Nadrieril: Yes I do, if we're already inside a binder when we start substituting, variables pointing to binders in-scope when we start will get negative indices when we dig inside other binders. Only the variables with positive indices correspond to binders deeper than where we started and thus that need to be shifted.

sonmarcho: Oh indeed, I forgot about the negative binders. That would probably be worth a comment in the future.

PR #500, opened 2024-12-17
Add minor utilities

sonmarcho:
I need those on Aeneas' side.


Nadrieril:
I think we should move `region_var_group` and related helpers to aeneas, it's not really a concept relevant to charon.

Issue #194, opened 2024-05-22
Clarify the encoding of closures


Nadrieril:
As of https://github.com/AeneasVerif/charon/pull/186, closures are implemented by letting the consumer of llbc track which function pointer goes with which closure, and the `Call` terminator mentions a built-in trait implementation for `Fn`/`FnMut`/`FnOnce` that the consumer of llbc must figure out by themselves.

I propose that we instead encode this:
```rust
pub fn test_closure_capture(x: u32, y: u32) -> u32 {
    let f = &|z| x + y + z;
    (f)(0)
}
```
as follows:
```rust
struct {test_closure_capture::closure#0} {
    x: u32,
    y: u32,
}

impl Fn<(u32,)> for {test_closure_capture::closure#0} {
    type Output = u32;
    fn call(&self, arg: (u32,)) -> u32 {
        self.x + self.y + arg.0
    }
}

pub fn test_closure_capture(x: u32, y: u32) -> u32 {
    let state = {test_closure_capture::closure#0} { x, y };
    state.call(0)
}
```

This is pretty much exactly how rustc represents closures internally, and has the benefit that consumers of llbc don't need to do anything special to support closures.

sonmarcho:
Small update: after a discussion I think we came up with a different solution, where the function pointer is not stored in the state but in the instance of `Fn<(u32,)>` (maybe it is actually what you mean above, but it is slightly unclear).

Nadrieril:
That is indeed what I meant above; as you can see, there is no function pointer stored anywhere in the desugared output.

ssyram:
Hi, I'm also curious about how is this issue now. I find the naive MIR encoding of closures are quite non-sense... They seem to invent a new currying syntax in MIR especially for closure, which, unfortunately, affects how LLBC generates now... I cannot see an explicit struct for the closure object, but instead syntax like that seen in the MIR:

```MIR
_8 = {closure@[src/lib.rs:16:25: 16:33]} { y: const 100_i32, z: copy _1 };
```

And the corresponding LLBC is like:

```LLBC
@8 := {project::test_closure::closure} {copy (y@2), copy (some@1)}
```

ssyram:
[...]

PR #510, opened 2024-12-31
Check generics inside bodies too

Nadrieril:
Until now, the `check_generics` pass looked at items but not at function bodies. Turns out there were quite a few issues with generics inside function bodies. This PR fixes most of them, except closure-related ones which I'm leaving for #194. This builds on https://github.com/hacspec/hax/pull/1215.


Nadrieril:
I'm merging with a `Cargo.toml` pointing to https://github.com/hacspec/hax/pull/1215 which is not merged because I'd like to keep making progress while everyone is on vacation.

Issue #180, opened 2024-05-14
Translate defaulted trait methods implementations


Nadrieril:
Provided trait methods are tricky to handle for consumers of charon: they can be individually overriden, and mutually recursive with other trait methods (overriden or not). To simplify their task, charon could hide the difference between provided and required trait methods.

I see the following steps:
1. Duplicate all provided methods, so that any given trait impl specifies all trait methods;
2. Deduplicate simple cases, by sharing a single method definition for all trait impls that don't override it;
3. Deduplicate more cases, by defining shared methods that take function pointers as inputs.

Step 3 is necessary for e.g. `Iterator`, which has many provided methods all defined in terms of `self.next()`.

msprotz:
I recently tracked down an issue in compiling core::cmp::min which boiled down to exactly this. At call-site, the trait impl defines only one required method, the others are provided -- but I am expecting even provided methods to be provided at call-site by the trait impl, because I don't know how to resolve potential overrides.

Nadrieril:
Current plan is to simply reuse the default method by passing it the right arguments, which is enabled by https://github.com/AeneasVerif/charon/pull/512 and https://github.com/AeneasVerif/charon/pull/514. We may later duplicate methods to remove dependency cycles if needed.

sonmarcho:
> Current plan is to simply reuse the default method by passing it the right arguments, which is enabled by #512 and #514. We may later duplicate methods to remove dependency cycles if needed.

Following your plan what would be the LLBC of the following code?
```Rust
trait Trait { fn f(&self); fn g(&self) { self.f() }

impl Trait for i32 { fn f(&self) { self.g() }
impl Trait for u32 { fn f(&self) { self.g() } fn g(&self) { self.f() } }
```

Nadrieril:
It would be:
- Trait decl
```rust
trait test_crate::Trait<Self>
{
    fn f = test_crate::Trait::f<Self>[Self]
    fn g = test_crate::Trait::g<Self>[Self]
}

fn test_crate::Trait::f<Self>(@1: Self)
where
    [@TraitClause0]: test_crate::Trait<Self>,

fn test_crate::Trait::g<Self>(@1: Self)
where
    [@TraitClause0]: test_crate::Trait<Self>,
{
    @0 := @TraitClause0::f(move (@2))
    return
}
```
- Impl for i32:
```rust
fn test_crate::{impl test_crate::Trait for i32}::f(@1: i32)
{
[...]

PR #514, opened 2025-01-03
Make `Self` clause explicit in trait method declarations

Nadrieril:
For #180 I want trait methods to be completely non-magical functions. https://github.com/AeneasVerif/charon/pull/512 made huge progress towards this, but I had forgotten about `Self` clauses.

Within a trait declaration (including its associated items), we can use `TraitRefKind::SelfId` to refer to the ambiant `Self: ThisTrait` clause. This is used e.g. to recursively call other trait methods, or access the value of the `Self::Type` associated types. Trait implementations don't need this, because they always know the correct method/type to use directly.

In order for methods to be normal functions, I needed to remove uses of this implicit clause. This PR instead makes the clause explicit in the `FunDecl` that corresponds to each method declaration. Note that this does not affect method _signatures_: the clauses needed to call a method don't change. The special clause is passed when the method is bound in the `TraitDecl` (which we can express thanks to https://github.com/AeneasVerif/charon/pull/512).


sonmarcho:
I think this goes in the right direction, but there are some details I don't fully get. In particular I'm not sure I understand: "The special clause is passed when the method is bound in the `TraitDecl`": could you give an example?

Nadrieril:
Here's an example inspired from the tests diff:
```rust
trait test_crate::Trait<Self, B>
{
    parent_clause0 : [@TraitClause0]: core::marker::Sized<B>
    parent_clause1 : [@TraitClause1]: core::marker::Sized<Self::Item>
    type Item
    fn method<C, [@TraitClause0]: core::marker::Sized<C>> = test_crate::Trait::method<Self, B, C>[Self, @TraitClause0_0]
    // before this PR:
    // fn method<C, [@TraitClause0]: core::marker::Sized<C>> = test_crate::Trait::method<Self, B, C>[@TraitClause0_0]
}

fn test_crate::Trait::method<Self, B, C>() -> @TraitClause0::Item // was `-> Self::Item` before
where
    [@TraitClause0]: test_crate::Trait<Self, B>, // added by this PR
    [@TraitClause1]: core::marker::Sized<C>,
```
As you can see, since https://github.com/AeneasVerif/charon/pull/512 in a trait declaration/implementation a method is not just a `FunDeclId` it's a map from the method arguments to the arguments to pass to the function item. What this PR does is that the function item for a method declaration now has an extra clause argument. The implicit `Self` clause is no longer available in function items, all references to `Self` now use the new clause.

sonmarcho:
Ok, thanks! So we agree that now all trait method declarations are mutually recursive with the trait itself? Doesn't that cause issues on the Aeneas' side?

Nadrieril:
> So we agree that now all trait method declarations are mutually recursive with the trait itself?

Yep

> Doesn't that cause issues on the Aeneas' side?

I don't know yet, this requires more binder-related fixes on the aeneas side first. If it does cause issues I hope that'll be only for provided methods (which I'll be fixing next), we'll see.

sonmarcho:
But we agree that *implementations* of trait methods don't change (in particular their signature)?

Nadrieril:
> But we agree that implementations of trait methods don't change

Correct

> I don't fully understand this sentence: "Note that this does not affect method signatures: the clauses needed to call a method don't change."

I lack good terminology here. The important property about the two-binding-levels-for-methods I introduced recently is that the `Binder<FunDeclRef>` for a particular method in a `TraitDecl` binds exactly the same generics as the `Binder<FunDeclRef>`for that same method in a `TraitImpl` for that trait. These are exactly the generics that must be supplied by the `GenericArgs` of a `FnPtr::Trait` pointing to that method.

The note I was trying to make is that these generics are unchanged by this PR. The thing that _is_ changed by this PR is the generics of the `FunDecl`s that correspond to method declarations.

sonmarcho:
> > But we agree that implementations of trait methods don't change
> 
> Absolutely
> 
> > I don't fully understand this sentence: "Note that this does not affect method signatures: the clauses needed to call a method don't change."
> 
> I lack good terminology here. The important property about the two-binding-levels-for-methods I introduced recently is that the `Binder<FunDeclRef>` for a particular method in a `TraitDecl` binds exactly the same generics as the `Binder<FunDeclRef>`for that same method in a `TraitImpl` for that trait. These are exactly the generics that must be supplied by the `GenericArgs` of a `FnPtr::Trait` pointing to that method.
> 
> The note I was trying to make is that these generics are unchanged by this PR. The thing that _is_ changed by this PR is the generics of the `FunDecl`s that correspond to method declarations. Which is the only sensible thing to do anyway, maybe this note added more confusion than it removed ^^.

This is clear and makes sense to me, thanks :)

Issue #477, opened 2024-11-28
Support for Generalized Associated Types


R1kM:

Nadrieril:
There's a common trait in the rust-crypto ecosystem that has GATs. But this was just an issue to track known unsupported features.

PR #515, opened 2025-01-06
Improve the `check_generics` pass

Nadrieril:
This does a few things to improve the `check_generics` pass:
- We now track for each `GenericArgs` what item it applies to. This greatly simplifies the pass and will be a tremendous help for #127;
- We pretty-print the llbc before running the pass, which makes it possible to inspect the llbc output even when it faile;
- We now track which items have been visited so that errors have more context to them.

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.60"
+let supported_charon_version = "0.1.61"
--- a/charon-ml/src/generated/Generated_GAst.ml
+++ b/charon-ml/src/generated/Generated_GAst.ml
@@ -113,14 +113,6 @@ and call = { func : fn_operand; args : operand list; dest : place }
  *)
 and assertion = { cond : operand; expected : bool }
 
-(** The id of a translated item. *)
-and any_decl_id =
-  | IdType of type_decl_id
-  | IdFun of fun_decl_id
-  | IdGlobal of global_decl_id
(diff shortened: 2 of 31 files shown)


Nadrieril:
Other features I'd like to eventually add to this pass:
- Check that each generic variable reference points to something;
- Check the number and type of function arguments in function calls (will likely require [fixing closures](#194)  first);
- Check that referenced associated types actually exist (I'm waiting for #477);
- Check the number and type of aggregate fields (also waiting for closure fixes).

sonmarcho:
Checking that I understand what you are doing: the `GenericSource` tracks the item owning the generic args, correct?
For instance, if I have `Array::index<T, N>` then the generic source of `<T, N>` is `Builtin`, while it is `Item` if I have `Vec<T>`, right?
Whenever I need this kind of information in a micro-pass, I implement a visitor which carries an (optional) environment around, by overriding the proper methods to update the environments whenever I, say, visit a `Call`, `Adt` construct, etc. I'm wondering: why didn't you make that choice (it may be more straightforward/natural to do in OCaml of course)?

Nadrieril:
> For instance, if I have Array::index<T, N> then the generic source of <T, N> is Builtin, while it is Item if I have Vec<T>, right?

Correct

> Whenever I need this kind of information in a micro-pass, I implement a visitor which carries an (optional) environment around, by overriding the proper methods to update the environments whenever I, say, visit a Call, Adt construct, etc. I'm wondering: why didn't you make that choice (it may be more straightforward/natural to do in OCaml of course)?

This is what I did before, you can see the diff in the PR. The problem is that this requires manually listing every spot where a `GenericArgs` might be, which is error-prone and annoying to maintain. This would have been fine if it was just for this pass, but #127 requires something similar and even more subtle, hence this imo much cleaner solution.

sonmarcho:
That makes sense, thanks for the explanations!

Issue #507, opened 2024-12-23
Bug: faulty control-flow reconstruction in ML-DSA


msprotz:
Found while looking at something unrelated for @franziskuskiefer 

If I dump the LLBC I receive from Charon while running `cd libcrux/libcrux-ml-dsa && ./boring.sh` on https://github.com/cryspen/libcrux/tree/franziskus/mldsa-c-ci (commit 26c593e) I get the following:

```rust
fn libcrux_ml_dsa::encoding::signature::{libcrux_ml_dsa::encoding::signature::Signature<SIMDUnit, COMMITMENT_HASH_SIZE, COLUMNS_IN_A, ROWS_IN_A>[TraitClause@0, TraitClause@1]}::deserialize<0, SIMDUnit, COMMITMENT_HASH_SIZE, COLUMNS_IN_A, ROWS_IN_A, GAMMA1_EXPONENT, GAMMA1_RING_ELEMENT_SIZE, MAX_ONES_IN_HINT, SIGNATURE_SIZE>(serialized^1 : &0 (@Array<u8, SIGNATURE_SIZE>)) -> core::result::Result<libcrux_ml_dsa::encoding::signature::Signature<SIMDUnit, COMMITMENT_HASH_SIZE, COLUMNS_IN_A, ROWS_IN_A>[TraitClause@0, TraitClause@1], libcrux_ml_dsa::types::VerificationError>[core::marker::Sized<libcrux_ml_dsa::encoding::signature::Signature<SIMDUnit, COMMITMENT_HASH_SIZE, COLUMNS_IN_A, ROWS_IN_A>[TraitClause@0, TraitClause@1]>, core::marker::Sized<libcrux_ml_dsa::types::VerificationError>]
  where
    [TraitClause@0]: core::marker::Sized<SIMDUnit>,
    [TraitClause@1]: libcrux_ml_dsa::simd::traits::Operations<SIMDUnit>,  
{
// ... lots of stuff ...
    loop {
      v@45 := copy i^41;
      v@44 := move v@45 < ROWS_IN_A;
      if (move v@44) {
        drop v@45;
        v@46 := copy malformed_hint^42;
        if (move v@46) {
          break 0
        }
        else {
          v@50 := copy i^41;
          v@49 := MAX_ONES_IN_HINT + move v@50;
          drop v@50;
          v@153 := &*(hint_serialized^8);
          v@154 := move @SliceIndexShared<'_, u8>(move v@153, copy v@49);
[...]

PR #520, opened 2025-01-10
Improve the CFG reconstruction

sonmarcho:
@protz I believe this solves https://github.com/AeneasVerif/charon/issues/507 - could you confirm?
Note that I added the minimized examples as tests, but I don't know how to check that the ML-DSA examples is properly reconstructed.

### DIFF
--- a/charon/src/transform/ullbc_to_llbc.rs
+++ b/charon/src/transform/ullbc_to_llbc.rs
@@ -969,6 +969,48 @@ fn compute_switch_exits_explore(
     info
 }
 
+/// Auxiliary helper
+///
+/// Check if it is possible to reach the exit of an outer switch from [bid]
+/// without going through the [exit_candidate]. We use the graph without
+/// backward edges.
+fn can_reach_outer_exit(
+    cfg: &CfgInfo,
+    outer_exits: &HashSet<src::BlockId>,
+    start_bid: src::BlockId,
+    exit_candidate: src::BlockId,
+) -> bool {
+    // The stack of blocks
+    let mut stack: Vec<src::BlockId> = vec![start_bid];
+    let mut explored: HashSet<src::BlockId> = HashSet::new();
+
+    while let Some(bid) = stack.pop() {
+        // Check if already explored
+        if explored.contains(&bid) {
+            break;
+        }
+        explored.insert(bid);
+
+        // Check if this is the exit candidate
+        if bid == exit_candidate {
+            // Stop exploring
+            break;
+        }
+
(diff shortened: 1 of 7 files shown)


sonmarcho:
@protz I believe this solves https://github.com/AeneasVerif/charon/issues/507 - could you confirm?
Note that I added the minimized examples as tests, but I don't know how to check that the ML-DSA examples is properly reconstructed.

charon/src/transform/ullbc_to_llbc.rs
```diff
@@ -856,117 +872,86 @@ fn compute_switch_exits_explore(
...
+        acc.insert(make_ord_block_id(s.id, tsort_map));
+        acc
+    });
 
-            // Update the best intersection, if necessary
-            if current_number_inter > max_number_inter {
-                max_number_inter = current_number_inter;
-                max_inter_succs = current_inter_succs;
+    // Compute the "flows" (if there are children)
+    // TODO: this is computationally expensive...
```
Nadrieril, line 904: That should be linear in the size of the graph, no?

Nadrieril (approved):

charon/src/transform/ullbc_to_llbc.rs
sonmarcho: This is linear *for each node*.
And the size of the fraction is also linear in the total size of the graph...

Nadrieril: Can you not traverse the DAG in topological order to compute the flow in linear time in the size of the graph?

sonmarcho: The flow at a given node unfortunately depends on the node on which you start:
```
A
|  \
B C
|   |
|  D
|  /
E
```
The flow at `D` is 0.5 if you start from `A` and `1` if you start from `D`. So in the end we need to remember `V^2` values though there is probably a smart way of not remembering all values).

sonmarcho: For instance, I guess what's really important is the flow at the join nodes (i.e., starting from `C` we don't need to remember the flow at `D`, what really matters is the flow at `E` as there is more than one incoming edge in `E`).

Nadrieril: Ah I see, we need to compute it once for each entry node, right? I understand the comment now.

sonmarcho: Yes, exactly

PR #524, opened 2025-01-14
Simplify error reporting

Nadrieril:
This puts `ErrorCtx` under a `RefCell`, which removes the need for `&mut` references for error reporting. This makes it possible to simplify the error macros a bunch. While I was at it, I made them into proper formatting macros so we don't need to call `format!` when using them.


sonmarcho:
Oh great!! It was really annoying to carry the error context everywhere

PR #528, opened 2025-01-20
Update hax

Nadrieril:
Companion PR to https://github.com/cryspen/hax/pull/1237 and https://github.com/cryspen/hax/pull/1236.

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -124,9 +124,18 @@ pub enum TraitRefKind {
     #[charon::rename("Self")]
     SelfId,
 
-    /// A specific builtin trait implementation like [core::marker::Sized] or
-    /// auto trait implementation like [core::marker::Syn].
-    BuiltinOrAuto(PolyTraitDeclRef),
+    /// A trait implementation that is computed by the compiler, such as for built-in traits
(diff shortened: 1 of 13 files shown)


charon/src/ast/types.rs
```diff
@@ -124,9 +124,18 @@ pub enum TraitRefKind {
...
-    /// auto trait implementation like [core::marker::Syn].
-    BuiltinOrAuto(PolyTraitDeclRef),
+    /// A trait implementation that is computed by the compiler, such as for built-in traits
+    /// `Sized` or `FnMut`. This morally points to an invisible `impl` block; as such it contains
+    /// the information we may need from one.
+    BuiltinOrAuto {
+        trait_decl_ref: PolyTraitDeclRef,
+        /// The `ImplExpr`s required to satisfy the implied predicates on the trait declaration.
+        /// E.g. since `FnMut: FnOnce`, a built-in `T: FnMut` impl would have an `ImplExpr` for `T:
+        /// FnOnce`.
```
sonmarcho, line 134: I think it would be worth adding as a comment that in the case of non-builtin traits all the information we need is in the `impl` item that we can lookup, but in the case of builtin traits we don't have such an impl item which is why we put all the information like the parent trait clauses and the types directly in the trait ref.

Issue #531, opened 2025-01-21
Bug: type aliases don't work with `--remove-associated-type`


Nadrieril:
Type aliases in rust are glorified macros: they are not checked for well-formedness and as such don't require the proper trait clauses to be in scope:
```rust
trait Trait { type Output; }
type Type<F> = <F as Trait>::Output;
```
This already requires hack today, but with `--remove-associated-type` it's worse. Solutions include:
- Putting type aliases behind an option and error if both options are on;
- Add the missing trait clauses automatically somehow.

Issue #513, opened 2025-01-02
Bug: incorrect generics when implemented method has more specific signature


Nadrieril:
While working on https://github.com/AeneasVerif/charon/pull/512, I discovered that there are two ways that an implemented method may have a different signature than expected from the trait:
- the implemented method `fn<..> A -> B` type (including late-bound vars) may be a subtype of that of the declared method;
- the implemented method may have less strict trait bounds.

```rust
trait Trait: Sized {
    fn method1(self, other: &'static u32);
    fn method2<T: Copy>(self, other: T); 
}

impl Trait for () {
    // This implementation is more general because it works for non-static refs.
    fn method1(self, _other: &u32) {}
    // This implementation is more general because it works for non-`Copy` `T`s.
    fn method2<T>(self, _other: T) {}
}

fn main() {
    let _ = ().method1(&1u32);
    // Error: we pass incorrect predicates
    // let _ = ().method2(false);
}
```

In both cases our translation is incorrect. Once https://github.com/AeneasVerif/charon/pull/512 lands we'll be able to fix this by using the method binders to convert between the trait-declared method generics and the actually-implemented method generics.

Nadrieril:
https://github.com/hacspec/hax/pull/1215 fixes some of the first kind of error but not yet all. I left [a hack](https://github.com/Nadrieril/hax/blob/79e2da0452a0b94056becba46b98304ca230a502/frontend/exporter/src/types/new/full_def.rs#L808-L815) that we'll be able to remove after #127.

Nadrieril:
This happens in particular in the stdlib, for the `Iterator::fold` impl of `slice::Iter`.

Nadrieril:
Relevant rustc functions: `compare_impl_item`, `compare_method_predicate_entailment`.

Nadrieril:
The part about trait clauses was fixed in #1224. For the part about late-bound lifetimes, we generally don't handle them well, so I'll deref to https://github.com/AeneasVerif/charon/issues/1237.

Issue #123, opened 2024-04-12
Add support for `dyn Trait`


Nadrieril:

Nadrieril:
I investigated a bit. The issue is that we need to handle [ExistentialPredicate](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_type_ir/predicate/enum.ExistentialPredicate.html) which isn't quite a `TraitRef`. If I understand, after https://github.com/AeneasVerif/charon/issues/127 we might be able to use a `TraitRef` there.

sonmarcho:
What do you mean by "we might be able to use a `TraitRef` there?

Nadrieril:
`dyn Trait` refers to a trait with an [ExistentialPredicate](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_type_ir/predicate/enum.ExistentialPredicate.html), which seems to be a mapping from `Self` to a list of clauses. E.g. `dyn Iterator<Item=u32>` would be represented as `lambda Self -> Self: Iterator, Self::Item = u8`. If we do #127, I _think_ this simplifies to just a list of trait clauses on `Self`, which we can represent as a list `TraitRef`s.

sonmarcho:
I think a dyn trait really existentially quantifies a type, and we need to represent that: `lambda Self -> Self: Iterator, Self::Item = u8` is a predicate over some type `Self`, but must itself be quantified. If you have a function `fn foo(...) -> dyn Trait` in Lean we would generate something of the shape:
```lean
def foo ... : exists T, Trait T
```
or, written with a dependent tuple (we return a pair of a type `T` and the fact that it satisfies `Trait T`):
```lean
def foo ... : (T:Type, Trait T)
```

Nadrieril:
Yes, but with #127 the existentially quantified predicate will always be of the form `lambda Self -> Self: Trait1, Self: Trait2`, which we can represent as a list `[Trait1, Trait2]`

Nadrieril:
I was wrong about that: a `TraitRef` looks like `T: Into<u64>`, but what we need for `dyn` is just the `Into<u64>` part. In other words, it's a `TraitRef` where the `Self` argument is not supplied. In yet other words, it's an existentially quantified `TraitRef`.

Nadrieril:
Partial support here: https://github.com/AeneasVerif/charon/pull/276. Needs more work to support fully.

Nadrieril:
[...]

PR #532, opened 2025-01-21
Transform associated types into type parameters

### LINKED ISSUE #127: Hide associated types equality constraints

Nadrieril:
This PR introduces a `--remove-associated-types <name pattern>` option that replaces the associated types of the selected traits and converts them into type parameters.

Fixes #127.

This has 4 know limitations today:
- https://github.com/AeneasVerif/charon/issues/531;
- https://github.com/AeneasVerif/charon/issues/513;
- Doesn't yet work with `dyn Trait` (https://github.com/AeneasVerif/charon/issues/123);
- We erase some lifetimes in quantified clauses like `T: for<'a> Trait<'a, Type = &'a ()>` (https://github.com/AeneasVerif/charon/issues/534).

This also has three fundamental limitations that we can't work around:
- We can't handle GATs, as this would require passing type-level functions as type parameters;
- We can't handle `for<'a>` clauses with unconstrained associated types, for the same reason;
- We can't add parameters to self-recursive traits, as this would require an infinite list of parameters.

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.63"
+let supported_charon_version = "0.1.64"
--- a/charon-ml/src/generated/Generated_GAst.ml
+++ b/charon-ml/src/generated/Generated_GAst.ml
@@ -417,6 +417,8 @@ and cli_options = {
       (** Blacklist of items to keep opaque. These use the name-matcher syntax. *)
   exclude : string list;
       (** Blacklist of items to not translate at all. These use the name-matcher syntax. *)
+  remove_associated_types : string list;
+      (** List of traits for which we transform associated types to type parameters. *)
   hide_marker_traits : bool;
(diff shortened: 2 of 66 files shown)


sonmarcho:
This is really awesome to see this finally landing, thanks a lot for this! And specifying which traits this should be applied on by using a pattern is a good design choice.

I have one question: by `<'a> clauses with unconstrained associated types` you mean something like this?
```rust
trait Foo<'a> {
  type Bar;
  ...
}

fn f<T>(...) where for<'a> T : Foo<'a>
```

Nadrieril:
> I have one question: by "<'a> clauses with unconstrained associated types" you mean something like this?

Yep exactly, because we'd have to translate that to something like
```rust
fn f<T, Bar<'_>>(...) where for<'a> T : Foo<'a, Bar<'a>>
```
which doesn't exist in rust.

PR #537, opened 2025-01-24
Fix `merge_goto_chains`

Nadrieril:
The part that walks parents to find the first antecedent from which to merge was incorrect: we should of course only consider those antecedents that are plain gotos.


Nadrieril:
I should have added a test for this, I'll do that in the next PR.

Issue #561, opened 2025-02-14
Bug: inconsistent generics when using `Iterator::flat_map`


Nadrieril:
```rust
fn main() {
    let _ = std::iter::empty::<()>().flat_map(|_| vec![()]);
}
```
causes:
```text
error: Found inconsistent generics after transformations:
       Found incorrect clause var: 0_2
       Visitor stack:
         charon_lib::ast::types::TraitRefKind
         charon_lib::ast::types::TraitRef
         charon_lib::ids::vector::Vector<charon_lib::ast::types::vars::TraitClauseId, charon_lib::ast::types::TraitRef>
         charon_lib::ast::types::GenericArgs
         charon_lib::ast::types::TyKind
         charon_lib::ast::types::Ty
         charon_lib::ast::types::FunSig
         charon_lib::ast::gast::FunDecl
       Binding stack (depth 1):
         0: <T, A, U, F, [@TraitClause0]: @TraitDecl0<missing(@Type1_0)>, [@TraitClause1]: @TraitDecl0<missing(@Type1_1)>, [@TraitClause2]: @TraitDecl0<missing(@Type1_2)>, [@TraitClause3]: @TraitDecl0<missing(@Type1_3)>, [@TraitClause4]: @TraitDecl0<@Adt7<missing(@Type1_0), missing(@Type1_1)>[missing(@TraitClause1_0), missing(@TraitClause1_1)]>, [@TraitClause5]: @TraitDecl2<missing(@Type1_2)>, [@TraitClause6]: @TraitDecl4<missing(@Type1_3), (missing(@Type1_0))>, missing(@TraitClause1_6)::parent_clause0::Output = missing(@Type1_2)>
 --> /rustc/library/core/src/iter/traits/iterator.rs:1454:5
```

Nadrieril:
Minimized further:
```rust
//@ charon-args=--hide-marker-traits
use std::marker::PhantomData;

struct FlatMap<U: Clone>(U);

trait Iterator {
    fn flat_map<U: Clone>() -> FlatMap<U> {
        todo!()
    }
}

struct Empty<T: ?Sized>(PhantomData<T>);
impl<T> Iterator for Empty<T> {}

fn main() {}
```
It has to do with the defaulted method.

PR #562, opened 2025-02-14
Clarify uses of `Vector::len`

Nadrieril:
The fact that the vector need not be contiguous is a footgun, we should fix that. Fixes https://github.com/AeneasVerif/charon/issues/561.


Nadrieril:
I tried to make `Vector` contiguous but that's blocked on other changes, I'll get to that eventually.

PR #568, opened 2025-02-21
Monomorphize code via `--monomorphize`

N1ark:
##### Not anywhere near done, but I'll probably work on this in parallel as I need to cover more code -- for now I'm doing the bare minimum.

Add monomorphisation, by applying an algorithm similar [to what rustc does](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_monomorphize/collector/index.html#general-algorithm):
1. Find roots (public monomorphic functions)
2. Generate monomorphic items:
    1. Find uses of other items within the item (e.g. the function's body, the struct's fields...)
    2. For all used items that are polymorphic, make a monomorphic copy with the generic arguments substituted, if one doesn't exist already.
    3. Substitute these uses with the monomorphic version, which is added to the discovered items.
3. Remove all polymorphic items; if they were used anywhere, they have already been monomorphized and substituted.

Note a consequence of this algorithm is that it does dead code removal of polymorphic items; any unused polymorphic item will have 0 monomorphic items created, and will thus be removed.

Added a `--monomorphize` flag to do this.

### DIFF
--- a/charon/src/options.rs
+++ b/charon/src/options.rs
@@ -88,7 +88,7 @@ pub struct CliOpts {
               b0: switch x [true -> goto b1; false -> goto b2]
               b1: y := 0; goto b3
               b2: y := 1; goto b3
-              b3: return y      
+              b3: return y
               ```
 
             We want to reconstruct the control-flow as:
@@ -120,6 +120,9 @@ pub struct CliOpts {
     "))]
     #[serde(default)]
     pub no_code_duplication: bool,
+    #[clap(long = "monomorphize")]
+    #[serde(default)]
+    pub monomorphize: bool,
     /// Usually we skip the bodies of foreign methods and structs with private fields. When this
     /// flag is on, we don't.
     #[clap(long = "extract-opaque-bodies")]
@@ -285,6 +288,8 @@ pub struct TranslateOptions {
     /// Whether to hide the `Sized`, `Sync`, `Send` and `Unpin` marker traits anywhere they show
     /// up.
     pub hide_marker_traits: bool,
+    /// Monomorphize functions.
+    pub monomorphize: bool,
     /// Do not merge the chains of gotos.
     pub no_merge_goto_chains: bool,
     /// Print the llbc just after control-flow reconstruction.
@@ -365,6 +370,7 @@ impl TranslateOptions {
             mir_level,
             no_code_duplication: options.no_code_duplication,
             hide_marker_traits: options.hide_marker_traits,
+            monomorphize: options.monomorphize,
             no_merge_goto_chains: options.no_merge_goto_chains,
             print_built_llbc: options.print_built_llbc,
             item_opacities,
--- a/charon/src/transform/mod.rs
+++ b/charon/src/transform/mod.rs
@@ -13,6 +13,7 @@ pub mod inline_local_panic_functions;
 pub mod insert_assign_return_unit;
 pub mod lift_associated_item_clauses;
 pub mod merge_goto_chains;
+pub mod monomorphize;
 pub mod ops_to_function_calls;
 pub mod prettify_cfg;
 pub mod reconstruct_asserts;
@@ -127,6 +128,8 @@ pub static ULLBC_PASSES: &[Pass] = &[
     // # Micro-pass: remove the drops of locals whose type is `Never` (`!`). This
     // is in preparation of the next transformation.
     UnstructuredBody(&remove_drop_never::Transform),
+    // Monomorphize the functions and types.
+    UnstructuredBody(&monomorphize::Transform),
 ];
 
 /// Body cleanup passes after control flow reconstruction.
--- a/charon/src/transform/monomorphize.rs
+++ b/charon/src/transform/monomorphize.rs
@@ -1,9 +1,73 @@
 //! # Micro-pass: monomorphize all functions and types; at the end of this pass, all functions and types are monomorphic.
+use std::collections::{HashMap, HashSet};
 
 use crate::transform::TransformCtx;
 use crate::ullbc_ast::*;
 
 use super::ctx::UllbcPass;
+
+struct PassData {
+    // Map of (poly item, generic args) -> mono item
+    // None indicates the item hasn't been monomorphized yet
+    items: HashMap<(AnyTransId, GenericArgs), Option<AnyTransId>>,
+    worklist: Vec<AnyTransId>,
+    visited: HashSet<AnyTransId>,
+}
+
+impl PassData {
+    fn new() -> Self {
+        PassData {
+            items: HashMap::new(),
+            worklist: Vec::new(),
+            visited: HashSet::new(),
+        }
+    }
+}
+
+fn find_uses_in_body(data: &mut PassData, body: &ExprBody) {
+    body.locals.vars.iter().for_each(|var| match var.ty.kind() {
+        TyKind::Adt(TypeId::Adt(id), gargs) => {
+            if gargs.is_empty() {
+                return;
+            }
+
+            let key = (AnyTransId::Type(*id), gargs.clone());
+            data.items.entry(key).or_insert(None);
+        }
+        TyKind::Adt(TypeId::Tuple, _) => {}
+        TyKind::Literal(_) => {}
+        ty => warn!("Unhandled type kind {:?}", ty),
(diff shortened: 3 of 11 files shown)


msprotz:
This is super great and something that I will need at some point for Eurydice, too.

Just a heads-up that the main maintainer of Charon, @Nadrieril, will be back online on Monday to help with the review. Thank you!

charon/src/transform/monomorphize.rs
```diff
@@ -0,0 +1,543 @@
...
+        match kind {
+            AggregateKind::Adt(TypeId::Adt(id), _, _, gargs) => self.found_use_ty(id, gargs),
+            AggregateKind::Closure(fun_id, gargs) => self.found_use_fn(fun_id, gargs),
+            _ => {}
+        }
+
+        Continue(())
+    }
+
+    fn visit_ty_kind(&mut self, kind: &TyKind) -> ControlFlow<Infallible> {
```
Nadrieril, line 119: Note: since you're recursing unconditionally, you might find it more convenient to use `enter_ty_kind` instead of `visit_ty_kind`. This is guaranteed to be called for all `TyKind`s without having to deal with recursive calls yourself.

charon/src/options.rs
```diff
@@ -195,6 +199,10 @@ pub struct CliOpts {
     #[clap(long = "cargo-arg")]
     #[serde(default)]
     pub cargo_args: Vec<String>,
+    /// Do nothing! Just run cargo, don't do any translation.
+    #[clap(long = "only-cargo")]
+    #[serde(default)]
+    pub only_cargo: bool,
```
Nadrieril, line 205: What do you use this for?

charon/src/options.rs
```diff
@@ -119,6 +119,10 @@ pub struct CliOpts {
     "))]
     #[serde(default)]
     pub no_code_duplication: bool,
+    /// Monomorphize the code, replacing generics with their concrete types.
+    #[clap(long = "monomorphize")]
+    #[serde(default)]
+    pub monomorphize: bool,
```
Nadrieril, line 125: This changes the serialized AST, therefore please increment the version in `Charon.toml` and run `make test`, this ensures we get nice errors if an ocaml project tries to read an llbc file with the wrong charon version.

charon/src/transform/mod.rs
```diff
@@ -127,6 +128,8 @@ pub static ULLBC_PASSES: &[Pass] = &[
     // # Micro-pass: remove the drops of locals whose type is `Never` (`!`). This
     // is in preparation of the next transformation.
     UnstructuredBody(&remove_drop_never::Transform),
+    // Monomorphize the functions and types.
+    UnstructuredBody(&monomorphize::Transform),
```
Nadrieril, line 132: Your pass doesn't depend on details of llbc vs ullbc, could you make it work for both so that we can run it after other cleanup passes?

charon/src/transform/monomorphize.rs
```diff
@@ -0,0 +1,506 @@
...
+        }
+    }
+
+    fn exit_place(&mut self, place: &mut Place) {
+        match &mut place.kind {
+            PlaceKind::Projection(inner, ProjectionElem::Field(FieldProjKind::Adt(id, _), _)) => {
+                // Trick, we don't know the generics but the projected place does, so
+                // we substitute it there, then update our current id.
+                let (inner_id, _) = inner.ty.as_adt().unwrap();
+                *id = *inner_id.as_adt().unwrap()
```
Nadrieril, line 243: Hm, maybe I should remove this id and let users inspect the type of the projected place...

charon/src/transform/monomorphize.rs
```diff
@@ -0,0 +1,506 @@
...
+            .retain(|f| data.visited.contains(&AnyTransId::Fun(f.def_id)));
+        ctx.translated
+            .type_decls
+            .retain(|t| data.visited.contains(&AnyTransId::Type(t.def_id)));
+        // ctx.translated.trait_impls.retain(|t| t.generics.is_empty());
+
+        // TODO: Currently we don't update all TraitImpls/TraitDecls with the monomorphized versions
+        //       and removing the polymorphic ones, so this fails.
+        // Finally, ensure we didn't leave any IDs un-replaced
+        // check_missing_indices(&ctx.translated);
```
Nadrieril, line 500: You could run this just on the functions and type decls. Or you could remove all traits/impls since they aren't useful after monomorphization.

Nadrieril (approved):
That all looks good to me, I'm happy to merge once conflicts are resolved :)

charon/tests/ui/monomorphization/adt_proj.out
```diff
@@ -0,0 +1,41 @@
+# Final ULLBC before serialization:
+
+enum core::result::Result = // #1
```
Nadrieril, line 3: Eventually I'd want us to add a `PathElem::Monomorphized(GenericArgs)` at the end of the name so we can distinguish them.

charon/src/options.rs
N1ark: I need this to build a library with the same cargo/rustc version that charon uses, since the code I'm generating the ULLBC for is built with `--no-cargo`. I build the binaries with `charon --only-cargo`, then link the binaries with `charon --no-cargo --rustc-arg=-L<my/lib>`

N1ark: If there's a better way to do this, I'm very happy to remove this flag :)

charon/src/transform/monomorphize.rs
N1ark: Don't have a preference personally -- it's a small annoyance for substituting IDs but really not a big deal

charon/src/transform/monomorphize.rs
```diff
@@ -0,0 +1,506 @@
...
+                            let (AnyTransId::Fun(fun_id), gargs) = mono.hint_or(&key_pair) else {
+                                panic!("Unexpected ID type in hint_or");
+                            };
+                            let fun = ctx.translated.fun_decls.get(*fun_id).unwrap();
+                            let mut fun_sub = fun.clone().substitute(gargs);
+                            // fun_sub.signature.generics = GenericParams::empty();
+
+                            let fun_id_sub = ctx.translated.fun_decls.reserve_slot();
+                            fun_sub.def_id = fun_id_sub;
+                            ctx.translated.fun_decls.set_slot(fun_id_sub, fun_sub);
```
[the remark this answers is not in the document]
N1ark: aha ! yes

charon/src/transform/monomorphize.rs
N1ark: Yes, the final goal is being able to remove all traits/impls, however right now my monomorphisation doesn't go far enough (e.g. doesn't substitute `FunIdOrTraitMethodRef::Trait` with `FunIdOrTraitMethodRef::Fun`), so I can't :/ But eventually yes !

charon/tests/ui/monomorphization/adt_proj.out
N1ark: that's a great idea !! I'll keep this in mind for later but I really like this

charon/src/options.rs
N1ark: Added `charon toolchain-path` instead :)

N1ark:
I think most of the comments are done :) @Nadrieril let me know if there's anything else

charon/src/bin/charon/main.rs
```diff
@@ -187,6 +188,12 @@ fn ensure_rustup() {
     }
 }
 
+fn print_toolchain_path() -> ExitStatus {
+    let tc = toolchain::get_toolchain_channel();
+    println!("{}", tc);
+    ExitStatus::default()
+}
```
Nadrieril, line 195: That's not a path, that's the toolchain name. My intention was to return the path to the toolchain sysroot, where the `rustc` and `cargo` binaries are located. I'll do that, give me a moment.

Nadrieril: https://github.com/AeneasVerif/charon/pull/647

N1ark: Ah nice! Thank you :)
Either work for me so this is perfect, I was just doing `cargo +$(charon toolchain-path)`

PR #571, opened 2025-02-28
Make `NameMatcher` work for ULLBC

N1ark:
Currently, `NameMatcher` only works for `LlbcAst`, despite not doing anything specific to it -- change it to allow its use with `UllbcAst` too!


N1ark:
Hmm a signature changed so the aeneas/eurydice tests don't pass, what's usually the way of fixing this? Should I open a PR for these two too?

sonmarcho:
> Hmm a signature changed so the aeneas/eurydice tests don't pass, what's usually the way of fixing this? Should I open a PR for these two too?

This is the usual practice, yes, but here the fact that a signature changed is suspicious. Let me investigate.

Issue #573, opened 2025-03-03
Bug: `Null` value for the serialized field `crate.ordered_decls`


sonmarcho:
Call `charon --hide-marker-traits --remove-associated-types="*"` from the folder `crates/capability-engine` of [Tyche](https://github.com/epfl-dcsl/tyche-devel/tree/main/crates/capability-engine) (commit: 5f725c14).
The field `ordered_decls` of the serialized crate has the value `Null`, which causes an issue when de-serializing the crate using the `charon-ml` library. For instance, `aeneas` will fail with the following error:
```
           ...
          ("ordered_decls", `Null)])

[list_of_json] failed on: `Null

not a list: `Null
```

**Charon version**: bb09c0eb

PR #574, opened 2025-03-03
Continue running passes after an error

Nadrieril:
Each pass normally leaves the `TranslatedCrate` in a good state, so there's no reason not to keep going on subsequent passes. The exception is the `check_generics` pass, but it's a stopgap anyway.

Should fix https://github.com/AeneasVerif/charon/issues/573.


Nadrieril:
I should have added a test but I couldn't reproduce locally. If this isn't enough I'll put more effort into finding a reproducer.

Issue #592, opened 2025-03-11
charon-driver: rustc assertion failure


zjp-CN:
```rustc
tests\cargo\workspace $ charon --print-llbc --cargo-arg="-p=crate1"
error: process didn't exit successfully: `C:\Users\Administrator\.cargo\bin\charon-driver.exe C:\Users\Administrator\.rustup\toolchains\nightl
y-2024-10-23-x86_64-pc-windows-msvc\bin\rustc.exe -vV` (exit code: 101)
--- stderr
thread 'main' panicked at src/bin/charon-driver/main.rs:74:5:
assertion failed: origin_args[1].ends_with("rustc")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We should assert rustc argument by extracting file stem from the rustc path given by cargo.

PR #593, opened 2025-03-11
fix(charon-driver): assert rustc by extracting file stem from the path

zjp-CN:
close https://github.com/AeneasVerif/charon/issues/592


Nadrieril:
Thanks!

PR #594, opened 2025-03-11
Suggest `opam install . --deps-only`

Nadrieril:
This command will install the dependencies listed in the `*.opam` files, which are kept in sync with the `dune-project` file. The dune-project file still has to be hand-maintained, but this is still an improvement over maintaining a separate list in the README.


sonmarcho (approved):
Big improvement, indeed: thanks! :)

Issue #582, opened 2025-03-07
Rework the CLI to use subcommands


Nadrieril:
I'd like the main cli to be `cargo charon`, and to have subcommands for other operations like `cargo pretty-print`, `cargo rustc`.

zjp-CN:
Related issues:
* https://github.com/AeneasVerif/charon/issues/403
* https://github.com/AeneasVerif/charon/issues/587
* https://github.com/AeneasVerif/charon/issues/589

zjp-CN:
I think this issue can be closed now? All related work is done with a final cleanup https://github.com/AeneasVerif/charon/pull/619 .

PR #600, opened 2025-03-12
feat: add `charon cargo [charon options] -- [cargo check options]` subcommand

zjp-CN:
ref https://github.com/AeneasVerif/charon/issues/582

```bash
$ charon --help
Usage: charon [OPTIONS] [COMMAND]

Commands:
  pretty-print  Read a llbc or ullbc file and pretty print it
  cargo         Usage: `charon cargo [charon options] -- [cargo build options]`
  help          Print this message or the help of the given subcommand(s)

Options:
      --ullbc
          Extract the unstructured LLBC (i.e., don't reconstruct the control-flow)
    ...
```

```bash
$ charon cargo --help
Usage: `charon cargo [charon options] -- [cargo build options]`

Usage: charon cargo [OPTIONS] [-- <CARGO>...]

Arguments:
  [CARGO]...  Args that `cargo build` accepts

Options:
      --ullbc
          Extract the unstructured LLBC (i.e., don't reconstruct the control-flow)
    ...
```

```bash
$ charon cargo -- --help
Compile a local package and all of its dependencies

Usage: cargo.exe build [OPTIONS]

Options:
      --future-incompat-report   Outputs a future incompatibility report at the end of the build
      --message-format <FMT>     Error format
  -v, --verbose...               Use verbose output (-vv very verbose/build.rs output)
  -q, --quiet                    Do not print cargo log messages
      --color <WHEN>             Coloring: auto, always, never
      --config <KEY=VALUE|PATH>  Override a configuration value
  -Z <FLAG>                      Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details
  -h, --help                     Print help

Package Selection:
  -p, --package [<SPEC>]  Package to build (see `cargo help pkgid`)
  ...
```

### DIFF
--- a/charon/src/bin/charon-driver/main.rs
+++ b/charon/src/bin/charon-driver/main.rs
@@ -32,11 +32,9 @@ use crate::driver::{
     arg_values, get_args_crate_index, get_args_source_index, CharonCallbacks, CharonFailure,
     RunCompilerNormallyCallbacks,
 };
-use charon_lib::logger;
-use charon_lib::options;
-use charon_lib::trace;
+use charon_lib::{logger, options};
 use itertools::Itertools;
-use std::panic;
+use std::{env, panic};
 
 fn main() {
     // Initialize the logger
@@ -44,7 +42,7 @@ fn main() {
 
     // Retrieve the executable path - this is not considered an argument,
     // and won't be parsed by CliOpts
-    let origin_args: Vec<String> = std::env::args().collect();
+    let origin_args: Vec<String> = env::args().collect();
     assert!(
         !origin_args.is_empty(),
         "Impossible: zero arguments on the command-line!"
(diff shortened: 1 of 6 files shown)


zjp-CN:
Can we set `RUST_BACKTRACE=1` to see where's the error source?

https://github.com/AeneasVerif/charon/actions/runs/13830114808/job/38692306088

I don't know which line causes the error. tests/cli.rs passes on my machine.

Nadrieril:
> I think adding RUST_BACKTRACE on this [line](https://github.com/AeneasVerif/charon/blob/93af11c7f51aa553f8cab4519bf47c6491d2957e/charon/Makefile#L14) should be enough?

That won't work, we don't run `make test` in CI. You can add it in the `checkPhaseCargoCommand` in nix/charon.nix.

zjp-CN:
It's odd.

With `RUST_BACKTRACE=1 cargo test --test cli`, if the test fails on my machine, stdout will print backtraces pointing out the location, instead of mere fn names:



The error in the img are intended not the same as [CI](https://github.com/AeneasVerif/charon/actions/runs/13832102670/job/38698532438#step:4:93), since I'm unable to reproduce the `No such file or directory (os error 2)` CI error.

Why does backtaces differ?

Nadrieril:
Looks great!

PR #603, opened 2025-03-12
Replace calls to the discriminant_value intrinsic with appropriate MIR

Nadrieril:
[This intrinsic](https://doc.rust-lang.org/beta/core/intrinsics/fn.discriminant_value.html) is used in `derive(PartialEq)` for enums, hence shows up a lot. It's also tricky to handle because of its reliance on the built-in associated type `DiscriminantKind::Discriminant`. Replacing it here is not a complete solution but makes for a smoother experience for users of Charon.


msprotz:
🙇

PR #606, opened 2025-03-13
`charon cargo` handles --target, --lib, and --bin

zjp-CN:
* don't set --target if `-- --target` is specified by user
* don't set --lib or --bin: specify `-- --lib` and `-- --bin` instead
* move Toolchain and fns to toolchain module
* update cli.rs:
  * adjust tests/cargo/dependencies to support default feature and `-F test_feature`, thus dependencies.out is updated
  * add tests/cargo/non-host-target to support host target and no-std target, but the test is [skipped in nix](https://aeneas-verif.zulipchat.com/#narrow/channel/423740-dev/topic/charon.20CLI.20rework/near/505381429)

`detect_rustup_and_get_host_target` and `run_charon_cargo` are based on original logics, but slightly changed.


charon/tests/cargo/dependencies.out
```diff
@@ -1,6 +1,29 @@
 # Final LLBC before serialization:
```
Nadrieril, line 1: Can you update the test files in the same commit that caused the change please? This makes it easier to follow what's changing.

charon/tests/cli.rs
```diff
@@ -69,3 +69,45 @@ fn charon_cargo_p_crate2() -> Result<()> {
         },
     )
 }
+
+#[test]
+fn charon_cargo_features() -> Result<()> {
```
Nadrieril, line 74: This is redundant with the existing `cargo/dependencies` test. I'd prefer if you just updated the command-line instruction we use for that test in `cargo.rs`.

I prefer tests that commit the whole output because imo tests like this one that test for the presence/absence of a string are more annoying to fix when they break.

charon/tests/cargo/dependencies.out
zjp-CN: Ok. I may kinda mess the history like this when I do the rebase.

charon/tests/cli.rs
Nadrieril: You know what, I think your way is more robust actuallyl, nevermind.

charon/tests/cargo/non-host-target/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#![no_std]
+
+pub fn add(left: u64, right: u64) -> u64 {
```
Nadrieril, line 3: Could you add some target-dependent code here to we're sure that we're actually using the correct target?

charon/src/bin/charon/cli_rework/cargo.rs
```diff
@@ -10,3 +13,49 @@ pub struct CargoArgs {
     #[arg(last = true)]
     pub cargo: Vec<String>,
 }
+
+// Differences with current cargo call:
+// * cargo build rather than cargo rustc
+// * no longer set --lib and --bin: need to prune lib and bin fields in toml_config
+pub fn run_charon_cargo(mut options: CliOpts, host: &str) -> Result<ExitStatus> {
```
Nadrieril, line 20: I'd much rather you don't duplicate the logic. Switching to `cargo build` should work, and I don't see the harm in passing `--lib` ourselves if the user requests it.

Nadrieril: I'm making the `cargo build` change separately: https://github.com/AeneasVerif/charon/pull/607

charon/tests/cargo/non-host-target/src/lib.rs
zjp-CN: Actually, I would like this package to be able to compile on both host and non-host target. So it shouldn't be target-dependent.

The only requirement I think is cross-compilation is feasible on host machine.

zjp-CN: I'll add a comment for this package tho.

Or if you want this package to be only compile on single non-host target?

Nadrieril: I meant something like a `cfg(target = ...)` so we can observe a difference in the compiled code

charon/tests/cli.rs
```diff
@@ -111,3 +111,76 @@ fn charon_cargo_features() -> Result<()> {
...
+#[test]
+#[ignore = "Network is inaccessible in nix"]
+fn charon_cargo_target() -> Result<()> {
+    let target = "riscv64gc-unknown-none-elf";
+    rustup_install_target(target)?;
+
+    let dir = "tests/cargo/multi-targets";
+    let fn_ = "pub fn multi_targets::";
+
+    #[cfg(target_family = "unix")]
```
Nadrieril, line 136: This looks incorrect: we're in a cross-compilation context; the test itself should be compiled and run on the host, only charon itself should be invoqued with the required target. The commands should be `charon cargo -- --target=blah`, no?

Nadrieril: Ah no I see, you're also testing the case where the host is different.

charon/tests/cargo/multi-targets/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[cfg(target_os = "windows")]
+pub fn on_windows(left: u64, right: u64) -> u64 {
+    left + right
+}
+
+#[cfg(target_family = "unix")]
+pub fn on_unix(left: u64, right: u64) -> u64 {
+    left + right
+}
```
Nadrieril, line 19: Could you add one more of these that for `riscv64gc-unknown-none-elf`?

PR #611, opened 2025-03-17
feat: deploy index.html and rustc internal APIs to Github Pages

### LINKED ISSUE #610: Add index.html and rustc internal API docs for Github Pages
Currently [aeneasverif.github.io/charon](https://aeneasverif.github.io/charon) is 404.

I can add an index.html file in deployment path to point to these pages:
* charon's API docs, i.e.  [aeneasverif.github.io/charon/charon_lib](https://aeneasverif.github.io/charon/charon_lib) introduceed in  https://github.com/AeneasVerif/charon/issues/445
* rustc internal API docs required by charon's toolchain

It'll look [like](https://os-checker.github.io/Rudra) this for modified Rudra, since I want to reference these APIs online when learning about the codebase.

zjp-CN:
close #610 

Preview: https://os-checker.github.io/charon

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -23,13 +23,21 @@ jobs:
           rustup toolchain list -v
           # Find toolchain path
           export TOOLCHAIN=$(rustup toolchain list -v | grep active | awk '{print $NF}')
-          echo "toolchain=$TOOLCHAIN"
+          echo "toolchain=\"$TOOLCHAIN\""
           # Remove rustc book
           rm $TOOLCHAIN/share/doc/rust/html/rustc -rf
           # Download rustc API docs
           rustup component add rustc-docs
           # Move the docs to deployment path
           mv $TOOLCHAIN/share/doc/rust/html/rustc charon/target/doc/rustc
+
+      - name: Prepare index.html
+        run: |
+          # Find toolchain name
+          export toolchain=$(rustup toolchain list | grep active | grep -oP "nightly-\d{4}-\d{2}-\d{2}")
+          echo "toolchain=\"$toolchain\""
+          # Fill in toolchain placeholder in index.html
+          sed "s/nightly-xxxx-xx-xx/$toolchain/" .github/index.html > charon/target/doc/index.html
 
       - name: Upload static files as artifact
         uses: actions/upload-pages-artifact@v3 # or specific "vX.X.X" version tag for this action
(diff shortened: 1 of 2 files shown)


.github/workflows/doc.yml
```diff
@@ -14,13 +14,24 @@ jobs:
...
-        id: build
         run: |
           cd charon
           make doc
 
+      - name: Install rustc internal docs
+        run: |
+          rustup toolchain list -v
+          # Find toolchain path
+          export TOOLCHAIN=$(rustup toolchain list -v | grep active | awk '{print $NF}')
```
Nadrieril, line 25: That command doesn't work on my machine. I think `$(rustc --print sysroot)/share/doc/...` would more more robustly.

zjp-CN: It also doesn't work on my machine, because the pattern string for grep is `override` if multiple toolchains are installed...

I changed that due to github runner behavior.



> I think `$(rustc --print sysroot)/share/doc/...` would more more robustly.

Good idea. I'll update it.

.github/workflows/doc.yml
```diff
@@ -23,14 +23,22 @@ jobs:
...
           rm $TOOLCHAIN/share/doc/rust/html/rustc -rf
           # Download rustc API docs
           rustup component add rustc-docs
           # Move the docs to deployment path
           mv $TOOLCHAIN/share/doc/rust/html/rustc charon/target/doc/rustc
 
+      - name: Prepare index.html
+        run: |
+          # Find toolchain name
+          export toolchain=$(rustup toolchain list | grep active | grep -oP "nightly-\d{4}-\d{2}-\d{2}")
```
Nadrieril, line 37: Same here, the toolchain name can be found in the `./rust-toolchain` toml file. Try:
```bash
cargo install toml2json
toolchain="$(toml2json rust-toolchain | jq -r .toolchain.channel)"
```

zjp-CN: Would you mind using https://github.com/taiki-e/install-action?tab=readme-ov-file#supported-tools to install toml2json? The action saves installing it from compiling from source code, it may download toml2json from github via cargo-binstall instead.

zjp-CN: Or rather `rustc --print sysroot | grep -oP "nightly-\d{4}-\d{2}-\d{2}"` just works.

zjp-CN:
Is `Build and test / aeneas` workflow cancelled manually?

Nadrieril:
Oh well spotted, I didn't think that `rerun` would cancel things, I'll have to fix that.

zjp-CN:
Yay, https://aeneasverif.github.io/charon looks right!

Issue #587, opened 2025-03-10
Fix the rustc argument handling


Nadrieril:
Today in `charon-driver/main.rs` we do some rather ugly string-based argument manipulation. I think a lot of that could be replaced with direct manipulation of rustc's `&mut Config` by hooking ourselves in the [`config`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_driver/trait.Callbacks.html#method.config) callback.

This would make the code more robust and likely fix some bugs.

Nadrieril:
@zjp-CN this is a different kind of CLI cleanup, thought you could possibly be interested.

Nadrieril:
Here is one bug likely caused by our current ad-hoc argument handling: https://github.com/AeneasVerif/charon/issues/403

PR #615, opened 2025-03-19
Rework rustc-args with Config

zjp-CN:
close https://github.com/AeneasVerif/charon/issues/587

This PR replaces  string-based argument manipulation of rustc args by functions from `config` submodule where a set of  callbacks takes `&mut Config` to be applied to `<CharonCallbacks as Callbacks>::config`.

### DIFF
--- a/charon/src/bin/charon-driver/driver.rs
+++ b/charon/src/bin/charon-driver/driver.rs
@@ -6,7 +6,10 @@ use charon_lib::transform::{
 use charon_lib::transform::{TransformCtx, SHARED_FINALIZING_PASSES};
 use charon_lib::{export, options};
 use rustc_driver::{Callbacks, Compilation};
-use rustc_interface::{interface::Compiler, Queries};
+use rustc_interface::{
+    interface::{Compiler, Config},
+    Queries,
+};
 use std::fmt;
 use std::ops::Deref;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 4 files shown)


charon/src/bin/charon-driver/driver.rs
```diff
@@ -131,6 +134,9 @@ impl Callbacks for CharonCallbacks {
                 }
             }
         });
+
+        // Mutate other fields in Config
+        MutateConfig::new().do_operation(config);
```
Nadrieril, line 139: This is more complicated than needed. Why not just call `disabled_mir_passes(config);` etc here?

zjp-CN: Done. I've also cleaned up the history commits with MutateConfig pruned.

charon/src/bin/charon-driver/driver/config_mutation.rs
```diff
@@ -53,3 +54,18 @@ pub fn disabled_mir_passes(config: &mut Config) {
...
+/// Don't even try to codegen. This avoids errors due to checking if the output filename is
+/// available (despite the fact that we won't emit it because we stop compilation early).
+pub fn no_codegen(config: &mut Config) {
+    config.opts.unstable_opts.no_codegen = true;
+
+    // Original:
+    // compiler_args.push("--emit=metadata".to_string());
+    // FIXME: [`OutputTypes`] doesn't expose mutation methods, so here construct a new
+    // value to replace the whole output_types. This discards all outputs analyzed from
+    // cli args, so not sure if this breaks something.
```
Nadrieril, line 67: You can do something like `OutputTypes::new(config.opts.output_types.iter().cloned().chain([(OutputType::Metadata, None)]).collect_vec().as_slice())`.

zjp-CN: Done.

Small adjustment: `config.opts.output_types.iter()` yields `(&K, &V)`, so it needs to clone each element to get owned values.

PR #624, opened 2025-03-24
Translate modules like other items

Nadrieril:
Until now crate translation worked in two phases: recurse through the modules to enqueue all the items we find, then translate enqueued items until exhaustion. This PR makes it so we also manage modules through the item queue. This reduces duplication in item_meta/opacity handling.


Nadrieril:
The output diff is too big.

PR #637, opened 2025-04-04
Update the rustc pin

Nadrieril:
This updates the rustc pin to a nightly from a few days ago. This is the companion PR to https://github.com/cryspen/hax/pull/1391.


charon/src/ast/expressions.rs
```diff
@@ -158,6 +158,8 @@ pub enum UnOp {
     /// (in debug mode) to check that it is not equal to the minimum integer
     /// value (for the proper type).
     Neg,
+    /// Retreive the metadata part of a fat pointer. For slices, this retreives their length.
```
N1ark, line 161:
```suggestion
    /// Retrieve the metadata part of a fat pointer. For slices, this retrieves their length.
```

Nadrieril: :D oops

PR #638, opened 2025-04-05
Add disambiguators in the name patterns

sonmarcho:
This PR introduces disambiguators in the name patterns.

Disambiguators are actually necessary for *identifier* elements (but not *impl* elements).

For instance, in the snippet of code below, the two const values will be defined as top-level definitions and will have the same name modulo the disambiguator:
```rust
fn f(...) {
  match ... {
    Variant1 => {
      const N_ROWS:usize = 4; // disambiguator 0
      ...
    }
    Variant2 => {
      const N_ROWS:usize = 4; // disambiguator 1
      ...
    }
  }
}
```

In order not to use disambiguators everywhere in patterns, we allow omitting the disambiguator when it is equal to 0.
For instance, pattern `f` will be parsed as `PIdent ("f", 0, [])` (disambiguator 0, no generics), while pattern `f#1` will be parsed as `PIdent ("f", 1, [])`.

Finally,  we still omit the disambiguators appearing in *impl* elements.

### DIFF
--- a/charon-ml/name_matcher_parser/Ast.ml
+++ b/charon-ml/name_matcher_parser/Ast.ml
@@ -28,7 +28,32 @@ and region = RVar of var option | RStatic
 and primitive_adt = TTuple | TArray | TSlice
 and mutability = Mut | Not
 and pattern = pattern_elem list
-and pattern_elem = PIdent of string * generic_args | PImpl of expr
+
+and pattern_elem =
+  | PIdent of string * int * generic_args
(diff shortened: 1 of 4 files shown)


Nadrieril:
Could you add tests for this please? We have testing infrastructure for the ml name matcher here: https://github.com/AeneasVerif/charon/blob/main/charon/tests/ui/ml-name-matcher-tests.rs .

PR #649, opened 2025-04-10
Represent global accesses via places

Nadrieril:
Globals (statics and consts) in MIR can be accessed either by copying the value or taking a reference/pointer to them. They very much behave like places, but until now we had a special-case rvalue for each of these possible operations. This PR represents globals as a `Place` instead, just like locals.


Nadrieril:
Superceded by https://github.com/AeneasVerif/charon/pull/783

PR #653, opened 2025-04-14
Add `Binop::{Cmp,Offset}`, `BuiltinFunId::PtrFromParts`

N1ark:
Add some additional MIR elements to the AST. `BuiltinFunId::PtrFromParts` is desugared from `AggregateKind::RawPtr`, though ideally it will be replaced with `core::ptr::from_raw_parts[_mut]` (or maybe `std::intrinsics::aggregate_raw_ptr`).

@Nadrieril is there a reason why you didn't want RawPtr as an aggregate kind?

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.76"
+let supported_charon_version = "0.1.77"
--- a/charon-ml/src/ExpressionsUtils.ml
+++ b/charon-ml/src/ExpressionsUtils.ml
@@ -18,5 +18,6 @@ let binop_can_fail (binop : binop) : bool =
   | Gt
   | CheckedAdd
   | CheckedSub
-  | CheckedMul -> false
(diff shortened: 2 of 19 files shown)


Nadrieril:
At the same time I'm not sure I'd want to replace every binary operation with a function call so there's a choice to be made somewhere.

N1ark:
> > @Nadrieril is there a reason why you didn't want RawPtr as an aggregate kind?
> 
> This is honestly a design question I'm not sure of. My full intent is that we replace all `BuiltinFunId`s we can with normal function items. The idea is that having a function item gives you its type signature and name (even possible contract!), which requires a lot less special-casing in users of Charon. E.g. right now both Aeneas and Eurydice hardcode the signature that the various indexing builtins should have.

Hmm fair enough! I think for our use case having all aggregates etc directly is more practical than builtin function calls but we can make do with either

N1ark:
Working on the tests -- fyi, I can't test `BuiltinFunId::PtrFromParts`, because it is tied to the definition of `Thin`, defined as `pub trait Thin = Pointee<Metadata = ()>;`, ie. a trait alias, which Charon doesn't support -- I'll leave the example here, for when we'll be able to add the test (except if you have a better idea).
```rust
#![feature(core_intrinsics)]
fn main() {
    let x: i32 = 11;
    let ptr: *const i32 = &x as *const i32;
    let slice: *const [i32] = unsafe { std::intrinsics::aggregate_raw_ptr(ptr, 1) };
}
```

PR #661, opened 2025-04-21
Add missing `StorageLive`s

N1ark:
Now that Charon supports [`StorageLive`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/enum.StatementKind.html#variant.StorageLive), we must ensure they are sensibly placed; notably, there are two places where locals are used before being allocated:
1. For the temporary locals used for the builtin function for array/slice indexing -- these are introduced in the `index_to_function_calls` micro-pass
2. For locals used to store the discriminant of an enum -- [this has already been reported](https://internals.rust-lang.org/t/semantics-of-storagelive-storagedead-in-mir/16184/7), and is a bit odd, the issue comes from rustc.

This behaviour is "ok" with the semantics (the locals, when no StorageLive is associated to them, are considered always allocated), but makes consuming the (U)LLBC code more tiresome, as one must check for these locals dynamically. This PR ensures these are fixed, by:
1. Adding the proper `StorageLive` statements in `index_to_function_calls` -- we do not add the `StorageDead`s, because that would make the micropass unecessarily complicated, and while for the indexing locals we can easily tell when the local becomes dead, it is non-trivial to find exactly when the result of the builtin function call becomes unused. So instead we just don't emit these, which is ok. Though it's not officially documented anywhere afaik, as per a comment from Ralf Jung:
    > If we don't emit StorageDead, that basically means there is an implicit StorageDead when the function returns. That seems fine for me.
2. Added a micro-pass `insert_storage_lives` to add missing `StorageLive`s at the start of functions; again, we could be more clever about it and attempt to detect the exact scope of the local, but I'm afraid this is risky and may mess with the semantics. 
  Rust in particular already considers locals that have no `StorageLive` associated as "always allocated", so adding a `StorageLive` at the start of the function mimics these semantics and doesn't play any risky games.

Note these changes do not include the return value (index `0`) and the function's arguments (index `1..=arg_count`), since these are known to always be allocated by the caller.

ci: use https://github.com/AeneasVerif/aeneas/pull/505


charon/src/transform/index_to_function_calls.rs
```diff
@@ -82,10 +82,15 @@ impl<'a> IndexVisitor<'a> {
...
-        // Push the statement:
-        //`tmp0 = &{mut}p`
+        // Push the statements:
+        // `storage_live tmp0`
+        // `tmp0 = &{mut}p`
         let input_var = {
             let input_var = self.fresh_var(None, input_ty);
+
+            let live_kind = RawStatement::StorageLive(input_var.local_id());
+            self.statements.push(Statement::new(self.span, live_kind));
```
Nadrieril, line 92: Could you add this to `self.fresh_var` so we can't miss one?

charon/src/transform/insert_storage_lives.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::transform::TransformCtx;
+use crate::ullbc_ast::{BlockId, RawStatement, Statement};
+
+use super::ctx::UllbcPass;
+
+pub struct Transform;
+impl UllbcPass for Transform {
+    fn transform_ctx(&self, ctx: &mut TransformCtx) {
+        ctx.for_each_fun_decl(|_ctx, fun| {
+            let Ok(Body::Unstructured(body)) = &mut fun.body else {
```
Nadrieril, line 19: You can implement `transform_body` here instead, which will do the `for_each_fun_decl` for you.

Nadrieril (changes requested):
I like this, this is very much the kind of thing we want CHaron to fix up.

Nadrieril:
Could you also add a comment to the `StorageLive` statement explaining when they exist, with a mention of the `insert_storage_lives` pass? So ppl know they can rely on them to be there.

charon/src/transform/index_to_function_calls.rs
N1ark: ah yes that's much smarter ;P

charon/src/transform/insert_storage_lives.rs
N1ark: This made me realise that this pass should definitely be `NonBody` and not `UnstructuredBody` :') so I changed that and made the pass handle both

N1ark:
All fixed up 😸

charon/src/transform/insert_storage_lives.rs
```diff
@@ -3,68 +3,100 @@
...
+            }
+            _ => {}
+        }
+        Continue(())
+    }
+}
 
 pub struct Transform;
-impl UllbcPass for Transform {
+impl TransformPass for Transform {
```
Nadrieril: Honestly that generality is not needed. Let's do it only on ullbc, that's how we do it for all other passes that modify statements. The passes for which it makes sense to work on both are "finalizing" passes that run at the end.

charon/src/transform/index_to_function_calls.rs
```diff
@@ -132,12 +128,9 @@ impl<'a> IndexVisitor<'a> {
             );
             self.statements.push(Statement::new(self.span, kind));
 
-            // `storage_live(index_var)`
```
Nadrieril: Let's keep these comments, they are accurate even if the `StorageLive` is added inside `fresh_var`

charon/src/transform/insert_storage_lives.rs
Nadrieril: Oh, you added it as a finalizing pass actually. All good then.

N1ark: yes, felt safer to do it at the end in case additional locals are introduced without storage lives

charon/src/transform/insert_storage_lives.rs
```diff
@@ -72,12 +73,16 @@ impl TransformPass for Transform {
...
-                        storage_visitor.unmentioned_locals.into_iter().map(|local| {
-                            llbc_ast::Statement::new(
-                                first_span,
-                                llbc_ast::RawStatement::StorageLive(local),
-                            )
-                        });
+                        storage_visitor
+                            .unmentioned_locals
+                            .iter()
+                            .sorted()
```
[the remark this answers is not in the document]
N1ark: Done with `BTreeMap<LocalId, ()>`

PR #662, opened 2025-04-22
Graceful `Opaque` consts

N1ark:
Add `RawConstantExpr::Opaque`, to represent constants that hax/Charon still don't translate properly. Currently, translating some code (e.g. a lot of the std lib) uses unsupported consts at one point or another, even if down the line the const is not needed.

This PR allows Charon to keep translating gracefully, avoiding unnecessary failures! The consumer is then free to decide what to do with the opaque const (give up, ignore, stub, etc.)

ci: use https://github.com/AeneasVerif/aeneas/pull/507

### DIFF
--- a/charon/src/ast/expressions.rs
+++ b/charon/src/ast/expressions.rs
@@ -485,6 +485,7 @@ pub enum RawConstantExpr {
     #[drive(skip)]
     RawMemory(Vec<u8>),
     /// A constant expression that Charon still doesn't handle, along with the reason why.
+    #[drive(skip)]
     Opaque(String),
 }
 
--- a/charon/src/ast/expressions.rs
+++ b/charon/src/ast/expressions.rs
@@ -484,6 +484,8 @@ pub enum RawConstantExpr {
     /// representation isn't possible (e.g. for unions) or just isn't implemented yet.
     #[drive(skip)]
     RawMemory(Vec<u8>),
+    /// A constant expression that Charon still doesn't handle, along with the reason why.
+    Opaque(String),
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Drive, DriveMut)]
--- a/charon/src/ast/visitor.rs
+++ b/charon/src/ast/visitor.rs
@@ -39,6 +39,8 @@ use indexmap::IndexMap;
     // Defines the `Visit[Mut]` traits and the `drive[_mut]` method that drives them.
     visitor(drive(&VisitAst)),
     visitor(drive_mut(&mut VisitAstMut)),
+    // Types we ignore
+    skip(String),
     // Types that we unconditionally explore.
     drive(
         AbortKind, Assert, BinOp, Body, BorrowKind, BuiltinFunId, BuiltinIndexOp, BuiltinTy, Call,
--- a/charon/src/ast/visitor.rs
+++ b/charon/src/ast/visitor.rs
@@ -39,8 +39,6 @@ use indexmap::IndexMap;
     // Defines the `Visit[Mut]` traits and the `drive[_mut]` method that drives them.
     visitor(drive(&VisitAst)),
     visitor(drive_mut(&mut VisitAstMut)),
-    // Types we ignore
-    skip(String),
     // Types that we unconditionally explore.
     drive(
         AbortKind, Assert, BinOp, Body, BorrowKind, BuiltinFunId, BuiltinIndexOp, BuiltinTy, Call,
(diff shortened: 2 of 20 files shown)


N1ark:
How do I make CI use the right hax version? CI fails because of this, despite what `select-dep-versions` says

charon/src/ast/expressions.rs
```diff
@@ -484,6 +484,8 @@ pub enum RawConstantExpr {
     /// representation isn't possible (e.g. for unions) or just isn't implemented yet.
     #[drive(skip)]
     RawMemory(Vec<u8>),
+    /// A constant expression that Charon still doesn't handle, along with the reason why.
+    Opaque(String),
```
Nadrieril, line 488: I prefer to skip here rather than in the visitor trait definition, because for cases like the `Vec<u8>` above, the visitor-wide `skip` would still iterate through the whole `Vec`.
```suggestion
    #[drive(skip)]
    /// A constant expression that Charon still doesn't handle, along with the reason why.
    Opaque(String),
```

charon/src/ast/visitor.rs
```diff
@@ -39,6 +39,8 @@ use indexmap::IndexMap;
     // Defines the `Visit[Mut]` traits and the `drive[_mut]` method that drives them.
     visitor(drive(&VisitAst)),
     visitor(drive_mut(&mut VisitAstMut)),
+    // Types we ignore
+    skip(String),
```
Nadrieril, line 43: See remark above
```suggestion
```

charon/src/bin/charon-driver/translate/translate_constants.rs
```diff
@@ -144,21 +144,13 @@ impl<'tcx, 'ctx> BodyTransCtx<'tcx, 'ctx> {
                 RawConstantExpr::Ref(Box::new(val))
             }
             ConstantExprKind::Cast { .. } => {
-                raise_error!(
-                    self,
-                    span,
-                    "Unsupported constant: `ConstantExprKind::Cast {{..}}`",
-                )
+                RawConstantExpr::Opaque("`ConstantExprKind::Cast {{..}}`".into())
```
Nadrieril, line 147: I'm ok with returning this, but we should still emit a warning imo. At least current Charon policy is to emit warnings when we can't trnaslate something. FOr that, replace the `raise_error` above with `register_error`.

charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -823,6 +823,7 @@ impl<C: AstFormatter> FmtWithCtx<C> for RawConstantExpr {
                 format!("{}", f.fmt_with_ctx(ctx))
             }
             RawConstantExpr::RawMemory(bytes) => format!("RawMemory({bytes:?})"),
+            RawConstantExpr::Opaque(cause) => format!("Opaque({cause})"),
```
Nadrieril, line 826: We should raise an error here too

Nadrieril: oh oops, I indeed meant to add a comment wherever we create them :D

N1ark:
My hax PR is in the merge queue, as soon as it's merged this is ready to go : )

N1ark:
I added the `register_error`s as recommended, but that made the tests fail again, which is somewhat expected. To still be able to test the behaviour of the new `Opaque` consts, I added a `//@ ignore-warnings` rule to tests, that doesn't add `--error-on-warnings` to the command :) and i copied the two failing tests with warnings disabled to show it works as expected

Nadrieril:
I do want these tests to fail tho

charon/src/bin/charon-driver/translate/translate_constants.rs
```diff
@@ -187,8 +191,8 @@ impl<'tcx, 'ctx> BodyTransCtx<'tcx, 'ctx> {
             }
             ConstantExprKind::Memory(bytes) => RawConstantExpr::RawMemory(bytes.clone()),
             ConstantExprKind::Todo(msg) => {
-                // Case not yet handled by hax
-                raise_error!(self, span, "Unsupported constant: {:?}", msg)
+                register_error!(self, span, "Unsupported constant by Hax: {:?}", msg);
+                RawConstantExpr::Opaque("Hax: ".to_owned() + msg)
```
Nadrieril, line 195: We should not mention hax to end-users if we can avoid it.
```suggestion
                register_error!(self, span, "Unsupported constant: {:?}", msg);
                RawConstantExpr::Opaque(msg)
```

PR #663, opened 2025-04-22
Fix OCaml chars

N1ark:
OCaml `char`s are ASCII characters, spanning one byte, whereas Rust `char`s are UTF-8 characters, spanning four bytes. This means that the parsing of certain characters in the OCaml library (e.g. `std::char::MAX`) would fail at the parsing stage. 

This PR addresses this, by using [`Uchar.t`](https://ocaml.org/manual/5.1/api/Uchar.html) instead of `char` in OCaml, which sports the same guarantees as Rust's char. Unfortunately, a few utility functions we would like having were only added in OCaml 4.14.0 (notably [`String.get_utf8_uchar`](https://ocaml.org/manual/5.3/api/String.html#utf_8)), and Charon is currently on OCaml 4.13.1, so I included a few methods from OCaml's stdlib in our version of it, as a temporary measure.

Another option would be storing the value as a `int` -- this works too, and may be simpler, but comes with the drawback of not validating the value for UTF-8, and doesn't make converting to/from a string much easier.

ci: use https://github.com/AeneasVerif/aeneas/pull/506

### DIFF
--- a/charon/tests/ui/max_char.out
+++ b/charon/tests/ui/max_char.out
@@ -1,0 +1,43 @@
+# Final LLBC before serialization:
+
+pub fn core::char::methods::{char}::MAX() -> char
+{
+    let @0: char; // return
+
+    @0 := const (􏿿)
+    return
+}
+
+pub global core::char::methods::{char}::MAX: char = core::char::methods::{char}::MAX()
+
+pub fn core::char::MAX() -> char
+{
+    let @0: char; // return
+    let @1: char; // anonymous local
+
+    storage_live(@1)
+    @1 := core::char::methods::{char}::MAX
+    @0 := move (@1)
+    return
+}
+
+pub global core::char::MAX: char = core::char::MAX()
+
+fn test_crate::main()
+{
+    let @0: (); // return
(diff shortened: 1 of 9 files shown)


Nadrieril:
I moved the minimum OCaml versino to 4.14: https://github.com/AeneasVerif/charon/pull/664.

Nadrieril (approved):

charon/tests/ui/max_char.out
```diff
@@ -0,0 +1,43 @@
+# Final LLBC before serialization:
+
+pub fn core::char::methods::{char}::MAX() -> char
+{
+    let @0: char; // return
+
+    @0 := const (􏿿)
```
N1ark, line 7: Note here, it doesn't show but between the parenthesis is the unicode character `0x10FFFF` which is indeed char::MAX

Nadrieril: In Firefox I do get a little symbol that shows its bytes :) 
image

N1ark: ohhhh cute! hahahha

PR #669, opened 2025-04-25
Translate promoted constants inline

Nadrieril:
Companion PR to https://github.com/cryspen/hax/pull/1420. Hax now gives a def_id to promoted constants, which makes it super easy for charon to support them. To keep the body looking like the user wrote it, this PR also adds a pass that inlines promoted constant bodies back into their parent body. This will be made optional eventually.

ci: use https://github.com/cryspen/libcrux/pull/936

### DIFF
--- a/charon/tests/ui/simple/multiple-promoteds.out
+++ b/charon/tests/ui/simple/multiple-promoteds.out
@@ -1,0 +1,107 @@
+# Final LLBC before serialization:
+
+fn test_crate::six::{promoted_const}#1() -> &'_ (u32)
+{
+    let @0: &'_ (u32); // return
+    let @1: u32; // anonymous local
+    let @2: (u32, bool); // anonymous local
+
+    storage_live(@1)
+    storage_live(@2)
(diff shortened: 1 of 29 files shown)


charon/tests/ui/simple/multiple-promoteds.out
```diff
@@ -0,0 +1,87 @@
...
+    storage_live(@5)
+    storage_live(@6)
+    storage_live(@7)
+    storage_live(@8)
+    storage_live(@9)
+    storage_live(@10)
+    storage_live(@11)
+    storage_live(@12)
+    storage_live(x@1)
+    @9 := const (0 : u32) checked.+ const (1 : u32)
```
sonmarcho, line 30: I'm surprised that there are no checks for overflow here. Is it normal or did Charon miss something?

Nadrieril: I must have gotten mixed up with commits because I noticed and was intending to fix that. I think what happens is that the overflow check stays in the function but the addition is promoted, which confuses the binop reconstruction pass.

PR #680, opened 2025-05-06
Add a few boxes to reduce ADT sizes

giltho:
- **Implement hash explicitely for TraitClause**
- **systematically Box generic args**
- **box constant exprs in Operands**


Nadrieril:
Ah, you have to run `make generate-ml` also

giltho:
I can't...
Something is broken with `topkg`'s opam package, I cannot install a new switch. And I cannot install ocamlformat.0.26.2 without changing my ocaml version.
So I can't run `dune fmt` 😭

PR #681, opened 2025-05-07
Closures

### LINKED ISSUE #194: Clarify the encoding of closures

N1ark:
Fixes  #194

Left to do: 

- [x] Add fun implementations for cases where the target FnKind isn't that of the closure (e.g. FnOnce->Fn, etc.)
- [x] Unpack the tupled closure arguments 
- [x] Test for closures with generics (probably broken)
- [x] Test for closures enclosed in generic parents (probably broken)
- [x] The above two together (probably broken)

ci: use https://github.com/AeneasVerif/aeneas/pull/527
ci: use https://github.com/AeneasVerif/eurydice/pull/202


N1ark:
@Nadrieril fyi this is ready for review :) once it's in a good shape ill update aeneas/eurydice

PR #682, opened 2025-05-07
Make Call a terminator

giltho:
Some passes are a bit better, some are much worse


Nadrieril:
Justification for this: this is needed to get unwinding information.

PR #685, opened 2025-05-09
Update ML lib with non-opaque index/slice places

N1ark:
PR #682 made `ops_to_function_calls` a LLBC-only pass, meaning when compiling as ULLBC, `Index`/`Slice` places and `ArrayToSlice` unops can appear in the output, but the ML library hasn't been updated to reflect this.

The next step to this PR is #686

ci: use https://github.com/AeneasVerif/aeneas/pull/519

### DIFF
--- a/charon-ml/src/ExpressionsUtils.ml
+++ b/charon-ml/src/ExpressionsUtils.ml
@@ -2,7 +2,7 @@ open Expressions
 
 let unop_can_fail (unop : unop) : bool =
   match unop with
-  | Neg | Cast _ | PtrMetadata -> true
+  | Neg | Cast _ | PtrMetadata | ArrayToSlice _ -> true
   | Not -> false
 
 let binop_can_fail (binop : binop) : bool =
--- a/charon-ml/src/PrintExpressions.ml
+++ b/charon-ml/src/PrintExpressions.ml
@@ -35,10 +35,17 @@ let (var_id_to_string
 let (var_to_string [@ocaml.alert deprecated "use [local_to_string] instead"]) =
   local_to_string
 
-let projection_elem_to_string (env : 'a fmt_env) (sub : string)
+let rec projection_elem_to_string (env : 'a fmt_env) (sub : string)
     (pe : projection_elem) : string =
   match pe with
   | Deref -> "*(" ^ sub ^ ")"
+  | ProjIndex (off, from_end) ->
+      let idx_pre = if from_end then "-" else "" in
+      "(" ^ sub ^ ")[" ^ idx_pre ^ operand_to_string env off ^ "]"
+  | Subslice (from, to_, from_end) ->
+      let idx_pre = if from_end then "-" else "" in
+      "(" ^ sub ^ ")[" ^ idx_pre ^ operand_to_string env from ^ ".."
+      ^ operand_to_string env to_ ^ "]"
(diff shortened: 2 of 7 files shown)


Nadrieril (approved):
Looks good, just needs the aeneas fix

charon-ml/src/ExpressionsUtils.ml
```diff
@@ -2,7 +2,7 @@ open Expressions
 
 let unop_can_fail (unop : unop) : bool =
   match unop with
-  | Neg | Cast _ | PtrMetadata -> true
+  | Neg | Cast _ | PtrMetadata | ArrayToSlice _ -> true
```
sonmarcho, line 5: Why is `ArrayToSlice` considered fallible?
(not important once https://github.com/AeneasVerif/charon/issues/686#issuecomment-2868078701 is completed of course)

N1ark: hmm wasn't too sure -- my intuition was just that maybe the parameter is not properly typed or something? maybe im wrong, though anyways this won't matter -- Aeneas uses LLBC so ArrayToSlice is removed from the AST

sonmarcho: Yes, it won't matter for applications consuming the LLBC and it will be removed anyway (but please note that generally speaking it's important to consider that Aeneas is not the only consumer of the ML library ;) ).

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -317,6 +317,20 @@ impl BodyTransCtx<'_, '_, '_> {
                         )
                     }
                     &hax::ProjectionElem::Subslice { from, to, from_end } => {
+                        // Rust doesn't update the projection's type to have the correct length;
+                        // if `x` has type `[u32; 4]` and one does `x[1..3]`, the type of the
+                        // projection will be `[u32; 4]`, and not `[u32; 2]` -- we fix this here,
+                        // by using a slice instead.
```
Nadrieril, line 323: The comment is wrong: rust does consider the projected type to be `[T]`, you can see the error [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ce84d88c04d94ca9c8cf7bb8546b32c9). Are you saying that we don't get that from hax? That's worrysome, I'd rather figure out what's happening here.

Nadrieril: Lol: https://github.com/cryspen/hax/blob/40961957dba9f338dcf89ee9272062b5c7657eb1/frontend/exporter/src/types/mir.rs#L847

Nadrieril: The whole thing on the hax side should just be using `place.ty()` instead of trying to recompute it.

Nadrieril: Please revert this change, we'll fix this from the hax side.

N1ark: oh wow that's terrible -- will do!

Nadrieril: Hax fix: https://github.com/cryspen/hax/pull/1457

N1ark: when merged this is ready too then

PR #672, opened 2025-05-05
Add `BuiltinFunId::CopyNonOverlapping`

N1ark:

PR #690, opened 2025-05-20
Add `CopyNonoverlapping` as a statement

N1ark:
Same as #672, but as a statement, since this is a non-diverging call (it can cause UB, but it cannot unwind).
Needed as a bunch of functions in `std::ptr` also use it, without going through `std::intrinsincs::copy_nonoverlapping`.

ci: use https://github.com/AeneasVerif/aeneas/pull/523


charon/src/ast/ullbc_ast.rs
```diff
@@ -23,6 +23,9 @@ pub enum RawStatement {
     Assign(Place, Rvalue),
     /// A call. For now, we don't support dynamic calls (i.e. to a function pointer in memory).
     SetDiscriminant(Place, VariantId),
+    /// Equivalent to std::intrinsics::copy_nonoverlapping; this is not modelled as a function
+    /// call as it cannot diverge
+    CopyNonOverlapping(Box<CopyNonOverlapping>),
```
sonmarcho, line 28: Is it really the proper way to treat `copy_nonoverlapping` as a separate statement rather than as a function call? In many situations (for instance, in a taint analysis) we do not care about the fact that it may diverge or not (what is the meaning of "diverge" here by the way?), while it is naturally handled in a manner similar to function calls. Also, I discussed with @Nadrieril about turning binops into function calls: the wrapping add operation can not panic either (I assume "diverge" means "panic").

charon/src/ast/ullbc_ast.rs
```diff
@@ -23,6 +23,9 @@ pub enum RawStatement {
     Assign(Place, Rvalue),
     /// A call. For now, we don't support dynamic calls (i.e. to a function pointer in memory).
```
sonmarcho, line 24: Just noticed that this comment is mis-placed.

charon/src/ast/ullbc_ast.rs
Nadrieril: It was annoying to transform to a function call because that requires braking up the list of statements and adding a new terminator. I'll clean that up when we figure out the big picture for binops/intrinsics.

sonmarcho: Why does it require adding a new terminator?

Nadrieril: Because `Call` is now a terminator (in preparation for adding unwinding paths).

sonmarcho: I had misunderstood your message: I thought you meant adding a new *kind* of terminator.

PR #696, opened 2025-05-23
Multiple documentation fixes

firefighterduck:
This PR fixes multiple issues with documentation:

* It fixes how `generate-ml` translates doc-comments, since the rustdoc code escape character \` does not work with odoc. This is implemented as a hacky string replacement for now. It would be better if we could use librustdoc to parse the doc comments and then print them in a way odoc understands, but that seems not possible at the moment.
* It fixes a bunch of rustdoc warnings in `charon`. I've tried to keep them as close to the intended results as possible.
* It adds a `make doc` on the top level as well as for `charon-ml` and fixes the `README.md` to refer to it.
* It fixes the symbolic link to the OCaml documentation (I'm not sure whether that's just a problem with my setup or more general.)

One caveat remains: odoc bundles highlight.js with only support for OCaml, Python, and a few others, but not Rust, so code highlighting only works if one exchanges the js file by hand. Since the OCaml documentation doesn't seem to be officially hosted anywhere, this seems to be only a problem for developers on their local machines.


Nadrieril:
Amazing, I've wanted to fix this doc syntax issue for a while :D

doc-ml.html
```diff
@@ -1 +1 @@
-./charon-ml/src/_build/default/_doc/_html/charon/index.html
\ No newline at end of file
+./_build/default/_doc/_html/charon/index.html
```
Nadrieril: Yeah that's been broken for a while because none of us use the ocaml doc x) Thanks for the fix!

charon/src/ast/builtins.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! **IMPORTANT**:
 //! When checking whether names are equal to one of the reference names below,
-//! we ignore the disambiguators (see [crate::names] and [crate::names_utils]).
+//! we ignore the disambiguators (see [crate::names] and [crate::names_utils](crate::ast::names_utils)).
```
Nadrieril: For this whole commit, I would prefer to fix the text too, not just the link:
```suggestion
//! we ignore the disambiguators (see [crate::names] and [crate::ast::names_utils]).
```

Nadrieril: oh perfect, you did that in the next commit

Nadrieril:
CI is failing because of a trailing space I think, just run `cargo fmt`

charon/src/ast/builtins.rs
firefighterduck: I've just looked for all instances of `(crate::` and fixed all but one (about the ids macro) occurrence according to VSCode. Would you like me to also inline the links to hax objects?

Nadrieril: Sure!

Issue #581, opened 2025-03-07
Feature request: extract type layouts


Nadrieril:
The one large missing bit of semantic information is type layouts. These are tricky to get with the current approach of charon because layouts only make sense for monomophic types. We should find a way to extract all the layouts relevant to the current crate.

firefighterduck:
After a short discussion with @N1ark (who does some layout computation on the OCaml side), I'd like to ask what the biggest blocker would be for having at least some layout information in (U)LLBC. For example, could we at least add size and alignment information for non-polymorphic types (which have a fixed layout, e.g. due to `repr(C)` or being compact by default, e.g. `(u32, u32)`)? I know that the full layout (especially for polymorphic types) is difficult/impossible to get, but maybe we (i.e., I would volunteer do something about it) can add support for some types incrementally.

PR #704, opened 2025-05-28
Add simple type layout information

firefighterduck:
Related to #581 

This PR adds simple type layout information for types with statically known layouts (i.e. no generic parameters etc.).


charon/src/ast/visitor.rs
```diff
@@ -51,7 +51,7 @@ use indexmap::IndexMap;
         TranslatedCrate, TypeDeclKind, TypeId, TypeVar, TypeVarId,
         ullbc_ast::BlockData, ullbc_ast::BlockId, ullbc_ast::ExprBody, ullbc_ast::RawStatement,
         ullbc_ast::RawTerminator, ullbc_ast::SwitchTargets, ullbc_ast::Terminator,
-        UnOp, Local, Variant, VariantId, LocalId, CopyNonOverlapping,
+        UnOp, Local, Variant, VariantId, LocalId, CopyNonOverlapping, u64,
```
Nadrieril, line 54: I prefer to `drive(skip)` on primitive fields like this

charon/tests/ui/type_layout.out
```diff
@@ -0,0 +1,23 @@
+# Final LLBC before serialization:
```
Nadrieril, line 1: The pretty-printed output won't include layout info, to test this add a test in `tests/crate_data.rs`

charon/src/ast/types.rs
```diff
@@ -345,6 +345,39 @@ pub enum PredicateOrigin {
...
+/// some of the layout parts are not available.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub struct Layout {
+    /// The size of the type in bytes.
+    #[drive(skip)]
+    pub size: Option<ByteCount>,
+    /// The alignment, in bytes.
+    #[drive(skip)]
+    pub align: Option<ByteCount>,
+    /// The offsets the discriminant as it is shared between all variants.
```
N1ark, line 372:
```suggestion
    /// The discriminant's offset, which is shared between all variants.
```

charon/src/ast/types.rs
```diff
@@ -367,6 +400,8 @@ pub struct TypeDecl {
     pub generics: GenericParams,
     /// The type kind: enum, struct, or opaque.
     pub kind: TypeDeclKind,
+    /// The layout of the type.
+    pub layout: Layout,
```
N1ark, line 404: exciting !!

charon/src/ast/types.rs
N1ark: yes I agree, phrasing is still not perfect :) maybe just "The discriminant's offset"?

charon/tests/crate_data.rs
```diff
@@ -711,3 +711,128 @@ fn known_trait_method_call() -> Result<(), Box<dyn Error>> {
...
+            variant_layouts
+        }
+    );
+    // NicheAdt
+    assert_eq!(
+        &crate_data.type_decls[5].layout,
+        &Layout {
+            size: Some(4),
+            align: Some(4),
+            ..default_layout.clone()
```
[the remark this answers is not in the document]
Nadrieril: Possibly what's missing is data about the discriminant encoding, i.e. which byte value gives which variant. This is distinct from the discriminant values returned by `read_discriminant`, which are already stored in `TypeDeclKind::Enum`.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -779,7 +781,101 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                    size,
+                    align,
+                    discriminant_offset,
+                    variant_layouts,
+                }
+            }
+            Err(_) => {
+                // If there is no sensible layout, we translate it to the empty layout.
+                // There doesn't seem to be any way to handle the errors in a better way.
+                Layout::default()
```
Nadrieril, line 874: Yeah, as I've said elsewwhere this isn't good: we should make sure to have the right number of variants and in each the right number of fields. Alternatively we could just store `Option<Layout>`, that's also ok.

charon/tests/crate_data.rs
```diff
@@ -711,3 +711,128 @@ fn known_trait_method_call() -> Result<(), Box<dyn Error>> {
...
+            size: Some(12),
+            align: Some(4),
+            discriminant_offset: None,
+            variant_layouts: Vector::new()
+        }
+    );
+    // GenericStruct
+    assert_eq!(&crate_data.type_decls[1].layout, &default_layout);
+    // UnsizedStruct
+    assert_eq!(&crate_data.type_decls[2].layout, &default_layout);
```
Nadrieril, line 779:
```suggestion
    // UnsizedStruct
    // We could in principle know the alignment, but rustc does not make it easy
    // to distinguish "dyn sized and dyn aligned" from "dyn sized and static aligned".
    assert_eq!(&crate_data.type_decls[2].layout, &default_layout);
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -779,7 +781,101 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                let (size, align) = if layout.is_sized() {
+                    (
+                        Some(layout.size().bytes()),
+                        Some(layout.align().abi.bytes()),
+                    )
+                } else {
+                    (None, None)
+                };
+                let mut variant_layouts = Vector::new();
+                let discriminant_offset = match layout.variants() {
```
Nadrieril, line 828: I'd like to try getting discriminant info for niches too. Surely it's not fundamentally different when niched and non-niched: in either case, we've got to read one or more bytes somewhere and deduce a discriminant value from it.

charon/src/ids/vector.rs
```diff
@@ -300,6 +307,16 @@ where
...
     }
+
+    /// Inserts the element at the given index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index` > `self.elem_count`.
+    pub fn insert(&mut self, index: I, element: T) {
+        self.vector.insert(index, Some(element));
+        self.elem_count += 1;
```
Nadrieril, line 318: This is almost surely an incorrect API for this type, because inserting not-at-the-end shifts the indices of the other entries, but this type cares very much about indices.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -779,7 +781,101 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+    /// Translate a type layout.
+    ///
+    /// Translates the layout as queried from rustc into
+    /// the more restricted [`Layout`].
+    #[tracing::instrument(skip(self))]
+    pub fn translate_layout(&self) -> Layout {
+        let tcx = self.t_ctx.tcx;
+        let rdefid = self.def_id.as_rust_def_id().unwrap();
+        let ty_env = hax::State::new_from_state_and_id(&self.t_ctx.hax_state, rdefid).typing_env();
+        let ty = tcx.type_of(rdefid).skip_binder();
```
Nadrieril, line 813: I'm careful about `skip_binder`s, because they tend to be mistakes. This one is ok so I'm adding a comment.
```suggestion
        // This `skip_binder` is ok because it's an `EarlyBinder`.
        let ty = tcx.type_of(rdefid).skip_binder();
```

charon/src/ast/types.rs
```diff
@@ -369,11 +371,14 @@ pub struct Layout {
...
     pub align: Option<ByteCount>,
-    /// The discriminant's offset, if any. Only relevant for
-    /// types with multiple variants and if the discriminant is
-    /// not in a niche.
+    /// The discriminant's offset, if any. Only relevant for types with multiple
+    /// variants.
     #[drive(skip)]
     pub discriminant_offset: Option<ByteCount>,
+    /// Whether the type is uninhabited.
+    /// This is required to differentiate ZSTs and uninhabited types.
```
```suggestion
    /// Whether the type is uninhabited, i.e. has any valid value at all.
    /// Note that uninhabited types can have arbitrary layouts: `(u32, !)` has space for the `u32`
    /// and `enum E2 { A, B(!), C(i32, !) }` may have space for a discriminant.
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -779,7 +781,118 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+        r_abi::FieldsShape::Primitive
+        | r_abi::FieldsShape::Union(_)
+        | r_abi::FieldsShape::Array { .. } => VariantLayout::default(),
+        r_abi::FieldsShape::Arbitrary { offsets, .. } => {
+            let mut v = Vector::with_capacity(offsets.len());
+            for (i, o) in offsets.iter_enumerated() {
+                // The fields should have a total increasing index, but to be
+                // absolutely sure, we insert them at the correct index.
+                v.insert(FieldId::from_usize(i.index()), Some(o.bytes()));
+            }
```
Nadrieril: You'll need the `TypeDecl` around to know how many variants/fields there are.
```suggestion
            // TODO: get the `fields: &Vector<FieldId, Field>` from the `TypeDecl`
            let mut v = fields.map_ref(|_| None);
            for (i, o) in offsets.iter_enumerated() {
                v[FieldId::from_usize(i.index())] = Some(o.bytes());
            }
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -858,20 +856,39 @@ impl ItemTransCtx<'_, '_> {
...
+                            }
+                            // It's the only one and will automatically be at index 0
+                            variant_layouts.push(VariantLayout {
+                                field_offsets: fields_of_single_variant,
+                            });
+                        }
+                        None
+                    }
+                    // An empty type has fields and certainly no discriminant.
                     _ => None,
```
Nadrieril:
```suggestion
                    r_abi::Variants::Empty => None,
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -858,20 +856,39 @@ impl ItemTransCtx<'_, '_> {
                             unreachable!()
                         }
                     }
+                    // If there is exactly one variant, i.e. if the type is a struct,
+                    // there is no discriminant, but still fields with offsets.
+                    r_abi::Variants::Single { index } if *index == r_abi::VariantIdx::ZERO => {
```
Nadrieril: Can you turn this `if` into an `assert!`? I don't know of a case where this isn't true, and if there is one we should handle it properly.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -858,20 +856,39 @@ impl ItemTransCtx<'_, '_> {
...
+                        // If the type has no fields (FieldsShape::{Primite,Union,Array}), we can skip this.
+                        if let r_abi::FieldsShape::Arbitrary { offsets, .. } = layout.fields() {
+                            let mut fields_of_single_variant = Vector::with_capacity(offsets.len());
+                            for (i, o) in offsets.iter_enumerated() {
+                                fields_of_single_variant
+                                    .insert(FieldId::from_usize(i.as_usize()), Some(o.bytes()));
+                            }
```
Nadrieril: Same as above:
```rust
                            // get `fields` from the `TypeDecl`
                            let mut fields_of_single_variant = fields.map_ref(|_| None);
                            for (i, o) in offsets.iter_enumerated() {
                                let field_id = FieldId::from_usize(i.as_usize());
                                fields_of_single_variant[field_id] = Some(o.bytes());
                            }
```

charon/tests/crate_data.rs
```diff
@@ -757,82 +758,324 @@ fn type_layout() -> anyhow::Result<()> {
...
 
     let default_layout = Layout::default();
     let default_variant_layout = VariantLayout::default();
 
     // SimpleStruct
+    let mut variant_layouts = Vector::with_capacity(1);
+    let mut single_variant_offsets = Vector::with_capacity(1);
+    single_variant_offsets.push(Some(0));
+    single_variant_offsets.push(Some(4));
+    single_variant_offsets.push(Some(8));
```
Nadrieril: You might enjoy `Vector::from([Some(0), Some(4), Some(8)])`

charon/src/ast/types.rs
```diff
@@ -345,6 +345,41 @@ pub enum PredicateOrigin {
...
+    #[drive(skip)]
+    pub align: Option<ByteCount>,
+    /// The discriminant's offset, if any. Only relevant for
+    /// types with multiple variants and if the discriminant is
+    /// not in a niche.
+    #[drive(skip)]
+    pub discriminant_offset: Option<ByteCount>,
+    /// Map from [VariantId] to the corresponding field layouts.
+    /// Structs are modeled as being the only variant, uninhabited types
+    /// as being without any variants, etc.
```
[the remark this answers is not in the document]
Nadrieril: You added the tests but forgot to update the comment

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Actually, have you seen any case where we have partial information? I expected `UnsizedStruct2` to give us that but apparently not.

charon/tests/crate_data.rs
```diff
@@ -711,3 +711,371 @@ fn known_trait_method_call() -> Result<(), Box<dyn Error>> {
     };
     Ok(())
 }
+
+#[test]
+fn type_layout() -> anyhow::Result<()> {
```
Nadrieril, line 716: This is turning into a lot of work to update the expected layouts. If you want, I propose that instead of adding a bunch of `assert!`s here, we build a `HashMap<String, Option<Layout>>` and dump it using `serde_json::to_writer_pretty` to a `tests/layouts.json` file.

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Yeah, the fact that the rustc side uses an `IndexVec` is sufficient for me to trust that all entries are there. Sorry for the back-and-forth, what you have here is ok and we can even remove the `Option`.

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: actually nevermind, see https://github.com/AeneasVerif/charon/pull/704#discussion_r2119145245

charon/tests/crate_data.rs
Nadrieril: or see [below](https://github.com/AeneasVerif/charon/pull/704#discussion_r2119142643)

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -779,7 +781,118 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+    match &variant_layout.fields {
+        r_abi::FieldsShape::Primitive
+        | r_abi::FieldsShape::Union(_)
+        | r_abi::FieldsShape::Array { .. } => VariantLayout::default(),
+        r_abi::FieldsShape::Arbitrary { offsets, .. } => {
+            let mut v = Vector::with_capacity(offsets.len());
+            for (i, o) in offsets.iter_enumerated() {
+                // The fields should have a total increasing index, but to be
+                // absolutely sure, we insert them at the correct index.
+                v.insert(FieldId::from_usize(i.index()), Some(o.bytes()));
```
Nadrieril: I'd prefer:
```suggestion
                let j = v.push(Some(o.bytes()));
                assert_eq!(j.index(), i.index());
```
so we can remove `Vector::insert`

PR #705, opened 2025-05-29
Misc updates to Charon-ML

sonmarcho:
This PR adds the `map_crate_with_span` visitor, which is a map visitor for LLBC crates (there is already `iter_crate_with_span`), and updates the name matcher to support patterns like `[@T; 32]`, where we use a concrete value for the length (rather than a variable).


Nadrieril:
Could you add a namematcher test for your new feature please?

PR #713, opened 2025-06-03
Add `FnDef`

N1ark:
Add `FnDef` as a type -- this represents a constant function, and is a ZST, unlike `Arrow` which is a function pointer.

Companion PR to https://github.com/cryspen/hax/pull/1487.

ci: use https://github.com/AeneasVerif/aeneas/pull/555
ci: use https://github.com/AeneasVerif/eurydice/pull/223

### DIFF
--- a/charon-ml/src/Substitute.ml
+++ b/charon-ml/src/Substitute.ml
@@ -629,6 +629,25 @@ let lookup_flat_method_sig (crate : 'a gcrate) (trait_id : trait_decl_id)
   in
   Some s
 
+(* Lookup the signature of a `Ty::FnDef`. *)
+let lookup_fndef_sig (crate : 'a gcrate) (fn_def : fun_decl_ref region_binder) :
+    (ty list * ty) region_binder option =
+  let fun_decl_id = fn_def.binder_value.id in
+  let* fun_decl = LlbcAst.FunDeclId.Map.find_opt fun_decl_id crate.fun_decls in
+  (* Substitute the signature to be valid under the binder. *)
+  let fn_sig =
+    st_substitute_visitor#visit_fun_sig
+      (make_subst_from_generics fun_decl.signature.generics
+         fn_def.binder_value.generics)
+      fun_decl.signature
+  in
+  (* Rebind everything *)
+  Some
+    {
+      binder_regions = fn_def.binder_regions;
+      binder_value = (fn_sig.inputs, fn_sig.output);
+    }
+
 (* Construct a set of generic arguments in the scope of `params` that matches
    `params` and feeds each required parameter with itself. E.g. given
    parameters for `<T, U> where U: PartialEq<T>`, the arguments would be `<T,


charon/src/ast/types.rs
```diff
@@ -704,12 +704,14 @@ pub enum TyKind {
     ///
     /// TODO: we don't translate this properly yet.
     DynTrait(ExistentialPredicate),
-    /// Arrow type, used for function pointers and reused for the unique type associated with each
-    /// function item.
+    /// Arrow type, used for function pointers.
```
```suggestion
    /// Function pointer type. This is a literal pointer to a region of memory that
    /// contains a callable function.
```

charon/src/ast/types.rs
```diff
@@ -704,12 +704,14 @@ pub enum TyKind {
...
     /// This is a function signature with limited generics: it only supports lifetime generics, not
-    /// other kinds of
-    /// generics.
+    /// other kinds of generics.
     Arrow(RegionBinder<(Vec<Ty>, Ty)>),
+    /// The unique type associated with each function item.
+    /// This is a function signature with limited generics: it only supports lifetime generics, not
+    /// other kinds of generics.
```
```suggestion
    /// The unique type associated with each function item. Each function item is given
    /// a unique generic type that takes as input the function's early-bound generics. This type
    /// is not generally nameable in Rust; it's a ZST (there's a unique value), and a value of that type
    /// can be cast to a function pointer or passed to functions that expect `FnOnce`/`FnMut`/`Fn` parameters.
```
If I'm right the `GenericArgs` won't match the item because it's missing the late-bound generics. I think the correct type would be `FnDef(RegionBinder<FunDeclRef>)` where the binder binds the late-bound variables.

charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -1667,6 +1667,30 @@ impl<C: AstFormatter> FmtWithCtx<C> for TraitTypeConstraint {
...
+        let ctx = &ctx.push_bound_regions(&self.regions);
+
+        write!(f, "fn")?;
+        if !self.regions.is_empty() {
+            write!(
+                f,
+                "<{}>",
+                self.regions.iter().map(|r| r.with_ctx(ctx)).format(", ")
+            )?;
+        }
```
Nadrieril, line 1699: Instead of this, could you adapt the `RegionBinder::fmt_split` helper to make it take a callback that formats the contents, and use that?

Nadrieril: oh you just copied that code ^^

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -296,6 +296,24 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                trait_refs,
+                fn_sig,
+            } => {
+                trace!("FnDef");
+                let f_id = self.register_fun_decl_id(span, def_id);
+                let generics = self.translate_generic_args(
+                    span,
+                    substs,
+                    trait_refs,
+                    Some(fn_sig.rebind(())),
```
Nadrieril: This will work, but will fill the late-bounds vars with `Region::Erased`. That's fine in a function body but this is type-level so we want to track lifetimes correctly. I'd like to do the thing I said above with `RegionBinder<FunDeclRef>`

charon/src/ast/types.rs
Nadrieril: And we might want to generate `FnOnce` etc impls for these types :D Not urgent though

N1ark: yes!! generating the FnOnces etc would def be a goal, rn i dont need it but yes it would be ideal, curious how well it will fit in the current system since these functions don't have a state param

charon/src/bin/charon-driver/translate/translate_types.rs
N1ark: for sure, will change

charon/src/pretty/fmt_with_ctx.rs
N1ark: yep :p

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -305,14 +305,32 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
             } => {
                 trace!("FnDef");
                 let f_id = self.register_fun_decl_id(span, def_id);
-                let generics = self.translate_generic_args(
+                let bound_substs = fn_sig.rebind((substs, trait_refs));
```
Nadrieril: I don't think this is accurate: the substs are not under that binder. In fact I think we should `translate_generic_args` outside of `translate_region_binder`, and then `generics.move_under_binder` to use them inside the binder. Maybe this would show the bug?
```rust
fn foo<'a, T>(_: &'a T) {}
fn bar<'b>() {
    let f = foo::<&'b ()>;
    f(&&());
}
```
Here if you move the `&'b ()` under the signature `<'a>` binder you'll get the wrong one. Not sure how to make this into something that actually crashes though.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -305,14 +305,32 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                        let generics = ctx.translate_generic_args(
+                            span,
+                            substs,
+                            trait_refs,
+                            None,
+                            GenericsSource::item(f_id),
+                        )?;
+                        let generics = generics.concat(
+                            GenericsSource::item(f_id),
+                            &ctx.innermost_binder().params.identity_args_at_depth(
```
Nadrieril: There's `identity_args` as alias for `identity_args_at_depth(_, zero())`

N1ark: nice thanks !

charon/src/transform/monomorphize.rs
```diff
@@ -124,9 +124,6 @@ impl VisitAst for UsageVisitor<'_> {
             TyKind::Adt(TypeId::Adt(id), gargs) => {
                 self.found_use_ty(id, gargs);
             }
-            TyKind::FnDef(id, gargs) => {
-                self.found_use_fn(id, gargs);
-            }
```
Nadrieril: Why did you remove this?

N1ark: it would monomorphize `for<'a> foo<'a>`, which is not the actual use of the fn :)

Nadrieril: If you don't want to deal with the binder, it's fine to erase regions during mono (there's `RegionBinder::erase` for you)

charon/src/bin/charon-driver/translate/translate_types.rs
N1ark: Something like this?
```rust
let f_id = self.register_fun_decl_id(span, def_id);
let generics = self.translate_generic_args(
    span,
    substs,
    trait_refs,
    None,
    GenericsSource::item(f_id),
)?;

let fnref = self.translate_region_binder(span, fn_sig, |ctx, _| {
    let generics = generics.move_under_binder().concat(
        GenericsSource::item(f_id),
        &ctx.innermost_binder()
            .params
            .identity_args(GenericsSource::item(f_id)),
    );
    Ok(FunDeclRef {
        id: f_id,
        generics: Box::new(generics),
    })
})?;
TyKind::FnDef(fnref)
```

charon/src/transform/monomorphize.rs
Nadrieril: > it would monomorphize `for<'a> foo<'a>`, which is not the actual use of the fn :)

why not?
```rust
fn foo<'a, T>(_: &'a T) {}
fn takes_closure(_: impl for<'a> Fn(&'a u32)) {}
fn main() {
    takes_closure(foo);
}
```

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: yep, exactly!

charon/src/transform/monomorphize.rs
N1ark: it was monomorphising the fn twice, e.g. for this code:
```rust
fn foo<'a, T>(x: &'a T) {}

fn bar() {
    let fooint1: fn(&u8) = foo;
    fooint1(&11);
}
```
it was generating this:
```
fn test_crate::foo::<'_0_0, u8>(@1: &missing('_0_0) (u8))
{
    let @0: (); // return
    let x@1: &'_ (u8); // arg #1

    @0 := ()
    @0 := ()
    return
}

fn test_crate::foo::<'_, u8>(@1: &'_ (u8))
{
    let @0: (); // return
    let x@1: &'_ (u8); // arg #1

    @0 := ()
    @0 := ()
    return
}
```

Nadrieril: ah yeah, you have to be careful with lifetimes when monomorphizing, that's for sure :p what does this do today:
```rust
fn foo<'a, T>(_: &'a T) {}
fn takes_closure(_: impl for<'a> Fn(&'a u32)) {}
fn main() {
    takes_closure(foo);
}
```

N1ark: your example generates this (without special-casing FnDef), which seems correct too?
```

fn test_crate::takes_closure::<for<'a> foo<'a, u32>[Sized<u32>]>(@1: for<'a> foo<'a, u32>[Sized<u32>])
{
    let @0: (); // return
    let @1: for<'a> foo<'a, u32>[Sized<u32>]; // arg #1

    @0 := ()
    drop @1
    @0 := ()
    return
}
```

N1ark: well the full thing is this (ignoring Fn traits etc):
```

fn test_crate::takes_closure::<for<'a> foo<'a, u32>[Sized<u32>]>(@1: for<'a> foo<'a, u32>[Sized<u32>])
{
    let @0: (); // return
    let @1: for<'a> foo<'a, u32>[Sized<u32>]; // arg #1

    @0 := ()
    drop @1
    @0 := ()
    return
}

fn test_crate::foo::<'_, u32>(@1: &'_ (u32))
{
    let @0: (); // return
    let @1: &'_ (u32); // arg #1

    @0 := ()
    @0 := ()
    return
}

// Full name: test_crate::main
fn main()
{
    let @0: (); // return
    let @1: (); // anonymous local

    storage_live(@1)
    @1 := test_crate::takes_closure::<for<'a> foo<'a, u32>[Sized<u32>]>(const (test_crate::foo::<'_, u32><'_>))
    storage_dead(@1)
    @0 := ()
    @0 := ()
    return
}
```

Nadrieril: That does look good, I guess I'm surprised why this works at all. Where does it find a reference to the function item if not through `FnDef`? Also, if you `erase()` the binder I think that would get rid of the duplication you mentioned.

N1ark: through the parameter ! it has a ConstantExpr, which itself has a `FnPtr` with the right parameters!

N1ark: basically a fn is used either via a ConstantExpr, which contains a FnPtr, or a call, that also has one, so it works out nicely

Nadrieril: Ok interesting. We still need to monomorphize all types though. I'm ok with `RegionBinder::new_empty(fix_args(binder.erase())` for now.

charon/tests/ui/monomorphization/fndefs-casts.rs
```diff
@@ -0,0 +1,23 @@
+//@ charon-args=--monomorphize
+// Ensures casts of FnDefs monomorphise properly
+
+fn foo<'a, T>(x: &'a T) {}
+
+fn takes_closure(_: impl for<'a> Fn(&'a u32)) {}
```
Nadrieril: actually could you add a call to the closure?

charon-ml/src/generated/Generated_Types.ml
```diff
@@ -422,11 +422,16 @@ and ty =
...
-      (** Arrow type, used for function pointers and reused for the unique type associated with each
-          function item.
+      (** Function pointer type. This is a literal pointer to a region of memory that
+          contains a callable function.
           This is a function signature with limited generics: it only supports lifetime generics, not
-          other kinds of
-          generics.
+          other kinds of generics.
+       *)
+  | TFnDef of fun_decl_ref region_binder
```
sonmarcho, line 399: Where is this being used?

N1ark: anywhere you use function pointers, you need to first refer to the type of the original function, e.g.
```rust
fn foo(x: u8) -> u8 { x }
let f: fn(u8) -> u8 = foo;
```
Here, `foo` has type `FnDef` of that function's ID, with no generics/binders

N1ark: A `FnDef` is a ZST, a `FnPtr` (or `Arrow`, as it's called in Charon) is an actual pointer to a function that exists in memory

Nadrieril: Also whenever you pass a function name as a closure

charon/src/transform/monomorphize.rs
N1ark: done, let me know if that's what you had in mind, and if there's anything else to do for this PR :p

charon/src/ast/types.rs
```diff
@@ -747,12 +747,16 @@ pub enum TyKind {
...
     /// This is a function signature with limited generics: it only supports lifetime generics, not
-    /// other kinds of
-    /// generics.
+    /// other kinds of generics.
     Arrow(RegionBinder<(Vec<Ty>, Ty)>),
+    /// The unique type associated with each function item. Each function item is given
+    /// a unique generic type that takes as input the function's early-bound generics. This type
+    /// is not generally nameable in Rust; it's a ZST (there's a unique value), and a value of that type
+    /// can be cast to a function pointer or passed to functions that expect `FnOnce`/`FnMut`/`Fn` parameters.
+    FnDef(RegionBinder<FunDeclRef>),
```
Nadrieril, line 803:
```suggestion
    /// There's a binder here because charon function items bind both early and late-bound variables;
    /// that binder binds the late-bound lifetimes (those that could appear in a function pointer type
    /// like `for<'a> fn(&'a u32)`).
    FnDef(RegionBinder<FunDeclRef>),
```

PR #716, opened 2025-06-04
`TypeDecl.kind`->`type_kind`, add `kind: ItemKind`

N1ark:
- Rename `TypeDecl.kind` to `type_kind` (consistent with `GlobalDecl`)
- Add `TypeDecl.kind: ItemKind`, with the type's kind (either `TopLevel` or `Closure`)


Nadrieril:
> Reworked a bit according to what you said @Nadrieril , getting some errors with generics (missing regions in `ClosureInfo`), should maybe`fn_mut_impl` and `fn_impl` be behind a region binder?... or maybe we skip ClosureInfos when checking for generics because really it doesn't matter? idk

All three of them should be behind a region binder indeed, that binders the same regions as the PolyFnSig that corresponds to the closure.

charon/src/bin/charon-driver/translate/translate_closures.rs
```diff
@@ -63,8 +63,31 @@ impl ItemTransCtx<'_, '_> {
         span: Span,
         args: &hax::ClosureArgs,
     ) -> Result<ClosureInfo, Error> {
+        let mut impl_ref_for = |kind: ClosureKind| -> Result<RegionBinder<TraitImplRef>, Error> {
+            self.translate_region_binder(span, &args.untupled_sig, |ctx, _| {
+                let mut impl_ref = ctx.translate_closure_impl_ref(span, args, kind)?;
+                impl_ref
+                    .generics
+                    .regions
+                    .extend(args.tupled_sig.bound_vars.iter().map(|_| Region::Erased));
```
Nadrieril: `translate_closure_impl_ref` already adds these extra regions, did you not get generics errors from that?

N1ark: Only adds them if it's a different def ID! In this case however the ClosureInfo belongs to the same def id so we must do it manually

Nadrieril: Yeah this is very much not what's intended, I've been looking into it and something very weird is happening

Nadrieril: Like, obviously the contract of `translate_closure_impl_ref` is that it should return a valid `TraitImplRef`, but it seems I'm messing something up somewhere.

Nadrieril: Yep: https://github.com/cryspen/hax/pull/1522, https://github.com/AeneasVerif/charon/pull/744

PR #718, opened 2025-06-05
Remove the version pin in charon-ml/.ocamlformat

sonmarcho:
The default version of ocamlformat is not 0.26.2 anymore. Instead of regularly updating the pinned version I suggest removing the pin altogether: if it leads to issues in the future (in particular, changes in the formatting depending on the version) I will put it back.

### DIFF
--- a/charon-ml/.ocamlformat
+++ b/charon-ml/.ocamlformat
@@ -1,5 +1,4 @@
 profile = default
-version = 0.26.2
 margin = 80
 
 break-cases = fit-or-vertical


Nadrieril:
I'm pretty sure I've observed that different versions format things differently. I would much prefer if you updated the pinned version to whichever you think is a sensible default.

Nadrieril:
I guess we'll see, just making a mental note in case something weird happens wrt formatting.

PR #719, opened 2025-06-06
added dst-meta-kind info

ssyram:
Added the Dynamically Sized Type (DST) meta-data info in `TypeDecl`. The key function used is the [`struct_tail_for_codegen`](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/ty/util.rs.html#206-220) function from the Rustc internal library.

Notably, the current implementation mimics that of the layout info, and will not handle the monomorphised types. This should be fixed in the future.

As a potential revision, a type can either be: (1) generic (types/consts, regardless of lifetime), (2) statically sized (hence with layout) or (3) dynamically sized. And the current `TypeDecl` seems to allow potentially wrong states, e.g., a type simultaneously with layout and dst-meta-info. I suggest revising this by introducing something like meta to make such illegal states irrepresentable.

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -353,16 +353,29 @@ type ByteCount = u64;
 /// Maps fields to their offset within the layout.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
 pub struct VariantLayout {
-    /// The offset of each field. `None` if it is not knowable at this point, either because of
-    /// generics or dynamically-sized types.
+    /// The offset of each field.
     #[drive(skip)]
     pub field_offsets: Vector<FieldId, ByteCount>,
+}
+
+/// Layout of the discriminant with its offset and representation type.
+///
+/// Does not include information about the value range.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum DiscriminantLayout {
+    Direct {
+        /// The offset of the discriminant in bytes.
+        offset: ByteCount,
+        /// The representation type of the discriminant.
+        repr: IntegerTy,
+    },
+    Niche, // TODO: Add more useful information about niches here in the future.
 }
 
 /// Simplified type layout information.
 ///
 /// Does not include information about niches.
-/// If the type does not have fully known layout (e.g. it is ?Sized)
+/// If the type does not have a fully known layout (e.g. it is ?Sized)
 /// some of the layout parts are not available.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
 pub struct Layout {
@@ -372,9 +385,10 @@ pub struct Layout {
     /// The alignment, in bytes.
     #[drive(skip)]
     pub align: Option<ByteCount>,
-    /// The discriminant's offset, if any. Only relevant for types with multiple variants.
+    /// The discriminant's layout, if any. Only relevant for types with multiple variants.
+    ///
     #[drive(skip)]
-    pub discriminant_offset: Option<ByteCount>,
+    pub discriminant_layout: Option<DiscriminantLayout>,
     /// Whether the type is uninhabited, i.e. has any valid value at all.
     /// Note that uninhabited types can have arbitrary layouts: `(u32, !)` has space for the `u32`
     /// and `enum E2 { A, B(!), C(i32, !) }` may have space for a discriminant.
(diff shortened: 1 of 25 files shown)


ssyram:
Thanks so much! Will revise now!

charon/src/ast/types.rs
```diff
@@ -390,13 +390,18 @@ pub struct Layout {
...
-/// For trait objects, this is the vtable of the trait.
-/// For user defined DSTs, this is inductively inherited from their last field.
+/// The metadata stored in a pointer. That's the information stored in pointers alongside
+/// their address. It's empty for `Sized` types, and interesting for unsized
+/// aka dynamically-sized types.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
-pub enum DstMetaKind {
+pub enum PtrMetadata {
+    /// Types that need no metadata, namely `T: Sized` types.
+    NoMetadata,
```
Nadrieril: In Rust we try not to repeat the type name in variant names. To avoid ocaml clashes we can rename individual variants like so:
```suggestion
    #[charon::rename("NoMetadata")]
    None,
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -396,10 +396,10 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
         // call the key method
         match tcx.struct_tail_for_codegen(ty, ty_env).kind() {
-            ty::Foreign(..) => None,
-            ty::Str | ty::Slice(..) => Some(DstMetaKind::Length),
-            ty::Dynamic(..) => Some(DstMetaKind::VTable(VTable)),
-            _ => None,
+            ty::Foreign(..) => Some(PtrMetadata::NoMetadata),
+            ty::Str | ty::Slice(..) => Some(PtrMetadata::Length),
+            ty::Dynamic(..) => Some(PtrMetadata::VTable(VTable)),
+            _ => Some(PtrMetadata::NoMetadata),
```
Nadrieril: That's not correct: the type could be generic in which case we should return `None`. You'll have to add more cases to this match.

charon/tests/dynamic-sized-type-info.rs
```diff
@@ -7,9 +7,9 @@ use indexmap::IndexMap;
 use util::*;
 
 #[test]
-fn dst_meta_kind() -> anyhow::Result<()> {
-  let crate_data = translate_rust_text(
-    r#"
+fn ptr_metadata() -> anyhow::Result<()> {
+    let crate_data = translate_rust_text(
+        r#"
     type Alias = str;
```
Nadrieril, line 14: Could you add tests with generics?
```rust
#![feature(ptr_metadata)]
use std::ptr::Thin;
struct GenericInLastField<T> {
    x: u32,
    y: T,
}
struct GenericNotLastField<T> {
    x: u32,
    y: T,
    z: u32,
}
struct GenericBehindIndirection<T> {
    x: u32,
    y: Box<T>,
}
// Charon doesn't recognize that we know the metadata in this case. That's ok.
struct ThinGeneric<T: Thin> {
    x: u32,
    y: T,
}
```

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: The generic is checked at the beginning?

ssyram:
Cases revised as advised.

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Oh sorry I forgot about that. Could you not check the generic at the beginning so that we can handle some of the test cases I proposed?

Nadrieril: Actually nevermind, that can completely be a separate PR

ssyram:
Thanks, content added. I think to get the exact info, we should inquire whether there is the `?Sized` clause for the final type variable. But I don't know how to inquire, would you please provide some pointers?

Nadrieril:
Hm, that would involve invoking the trait solver in a fallible way. I do not know how to do that. I'm ok to merge without that information and figure it out later if it proves important.

ssyram:
Well, then it would be hard for us to fully trust that this info is accurate, I would add more comments to state this.

Nadrieril:
I'll take care of it, we probably want to switch to 0.27.0 anyway.

Nadrieril:
Thanks for the PR!

PR #721, opened 2025-06-08
Fix infinite loops in mono

### LINKED ISSUE #675: Bug: infinite loop in `--monomorphize`
Compiling the following code with `charon rustc --monomorphize --extract-opaque-bodies` loop infinitely:
```rust
#![feature(btree_extract_if)]
use std::collections::BTreeSet;
use std::mem;

#[derive(PartialEq, Eq, PartialOrd, Ord)]
pub enum Foo {
    A(&'static str),
    _B,
    _C,
}

pub fn main() {
    let mut b = BTreeSet::new();
    b.insert(Foo::A("\'"));
}
```

N1ark:
Fixes #675 -- was caused by monomorphization detecting types in `PathElem::Impl`s as uses of that type

### DIFF
--- a/charon/tests/ui/monomorphization/rec-adt.out
+++ b/charon/tests/ui/monomorphization/rec-adt.out
@@ -1,0 +1,68 @@
+# Final LLBC before serialization:
+
+// Full name: core::marker::Sized
+#[lang_item("sized")]
+pub trait Sized<Self>
+
+// Full name: core::option::Option
+#[lang_item("Option")]
+pub enum Option<T>
+where
+    [@TraitClause0]: Sized<T>,
+{
+  None,
+  Some(T),
+}
+
+// Full name: core::ptr::non_null::NonNull
+#[lang_item("NonNull")]
+pub opaque type NonNull<T>
+
+// Full name: test_crate::Node
+struct Node<T>
+where
+    [@TraitClause0]: Sized<T>,
+{
+  next: Option<NonNull<Node<T>[@TraitClause0]>>[Sized<NonNull<Node<T>[@TraitClause0]>>],
+  element: T,
+}
+
+// Full name: test_crate::LinkedList
+pub struct LinkedList<T>
+where
+    [@TraitClause0]: Sized<T>,
+{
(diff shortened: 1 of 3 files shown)


charon/tests/ui/monomorphization/rec-adt.rs
```diff
@@ -0,0 +1,21 @@
+use std::marker::PhantomData;
```
Nadrieril, line 2:
```suggestion
//@ charon-args=--monomorphize
use std::marker::PhantomData;
```

charon/tests/ui/monomorphization/rec-adt.out
```diff
@@ -0,0 +1,38 @@
+# Final LLBC before serialization:
+
+// Full name: core::marker::Sized
+#[lang_item("sized")]
+pub trait Sized<Self>
+
+pub struct test_crate::LinkedList::<u8> {
+  head: core::option::Option::<NonNull<Node<u8>[Sized<u8>]>>,
```
Nadrieril, line 8: How come the output doesn't include that type?

N1ark: oh that is.. a good question????

N1ark: wait i think it's because of short names??? because the type decl is in the .llbc file so ??

Nadrieril: Short names should not change anything, they're only used for pretty-printing

Nadrieril: I do see the item in the llbc file :thinking: :thinking: tha'ts weird

N1ark: no ideak actually - the llbc file has all the right information, i dont know why its skipped at printing

Nadrieril: Oh it's because of `ordered_decls`. We use that as a basis for printing items, but it's missing `TypeDeclId(6)` for some reason

Nadrieril: Let's fix that separately

PR #724, opened 2025-06-11
Add more Discriminant layout information

firefighterduck:
While working with the new layout information, I found that we currently don't have enough information about the discriminant of enums. Specifically, we don't know their size (and how they are represented in memory)[^1]. To fix this, I add the representation type in memory, which implicitly also encodes the size of the field. Currently, this does not support niches with non-integer types (i.e. float/reference according to `rustc_abi::Scalar`), since Charon has no good way to handle these uniformly.

[^1]: The `TypeDeclKind` has the discriminants of all variants as `ScalarValue`, but these are not equivalent to the representation in memory and have potentially different sizes.

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -353,16 +353,24 @@ type ByteCount = u64;
 /// Maps fields to their offset within the layout.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
 pub struct VariantLayout {
-    /// The offset of each field. `None` if it is not knowable at this point, either because of
-    /// generics or dynamically-sized types.
+    /// The offset of each field.
     #[drive(skip)]
     pub field_offsets: Vector<FieldId, ByteCount>,
+}
+
+/// Layout of the discriminant with its size and offset.
+///
+/// Does not include information about the value range or encoding.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DiscriminantLayout {
+    pub offset: ByteCount,
+    pub size: ByteCount,
 }
 
 /// Simplified type layout information.
 ///
 /// Does not include information about niches.
-/// If the type does not have fully known layout (e.g. it is ?Sized)
+/// If the type does not have a fully known layout (e.g. it is ?Sized)
 /// some of the layout parts are not available.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
 pub struct Layout {
@@ -372,9 +380,9 @@ pub struct Layout {
     /// The alignment, in bytes.
     #[drive(skip)]
     pub align: Option<ByteCount>,
-    /// The discriminant's offset, if any. Only relevant for types with multiple variants.
+    /// The discriminant's layout, if any. Only relevant for types with multiple variants.
     #[drive(skip)]
-    pub discriminant_offset: Option<ByteCount>,
+    pub discriminant_layout: Option<DiscriminantLayout>,
     /// Whether the type is uninhabited, i.e. has any valid value at all.
     /// Note that uninhabited types can have arbitrary layouts: `(u32, !)` has space for the `u32`
     /// and `enum E2 { A, B(!), C(i32, !) }` may have space for a discriminant.
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -358,13 +358,17 @@ pub struct VariantLayout {
     pub field_offsets: Vector<FieldId, ByteCount>,
 }
 
-/// Layout of the discriminant with its size and offset.
+/// Layout of the discriminant with its offset and representation type.
 ///
-/// Does not include information about the value range or encoding.
+/// Does not include information about the value range.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct DiscriminantLayout {
+    /// The offset of the discriminant in bytes.
     pub offset: ByteCount,
-    pub size: ByteCount,
+    /// The representation type of the discriminant.
+    /// If the discriminant is in a niche of a non-scalar type,
+    /// this is `None`.
+    pub repr: Option<IntegerTy>,
 }
 
 /// Simplified type layout information.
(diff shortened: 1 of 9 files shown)


charon/src/ast/types.rs
```diff
@@ -353,16 +353,28 @@ type ByteCount = u64;
...
-    /// generics or dynamically-sized types.
+    /// The offset of each field.
     #[drive(skip)]
     pub field_offsets: Vector<FieldId, ByteCount>,
 }
 
+/// Layout of the discriminant with its offset and representation type.
+///
+/// Does not include information about the value range.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
Nadrieril: Do we know what to do with the resulting integer? Should there be a `discr: ScalarValue` in each `VariantLayout` that says which integer means which variant?

Nadrieril: Actually should be `discr: Option<ScalarValue>` to support niches. The niched variant would have `discr: None` which is understood to mean that if we read a discriminant that's not for one of the other variants then we're in the niched one.

charon/src/ast/types.rs
```diff
@@ -353,16 +353,28 @@ type ByteCount = u64;
...
+pub struct DiscriminantLayout {
+    /// The offset of the discriminant in bytes.
+    pub offset: ByteCount,
+    /// The representation type of the discriminant.
+    /// If the discriminant is in a niche of a non-scalar type,
+    /// this is `None`.
+    pub repr: Option<IntegerTy>,
```
Nadrieril: If we wanted to support floats we could use the badly-named `LiteralTy` here. Are you sure that there can be float discriminants though?

charon/src/ast/types.rs
```diff
@@ -353,16 +353,28 @@ type ByteCount = u64;
...
+/// Layout of the discriminant with its offset and representation type.
+///
+/// Does not include information about the value range.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DiscriminantLayout {
+    /// The offset of the discriminant in bytes.
+    pub offset: ByteCount,
+    /// The representation type of the discriminant.
+    /// If the discriminant is in a niche of a non-scalar type,
+    /// this is `None`.
```
```suggestion
    /// If the discriminant is in a niche or a non-scalar type,
    /// this is `None`.
```

Nadrieril: In fact that looks incorrect: for `NicheAdt` in the tests we do get something. Let's just support niches.

charon/src/ast/types.rs
Nadrieril: Given that `#[repr(f32)]` is an error, I'm confident that the answer is "no". Hence let's make this field non-optional.

charon/src/ast/types.rs
Nadrieril: To test for an interesting niche:
```rust
enum NonZeroNiche {
    A(char),
    B,
    C,
}
```
This should pick discriminants within the surrogate code points.

Nadrieril: And to test for just weird discriminant values:
```rust
#[repr(u32)]
enum ArbitraryDiscriminants {
    A(String) = 12,
    B(u32) = 43,
    C = 123456,
}
```

firefighterduck: The `LayoutData` documentation is a bit ambiguous about how the tag (the in-memory representation) and the discriminant actually correlate. I guess, that it should be a simple cast for integers, but then we can probably simply use the `ScalarValue`s in the `TypeDeclKind::Enum` variant map.

charon/src/ast/types.rs
firefighterduck: In `rustc_abi` the tag can be an ~~integer, a float or a pointer (presumably related to `NonZero` and `NonNull`).~~ After looking into it  a bit more, it seems that the used `Primitive` type describes in what kind of register a value can be put, not necessarily that these are all possible tag types. ~~So depending on if we can/want to support these niche representations, we might have a float.~~

charon/src/ast/types.rs
firefighterduck: The support for integer niches was intentional (and motivated by `NicheAdt`). Depending on whether we want to support other niches as well (and in general have more information about them in the layout), we could allow more than just integers.

charon/src/ast/types.rs
Nadrieril: Yeah, `Primitive` is a general-purpose internal type, I'm pretty confident there a no non-integer discriminants. I propos to emit an error if we find such a type, and whoever needs that can open an issue.

charon/src/ast/types.rs
Nadrieril: I'm not understanding you. If I may clarify: the type you're getting here is the type of the discriminant. This is entirely separate from the type of the value that may live in the same bytes in case of a niched discriminant. We don't care about that type. E.g. for `Option<&'a T>` I think the discriminant type is `u8`.

If we also add the assumption that discriminants are integers, then we get that the current implementation handles all niches and guarantees that `repr` is `Some`. Hence my proposal to remove the `Option`.

charon/src/ast/types.rs
firefighterduck: This unfortunately makes `Option<NonNull>` also an error (if I understand you correctly), but sounds like a reasonable solution for now.

charon/src/ast/types.rs
Nadrieril: Ah, [`TagEncoding`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_abi/enum.TagEncoding.html) explains it: for non-niched, it's indeed an integer cast; for niched it's more complicated but still explained. That's different enough from the `ScalarValue`s in `TypeDeclKind::Enum` (these are what is output by the `read_discriminant` intrinsic) that I'd prefer to store the layout-relevant discriminant values in the `VariantLayout`s. 

That can totally be a separate PR btw.

charon/src/ast/types.rs
firefighterduck: Huh, it seems I have to retract my previous assessment. Making it a hard error leads to lots of tests failing (they all have some kind of pointer niche). Do you think that it makes more sense to have no discriminant layout (i.e. no offset) at all for enums with niches?

Nadrieril: Oh, then I'm the one misunderstanding what this type means :thinking:

charon/src/ast/types.rs
Nadrieril: Yep no I'm the one confused

Issue #689, opened 2025-05-20
Feature request: disable custom Box translation


N1ark:
Currently, Box is treated as a special type, constructed from a pointer; this isn't accurate, as in reality a `Box<T>` has two fields: a `Unique<T>` and an allocator. Similarly, accesses to the first field are translated in Charon as a dereference of the inner pointer, removing a layer of indirection.

Some of the code in `std::alloc::boxed` doesn't compile with Charon, because it attempts accessing the allocator of the `Box`, which doesn't exist on the Charon side. I would be interested in having a flag to disable this custom `Box` handling and use the naive Rust translation, to support more code out of the box.

Either way, the case in which the accessed field of `Box` is not field 0 but 1 should be accessed more gracefully; rather than an assert, it could be replaced with a new projection type (e.g. `BoxAllocator`?) for now, to allow full translation.

Currently, these two fragments of code fail translating properly:
```rust
let x = 0u8;
let ptr = &x as *const u8 as *mut u8;
unsafe { Box::from_raw(ptr) };
```
```
thread 'rustc' panicked at src/bin/charon-driver/translate/translate_bodies.rs:634:25:
assertion failed: matches!(&type_id, TypeId::Adt(_))
```
```rust
let _ = Box::leak(Box::new(0));
```
```
thread 'rustc' panicked at src/bin/charon-driver/translate/translate_bodies.rs:270:41:
assertion failed: field_id == FieldId::ZERO
```

Nadrieril:
Absolutely, we should make this optional

PR #725, opened 2025-06-11
Add `--raw-boxes`

N1ark:
Add a `--raw-boxes` flag to disable box translation
Rename `MutPtr` -> `ConstPtr`, with added ref kind
Translate `ConstantExprKind::RawBorrow { mutability: false, .. }`

Fixes https://github.com/AeneasVerif/charon/issues/689

### DIFF
--- a/charon/src/ast/expressions.rs
+++ b/charon/src/ast/expressions.rs
@@ -475,11 +475,11 @@ pub enum RawConstantExpr {
     /// We eliminate this case in a micro-pass.
     #[charon::opaque]
     Ref(Box<ConstantExpr>),
-    /// A mutable pointer to a mutable static.
+    /// A pointer to a mutable static.
     ///
     /// We eliminate this case in a micro-pass.
     #[charon::opaque]
-    MutPtr(Box<ConstantExpr>),
+    Ptr(RefKind, Box<ConstantExpr>),
     /// A const generic var
     Var(ConstGenericDbVar),
     /// Function pointer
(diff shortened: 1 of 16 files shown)


charon/src/ast/expressions.rs
```diff
@@ -479,7 +479,7 @@ pub enum RawConstantExpr {
     ///
     /// We eliminate this case in a micro-pass.
     #[charon::opaque]
-    MutPtr(Box<ConstantExpr>),
+    ConstPtr(RefKind, Box<ConstantExpr>),
```
Nadrieril: Why `ConstPtr` if it can be mutable too?

N1ark: because it's a pointer to a constexpr but admittedly it's a bad name hahahah -- other ideas?

Nadrieril (changes requested):
Looks great, I would have done it in a much more complicated way :D One nit: `ConstPtr` should be just `Ptr`, or maybe `RawPtr`

charon/src/ast/expressions.rs
Nadrieril: Oh, I'm silly, I read it as `*const T`. Uhhh maybe `PtrToConst`?

Nadrieril: Wait no it's already in `ConstantExpr`, we know it's to a constant. I like `RawPtr`.

N1ark: i think that's also the name of the aggregate for pointers but sure !

Nadrieril: `ConstantExpr`s are mostly temporary things that we convert to aggregates and other rvalues, so that's a good thing.

PR #727, opened 2025-06-12
Translate blocks as an ocaml list instead of a recursive structure

Nadrieril:


charon-ml/src/LlbcAstUtils.ml
```diff
@@ -20,77 +20,24 @@ let get_fun_args (fun_decl : fun_decl) : local list option =
...
-        List.map (fun (svl, br) -> (svl, chain_statements br st)) branches
-      in
-      let otherwise = chain_statements otherwise st in
-      SwitchInt (op, int_ty, branches, otherwise)
-  | Match (op, branches, otherwise) ->
-      let branches =
-        List.map (fun (svl, br) -> (svl, chain_statements br st)) branches
-      in
-      let otherwise = Option.map (fun b -> chain_statements b st) otherwise in
-      Match (op, branches, otherwise)
```
sonmarcho, line 92: I had forgotten those clumsy helpers: happy to see that we can finally get rid of them :)

PR #736, opened 2025-06-17
Better handle binary operations

N1ark:
Update hax to have correct translation of `Unchecked{Add,Sub,Mul}`.
I also disabled UB checks (which affected a usage of `Box<T>`), though I can get rid of that if you prefer; it just seemed sensible, and like something Charon does :p

Depends on https://github.com/cryspen/hax/pull/1513

(Note I specify another hax branch than that of the PR in `Cargo.toml`, while waiting for you to close https://github.com/AeneasVerif/charon/pull/730)

ci: use https://github.com/AeneasVerif/eurydice/pull/224
ci: use https://github.com/AeneasVerif/aeneas/pull/557


Nadrieril:
> (Note I specify another hax branch than that of the PR in Cargo.toml, while waiting for you to close https://github.com/AeneasVerif/charon/pull/730)

Yep, you can see the "check-hax-commit" job checks that and prevents us from accidentally merging with a non-upstream hax commit :).

> I also disabled UB checks

Could you make that optional? These checks provide good value for some of our users.

Nadrieril:
In fact you can just pass `--rustc-arg=-Zno-ub-checks` or however it's called, this doesn't require anything special in charon.

N1ark:
sure yep !

charon/src/transform/remove_dynamic_checks.rs
```diff
@@ -143,6 +175,9 @@ fn remove_dynamic_checks(
         }, rest @ ..]
             if cond == is_zero =>
         {
+            make_overflow_panic(rest, |bop, _, r| {
+                matches!(bop, BinOp::Div(_) | BinOp::Rem(_)) && equiv_op(r, y_op)
+            });
             rest
```
Nadrieril: If we didn't find the operation to change then we mustn't remove the `assert` (i.e. we mustn't return `rest`)

N1ark: right! thanks

PR #738, opened 2025-06-18
Add more layout information - Discriminants and Tags

firefighterduck:
This PR adds more layout information about discriminants and their memory representation as tags. It is still a draft, since there are a few possible design choice in implementing this:

* Currently implemented: Store the same information as in rustc and be able to translate between tags and discriminants on the fly. This allows for a more flexible handling of tags.
* Pre-compute the tags and store them in the layout. This would make handling of tags in OCaml a lot easier.
* It is possible to either follow the current code-gen implementations in rustc and assume that variant IDs and discriminants coincide in the case of niche encoded tags ...
* ... or try to be more robust and explicitly translate between them.

Lastly, it is quite difficult to check whether the computation is correct, since we'd need to compare it with what the code-gen in rustc does. I'm not sure, I got all the casts and bit moving right, especially when it comes to the OCaml side (which is currently still WIP).


charon-ml/src/TypesUtils.ml
```diff
@@ -169,6 +169,40 @@ let trait_instance_id_as_trait_impl (id : trait_instance_id) :
...
+  | I128 | U128 ->
+      true (* Only true if derived from anything representable in rust *)
+
+let max_of (int_ty : integer_type) =
+  (* FIXME: Not sure whether all of these values are even representable before being made into Z.t. *)
+  match int_ty with
+  | I8 -> Z.of_int 0x7FF
+  | I16 -> Z.of_int 0x7FFF
+  | I32 -> Z.of_int32 Int32.max_int
+  | Isize | I64 -> Z.of_int64 Int64.max_int
```
Nadrieril: This implicitly assumes a 64 bits target. I'm not sure how do deal with this well, maybe we can just document this assumption in the function documentation.

charon/src/ast/types.rs
```diff
@@ -333,20 +333,41 @@ pub struct VariantLayout {
     /// The offset of each field.
     #[drive(skip)]
     pub field_offsets: Vector<FieldId, ByteCount>,
+    /// Whether this variant is uninhabited. If `field_offsets` is empty
+    /// this could also mean that the variant is a ZST. Thus we need this flag
+    /// additionally.
```
```suggestion
    /// Whether the variant is uninhabited, i.e. has any valid possible value.
    /// Note that uninhabited types can have arbitrary layouts.
```

charon/src/ast/types.rs
```diff
@@ -333,20 +333,41 @@ pub struct VariantLayout {
...
+    #[drive(skip)]
+    pub uninhabited: bool,
+    /// The memory representation of the discriminant corresponding to this
+    /// variant. It must be of the same type as the corresponding [`DiscriminantLayout::tag_ty`].
+    ///
+    /// If it's `None`, then this variant is either the untagged variant (cf. [`TagEncoding::Niche::untagged_variant`])
+    /// containing the relevant niche or the single variant of a struct or uninhabited.
```
Nadrieril: Your comment was correct but I got confused reading it for a moment.
```suggestion
    /// If it's `None`, then this variant is either:
    /// - the untagged variant (cf. [`TagEncoding::Niche::untagged_variant`]) of a niched enum;
    /// - the single variant of a struct;
    /// - uninhabited.
```

charon/src/ast/types.rs
```diff
@@ -333,20 +333,41 @@ pub struct VariantLayout {
...
-        /// The representation type of the discriminant.
-        repr: IntegerTy,
-    },
-    Niche, // TODO: Add more useful information about niches here in the future.
+pub enum TagEncoding {
+    /// Represents the direct encoding of the discriminant as the tag via integer casts.
+    Direct,
+    /// Represents the encoding of the discriminant in the niche of variant `untagged_variant`.
+    Niche { untagged_variant: VariantId },
+}
```
Nadrieril, line 358: That feels redundant with the `tag` field of each variant. The only sensible way to read/write a discriminant needs to go through that `tag` field so I think this enum can be removed.

charon/src/ast/types.rs
```diff
@@ -333,20 +333,41 @@ pub struct VariantLayout {
...
+    /// If it's `None`, then this variant is either the untagged variant (cf. [`TagEncoding::Niche::untagged_variant`])
+    /// containing the relevant niche or the single variant of a struct or uninhabited.
+    #[drive(skip)]
+    pub tag: Option<ScalarValue>,
 }
 
-/// Layout of the discriminant with its offset and representation type.
-///
-/// Does not include information about the value range.
+/// Describes how the discriminant is related to the tag, i.e. its memory representation.
```
```suggestion
/// Describes how we represent the active enum variant in memory.
```

charon/src/ast/types_utils.rs
```diff
@@ -901,6 +913,66 @@ pub trait TyVisitable: Sized + AstVisitable {
     }
 }
 
+impl TypeDecl {
+    /// Computes the variant from the tag.
```
Nadrieril:
```suggestion
    /// Computes the variant from the tag (i.e. the in-memory bytes that represent the discriminant).
    /// Returns `None` for types that don't have a relevant discriminant (e.g. uninhabited types).
```

charon/tests/layout.rs
```diff
@@ -111,8 +111,78 @@ fn type_layout() -> anyhow::Result<()> {
...
+                let name = repr_name(&crate_data, &tdecl.item_meta.name);
+                let variants_from_kind = match &tdecl.kind {
+                    TypeDeclKind::Enum(variants) => variants,
+                    _ => unreachable!(),
+                };
+                for (var_id, variant) in layout.variant_layouts.iter_indexed() {
+                    let discr = variants_from_kind.get(var_id).unwrap().discriminant;
+
+                    // As discussed in https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20VariantId.3DDiscriminant.20when.20tag.20is.20niche.20encoded.3F/with/524384295
+                    // for niche-encoded tags, the variant id's should coincide with their discriminants.
```
Nadrieril: That's an implementation detail of today's rustc; rust does not guarantee this. So I'd prefer we don't test this, because if this ever breaks it won't be our fault and we don't even want to fix it.

charon/src/ast/types.rs
Nadrieril: Ah no, I forgot about uninhabited variants. That's ok then.

charon-ml/src/TypesUtils.ml
```diff
@@ -337,3 +371,47 @@ let generic_params_lengths (args : generic_params) : int * int * int * int =
...
+                else find_mapi f (i + 1) tl
+          in
+
+          match discr_layout.encoding with
+          | Direct -> begin
+              assert (discr_layout.tag_ty = tag.int_ty);
+              let discr =
+                if is_in_bounds discr_ty tag.value then
+                  Some { tag with int_ty = discr_ty }
+                else None
```
Nadrieril, line 403: Why this extra check?

charon/src/ast/types_utils.rs
```diff
@@ -901,6 +913,66 @@ pub trait TyVisitable: Sized + AstVisitable {
...
+                    None
+                } else {
+                    let discr_ty = variants.get(VariantId::ZERO)?.discriminant.get_integer_ty();
+
+                    match &discr_layout.encoding {
+                        TagEncoding::Direct => {
+                            assert_eq!(tag.get_integer_ty(), discr_layout.tag_ty);
+                            let discr = ScalarValue::from_bits(discr_ty, tag.to_bits());
+                            // Find corresponding variant.
+                            variants.iter_indexed().find_map(|(id, variant)| {
```
Nadrieril: We must use `variant_layouts` in both cases. The fact that discriminant values equal tag values is an implementation detail I don't want to rely on.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -424,6 +429,55 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                            .get(variant)
+                            .expect("Variant index out of bounds while translating tag")
+                            .discriminant
+                            .to_bits()
+                    };
+                    let discr_bits = discr.to_bits();
+                    if discr_bits == get_bits_of_variant(*untagged_variant) {
+                        None // This variant does not have a tag.
+                    } else {
+                        let discr_rel = discr_bits - get_bits_of_variant(*niche_variants.start());
```
Nadrieril: That's not [the formula used by cg_clif](https://github.com/rust-lang/rustc_codegen_cranelift/blob/8dea73acfd76a38173975ff1e60b7770af677768/src/discriminant.rs#L58-L70). Discriminants don't play into it; instead, it's only comparisons on variant ids. This function should take the current variant_id as input, then
```rust
let discr_rel = variant_index.as_u32() - niche_variants.start().as_u32();
let tag_bits = (discr_rel as u128).wrapping_add(niche_start);
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -424,6 +429,55 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                    let get_bits_of_variant = |variant: r_abi::VariantIdx| {
+                        let variant = translate_variant_id(variant);
+                        variants
+                            .get(variant)
+                            .expect("Variant index out of bounds while translating tag")
+                            .discriminant
+                            .to_bits()
+                    };
+                    let discr_bits = discr.to_bits();
+                    if discr_bits == get_bits_of_variant(*untagged_variant) {
```
Nadrieril: Please do the comparison on variant ids directly

charon/tests/layout.rs
firefighterduck: Funnily enough, there is a similar check in the code-gen_ssa in rustc. Nonetheless, I agree that we can remove this check.

Nadrieril: Interesting, I do wonder why; they might be relying on that for optimizations

charon-ml/src/TypesUtils.ml
firefighterduck: For some reason, in Rust the `ScalarValue::Isize` contains an i64. Should we also change that?

charon/tests/layout.rs
firefighterduck: For reference, it's here: https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_ssa/src/mir/operand.rs#L485

charon-ml/src/TypesUtils.ml
Nadrieril: `i64` is big enough to contain both a `i32` or a `i64` so that's ok. It's just that on 32 bit platforms the max value stored in it will be `i32::MAX`.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -424,6 +429,46 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+        ) -> Option<ScalarValue> {
+            match &encoding {
+                // The direct encoding is just a cast.
+                r_abi::TagEncoding::Direct => Some(ScalarValue::from_bits(tag_ty, discr.to_bits())),
+                r_abi::TagEncoding::Niche {
+                    untagged_variant,
+                    niche_variants,
+                    niche_start,
+                } => {
+                    if variant.index() as u32 == untagged_variant.as_u32() {
```
Nadrieril: Doesn't typed equality `variant == *untagged_variant` work?

firefighterduck: Ah yes, I simply forgot to update that part.

charon-ml/src/TypesUtils.ml
firefighterduck: Okay, I switched to using nativeint to model this more closely.

charon/src/ast/types_utils.rs
firefighterduck: Fixed. I kinda overlooked this yesterday.

Nadrieril:
Looks great, thank you for this!

PR #748, opened 2025-07-01
Switch tag computation to rustc query

### LINKED ISSUE #746: Use tag computation query from rustc
Charon currently computes the tag from the discriminant for each variant by itself. However, there is a query in rustc `tag_for_variant` which does exactly this in a canonical way. We currently cannot use it, since it had a bit of a buggy setup that was [only fixed](https://github.com/rust-lang/rust/pull/143096) in `nightly-2025-06-29` which doesn't work with the currently used hax version. As soon as hax can be used with this or any newer nightly, we should switch to using the query instead of doing the computation in Charon.

firefighterduck:
Fixes #746


Nadrieril:
Amazing, glad to remove all that subtle code

PR #755, opened 2025-07-04
Add a unique identifier to LLBC statements

sonmarcho:
This PR adds unique ids to the LLBC statements.

ci: use https://github.com/AeneasVerif/aeneas/pull/566


charon/src/ast/llbc_ast_utils.rs
```diff
@@ -62,6 +62,7 @@ impl Statement {
     pub fn new(span: Span, content: RawStatement) -> Self {
         Statement {
             span,
+            id: StatementId::ZERO,
```
Nadrieril: We don't care about the value of these ids, so we could use a global variable to make unique ids and remove the micro-pass:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Drive, DriveMut)]
struct StatementId(#[drive(skip)] usize);
impl StatementId {
  fn new() -> Self {
	static COUNTER: AtomicUsize = AtomicUsize::new(0);
	let id = COUNTER.fetch_add(1, Ordering::Relaxed);
    Self(id)
  }
}
```
This would make it possible to use these ids in charon's passes and generally avoid surprises.

sonmarcho: It is actually useful to keep the value of those ids low and thus refresh the counter for each function, because when deboguing (for instance on the Aeneas side) it is easier to distinguish small numbers (e.g., `8` from `42` rather than `118354` from `118392`). It is an issue I encountered when reading traces of the symbolic execution, where we need to generate fresh ids for, for instance, borrows, and the value of those ids doesn't matter *in theory*.
This being said, I actually recompute those ids on the Aeneas side because the pre-passes introduce new statements, so there is no strong requirement on my side (I actually hesitated to put those ids into an option).

sonmarcho: (feel free to update the code to remove the micro-pass, I really don't mind)

Nadrieril: That's a very fair point. The risk of creating new statements in a micro-pass and not updating them feels too high tho, I prefer the illegible ids for robustness.

charon/src/transform/mod.rs
```diff
@@ -6,6 +6,7 @@ pub mod duplicate_return;
 pub mod expand_associated_types;
 pub mod filter_invisible_trait_impls;
 pub mod filter_unreachable_blocks;
+pub mod generate_statement_ids;
```
Nadrieril: You had forgotten to add that pass to the list :D

PR #762, opened 2025-07-08
Supporting `dyn Trait`

ssyram:
Thanks to the hard and quick work by @Nadrieril , now most parts of this PR have been merged into the main branch, with only two non-urgent parts:

- The drop shim for array or tuples could be handled by some parts of [VTable Metadata pass: size, align & drop](https://github.com/AeneasVerif/charon/pull/762/commits/2ae73fb5c74a5ee954cdebf31dce7bfd657743ab) , this is yet to separate and integrate to the current mechanism.
- The vtable instances support for closures.

These will be further split into new PRs, but as they are not that urgent, it may take place in the coming year.

# Former Text (As History of What this PR Did)

Now the `dyn Trait` Support should be quite ready with a few TODOs.

## Features Implemented

- Supporting associated types, now the vtable structs & instances all additionally take additional generic types corresponding to and taken from `&dyn Trait`'s associated type restrictions. The basic support of the associated types are from #796.
- The vtable instances are now properly generated for the concrete impls, i.e., those with explicit `impl Trait for Ty` environments.
- The vtable method calls are propertly generated that, calling methods of `&dyn Trait` will be translated to finding the corresponding function pointer from the vtable (fetched by `ptr_metadata` operation) and then call the function pointer.
- A new pass is added to compute the correct vtable instance metadata when possible, i.e., the `size`, `align` and, the most complex case, the `drop` fields. Naturally, the content will not be computable if the stuff is behind a generics, this is inevitable and the fields will remain unknown. But everything should be explicit in the monomorphized case in the future. Specifically, the `size` and `align` subjects to whether the layout of the type is properly computable. The `drop` field, as the most complex case, now also handles the cases when the concrete impl types are `Array`s & `Tuple`s -- they should be the only composite cases.
- The vtable instances support for closures are now added. All tests passed. But there are identified issues, see ssyram/charon#20 for the closure itself. But it should work properly in most of the cases -- as `dyn FnOnce` is really strange, and rare.

## TODOs

- [ ] Monomorphization support for `dyn Trait`.
- [x] Vtable instances for closures.

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -467,6 +467,9 @@ pub struct TypeDecl {
     /// Is `None` for foreign type declarations.
     #[drive(skip)]
     pub repr: Option<ReprOptions>,
+    /// The drop implementation for this type, if any.
+    /// This is `Some` if and only if the given type has a drop implementation.
+    pub drop_glue: Option<TraitImplRef>,
 }
 
 generate_index_type!(VariantId, "Variant");
--- a/charon/src/bin/charon-driver/translate/translate_closures.rs
+++ b/charon/src/bin/charon-driver/translate/translate_closures.rs
@@ -282,7 +282,7 @@ impl ItemTransCtx<'_, '_> {
 
     /// Given an item that is a closure, generate the signature of the
     /// `call_once`/`call_mut`/`call` method (depending on `target_kind`).
-    fn translate_closure_method_sig(
+    pub fn translate_closure_method_sig(
         &mut self,
         def: &hax::FullDef,
         span: Span,
--- a/charon/src/bin/charon-driver/translate/translate_crate.rs
+++ b/charon/src/bin/charon-driver/translate/translate_crate.rs
@@ -69,7 +69,7 @@ pub enum TransItemSourceKind {
     /// Shim function to store a method in a vtable; give a method with `self: Ptr<Self>` argument,
     /// this takes a `Ptr<dyn Trait>` and forwards to the method. The `DefId` refers to the method
     /// implementation.
-    VTableMethod,
+    VTableMethod(TraitImplSource),
 }
 
 /// The kind of a [`TransItemSourceKind::TraitImpl`].
@@ -282,7 +282,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
                     | ClosureAsFnCast
                     | DropGlueMethod
                     | VTableInstanceInitializer(..)
-                    | VTableMethod => ItemId::Fun(self.translated.fun_decls.reserve_slot()),
+                    | VTableMethod(..) => ItemId::Fun(self.translated.fun_decls.reserve_slot()),
                     InherentImpl | Module => return None,
                 };
                 // Add the id to the queue of declarations to translate
--- a/charon/src/bin/charon-driver/translate/translate_crate.rs
+++ b/charon/src/bin/charon-driver/translate/translate_crate.rs
@@ -425,7 +425,6 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
     }
 
     /// Register this item without enqueueing it for translation.
-    #[expect(dead_code)]
     pub(crate) fn register_item_no_enqueue<T: TryFrom<ItemId>>(
         &mut self,
         span: Span,
@@ -486,7 +485,7 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
         span: Span,
         item: &hax::ItemRef,
     ) -> Result<TypeDeclRef, Error> {
-        match self.recognize_builtin_type(item)? {
+        match self.recognize_builtin_type(span, item)? {
             Some(id) => {
                 let generics =
                     self.translate_generic_args(span, &item.generic_args, &item.impl_exprs)?;
--- a/charon/src/bin/charon-driver/translate/translate_meta.rs
+++ b/charon/src/bin/charon-driver/translate/translate_meta.rs
@@ -362,7 +362,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
                 name.name
                     .push(PathElem::Ident("{vtable}".into(), Disambiguator::ZERO));
             }
-            TransItemSourceKind::VTableMethod => {
+            TransItemSourceKind::VTableMethod(..) => {
                 name.name.push(PathElem::Ident(
                     "{vtable_method}".into(),
                     Disambiguator::ZERO,
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -16,10 +16,6 @@ fn dummy_public_attr_info() -> AttrInfo {
     }
 }
 
-fn usize_ty() -> Ty {
-    Ty::new(TyKind::Literal(LiteralTy::UInt(UIntTy::Usize)))
-}
-
 /// Takes a `T` valid in the context of a trait ref and transforms it into a `T` valid in the
 /// context of its vtable definition, i.e. no longer mentions the `Self` type or `Self` clause. If
 /// `new_self` is `Some`, we replace any mention of the `Self` type with it; otherwise we panic if
@@ -294,9 +290,9 @@ impl ItemTransCtx<'_, '_> {
 
         // Add the basic fields.
         // Field: `size: usize`
-        mk_field("size".into(), usize_ty());
+        mk_field("size".into(), Ty::mk_usize());
         // Field: `align: usize`
-        mk_field("align".into(), usize_ty());
+        mk_field("align".into(), Ty::mk_usize());
         // Field: `drop: fn(*mut Self)` -- `Self` is just a placeholder, will be dynified below.
         mk_field("drop".into(), {
             let self_ty = TyKind::TypeVar(DeBruijnVar::new_at_zero(TypeVarId::ZERO)).into_ty();
@@ -449,6 +445,7 @@ impl ItemTransCtx<'_, '_> {
             // A vtable struct is always sized
             ptr_metadata: PtrMetadata::None,
             repr: None,
+            drop_glue: None,
         })
     }
 }
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -230,6 +230,18 @@ impl ItemTransCtx<'_, '_> {
         else {
             return Ok(());
         };
+        // The `FnOnce::call_once` is not compatible with the dyn-compatibility as described in:
+        // https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility
+        // But the Rustc offers a special backdoor for this and allows this method to be `vtable_safe`.
+        // The only way to call such a function seems to be with `Box<dyn FnOnce>`.
+        // The implementation of `Box<dyn FnOnce>::call_once` is a special case and should translate specially with some special magic.
+        // TODO(ssyram): in the future we may implement the function with type `fn call_once(self: Box<Self>, Args) -> Output`
+        //               This function is the vtable-specific version of the `call_once` method.
+        //               A very special case that the type signature differs from the original signature.
+        // Anyway, if we allow this to happen, there will be a severe breakage in that we can perform concretization on `dyn FnOnce`,
+        // This will make the life in backend tools, especially Eurydice much harder.
+        // Likewise, we skip translation of the `call_once` method below in the instance.
+        if trait_def.lang_item == Some("fn_once".into()) { return Ok(()); }
 
         let item_name = self.t_ctx.translate_trait_item_name(item_def_id)?;
         // It's ok to translate the method signature in the context of the trait because
@@ -477,22 +489,48 @@ impl ItemTransCtx<'_, '_> {
     fn get_vtable_instance_info<'a>(
         &mut self,
         span: Span,
-        impl_def: &'a hax::FullDef,
+        def: &'a hax::FullDef,
         impl_kind: &TraitImplSource,
     ) -> Result<(TraitImplRef, TraitDeclRef, TypeDeclRef), Error> {
-        let implemented_trait = match impl_def.kind() {
-            hax::FullDefKind::TraitImpl { trait_pred, .. } => &trait_pred.trait_ref,
-            _ => unreachable!(),
+        let (implemented_trait, impl_ref) = match def.kind() {
+            hax::FullDefKind::TraitImpl { trait_pred, .. } => (
+                &trait_pred.trait_ref,
+                self.translate_item(span, def.this(), TransItemSourceKind::TraitImpl(*impl_kind))?,
+            ),
+            hax::FullDefKind::Closure {
+                fn_once_impl,
+                fn_mut_impl,
+                fn_impl,
+                args,
+                ..
+            } => {
+                // For closures, get the trait implementation based on the closure kind
+                let closure_trait_impl = match impl_kind {
+                    TraitImplSource::Closure(ClosureKind::FnOnce) => fn_once_impl,
+                    TraitImplSource::Closure(ClosureKind::FnMut) => {
+                        fn_mut_impl.as_ref().expect("FnMut impl should exist")
+                    }
+                    TraitImplSource::Closure(ClosureKind::Fn) => {
+                        fn_impl.as_ref().expect("Fn impl should exist")
+                    }
+                    _ => unreachable!("Expected closure trait impl source"),
+                };
+                let target_kind = match impl_kind {
+                    TraitImplSource::Closure(kind) => *kind,
+                    _ => unreachable!(),
+                };
+                (
+                    &closure_trait_impl.trait_pred.trait_ref,
+                    self.translate_closure_bound_impl_ref(span, args, target_kind)?
+                        .erase(),
+                )
+            }
+            _ => panic!("Unknown type of impl full def: {:?}", def.kind()),
         };
         let vtable_struct_ref = self
             .translate_vtable_struct_ref(span, implemented_trait)?
             .expect("trait should be dyn-compatible");
         let implemented_trait = self.translate_trait_decl_ref(span, implemented_trait)?;
-        let impl_ref = self.translate_item(
-            span,
-            impl_def.this(),
-            TransItemSourceKind::TraitImpl(*impl_kind),
-        )?;
         Ok((impl_ref, implemented_trait, vtable_struct_ref))
     }
 
@@ -567,7 +605,7 @@ impl ItemTransCtx<'_, '_> {
                 // generics -- the lifetime binder will be added as `Erased` in `translate_fn_ptr`.
                 let item_ref = impl_def.this().with_def_id(self.hax_state(), item_def_id);
                 let shim_ref = self
-                    .translate_fn_ptr(span, &item_ref, TransItemSourceKind::VTableMethod)?
+                    .translate_fn_ptr(span, &item_ref, TransItemSourceKind::VTableMethod(TraitImplSource::Normal))?
                     .erase();
                 ConstantExprKind::FnPtr(shim_ref)
             }
@@ -591,12 +629,30 @@ impl ItemTransCtx<'_, '_> {
         impl_def: &hax::FullDef,
         mut mk_field: impl FnMut(ConstantExprKind),
     ) -> Result<(), Error> {
-        let hax::FullDefKind::TraitImpl {
-            implied_impl_exprs, ..
-        } = impl_def.kind()
-        else {
-            unreachable!()
+        let implied_impl_exprs = match impl_def.kind() {
+            hax::FullDefKind::TraitImpl {
+                implied_impl_exprs, ..
+            } => implied_impl_exprs,
+            hax::FullDefKind::Closure { fn_once_impl, fn_mut_impl, fn_impl, .. } => {
+                match trait_def.lang_item {
+                    Some("fn") | Some("r#fn") => &fn_impl.as_ref().unwrap().implied_impl_exprs,
+                    Some("fn_once") => &fn_once_impl.as_ref().implied_impl_exprs,
+                    Some("fn_mut") => &fn_mut_impl.as_ref().unwrap().implied_impl_exprs,
+                    _ => unreachable!(),
+                }
+            },
+            kind => raise_error!(
+                self,
+                span,
+                "TODO: Unknown type of impl full def: {kind:?}"
+            ),
         };
+        // let hax::FullDefKind::TraitImpl {
+        //     implied_impl_exprs, ..
+        // } = impl_def.kind()
+        // else {
+        //     unreachable!()
+        // };
         let hax::FullDefKind::Trait {
             implied_predicates, ..
         } = trait_def.kind()
@@ -634,8 +690,107 @@ impl ItemTransCtx<'_, '_> {
                     });
                     ConstantExprKind::Ref(global)
                 }
-                // TODO(dyn): builtin impls
-                _ => ConstantExprKind::Opaque("missing supertrait vtable".into()),
+                hax::ImplExprAtom::Builtin { trait_data, .. } => {
+                    // Handle built-in implementations, including closures
+                    let tref = &impl_expr.r#trait;
+                    let hax_state = self.hax_state_with_id();
+                    let trait_def = self.hax_def(
+                        &tref.hax_skip_binder_ref().erase(&hax_state),
+                    )?;
+                    
+                    trace!("Processing builtin impl for trait {:?}, lang_item: {:?}, trait_data: {:?}", 
+                           trait_def.def_id(), trait_def.lang_item, trait_data);
+                           
+                    let closure_kind =
+                        trait_def.lang_item.as_deref().and_then(|lang| {
+                            match lang {
+                                "fn_once" => Some(ClosureKind::FnOnce),
+                                "fn_mut" => Some(ClosureKind::FnMut),
+                                "fn" | "r#fn" => Some(ClosureKind::Fn),
+                                _ => None,
+                            }
+                        });
+
+                    // Check if this is a closure trait implementation
+                    if let Some(closure_kind) = closure_kind
+                        && let Some(hax::GenericArg::Type(closure_ty)) =
+                            impl_expr
+                                .r#trait
+                                .hax_skip_binder_ref()
+                                .generic_args
+                                .first()
+                        && let hax::TyKind::Closure(closure_args) =
+                            closure_ty.kind()
+                    {
+                        // Register the closure vtable instance
+                        let vtable_instance_ref: GlobalDeclRef = self.translate_item(
+                            span,
+                            &closure_args.item,
+                            TransItemSourceKind::VTableInstance(
+                                TraitImplSource::Closure(closure_kind),
+                            ),
+                        )?;
+                        let global = Box::new(ConstantExpr {
+                            kind: ConstantExprKind::Global(vtable_instance_ref),
+                            ty: fn_ptr_ty,
+                        });
+                        ConstantExprKind::Ref(global)
+                    } else if self.translate_vtable_struct_ref(span, impl_expr.r#trait.hax_skip_binder_ref())?.is_some()
+                    {
+                        // For other builtin traits that are dyn-compatible, try to create a vtable instance
+                        trace!("Handling dyn-compatible builtin trait: {:?}", trait_def.lang_item);
+                        
+                        // TODO(ssyram): for now, we don't know how to translate other kinds of built-in traits, just take their names
+                        // The challenge is that we need an impl_ref, but builtin traits don't have concrete impls
+                        // Let's try using the trait reference itself as the impl reference
+                        // A simple case would be that they are all marker traits like `core::marker::MetaSized`
+                        let trait_ref = impl_expr.r#trait.hax_skip_binder_ref();
+                        let mut vtable_instance_ref: GlobalDeclRef = self.translate_item_no_enqueue(
+                            span,
+                            trait_ref,
+                            TransItemSourceKind::VTableInstance(
+                                TraitImplSource::Normal,  // Builtin traits are normal impls
+                            ),
+                        )?;
+                        // Remove the first `Self` argument
+                        vtable_instance_ref.generics.types.remove_and_shift_ids(TypeVarId::ZERO);
+                        let global = Box::new(ConstantExpr {
+                            kind: ConstantExprKind::Global(vtable_instance_ref),
+                            ty: fn_ptr_ty,
+                        });
+                        ConstantExprKind::Ref(global)
+                    } else {
+                        // For non-dyn-compatible builtin traits, we don't need vtable instances
+                        trace!("Non-dyn-compatible builtin trait: {:?}", trait_def.lang_item);
+                        ConstantExprKind::Opaque(format!("non-dyn-compatible builtin trait {:?}", 
+                                                        trait_def.lang_item.as_deref().unwrap_or("unknown")).into())
+                    }
+                }
+                hax::ImplExprAtom::LocalBound { .. } => {
+                    // No need to register anything here as there is no concrete impl
+                    // This results in that: the vtable instance in generic case might not exist
+                    // But this case should not happen in the monomorphized case
+                    if self.monomorphize() {
+                        raise_error!(
+                            self,
+                            span,
+                            "Unexpected `LocalBound` in monomorphized context"
+                        )
+                    } else {
+                        ConstantExprKind::Opaque("generic supertrait vtable".into())
+                    }
+                }
+                hax::ImplExprAtom::Dyn | hax::ImplExprAtom::Error(..) => {
+                    // No need to register anything for these cases
+                    ConstantExprKind::Opaque("dyn or error supertrait vtable".into())
+                }
+                hax::ImplExprAtom::SelfImpl { .. } => {
+                    raise_error!(
+                        self,
+                        span,
+                        "`SelfImpl` should not appear in vtable construction"
+                    )
+                }
             };
             mk_field(kind);
         }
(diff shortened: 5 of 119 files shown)


charon/src/ast/types.rs
```diff
@@ -902,3 +896,37 @@ pub struct FunSig {
...
+    DynStar,
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub struct ExistentialTraitRef {
+    /// This field corresponds to all required info in Rustc
+    pub trait_ref: TraitDeclRef,
+    // the following field(s) is/are addional information in Charon
+    /// The trait vtable structure, referring to a (potentially generic) ADT
+    pub vtable_typ_id: TypeDeclId,
```
N1ark: if this is potentially generic, wouldn't it be a `TypeDeclRef`?

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -42,6 +42,45 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+            .collect::<Vec<String>>()
+            .join(" + ");
+        let str = format!(
+            "dyn multiple traits is not supported as per Rustc 1.90.0. I.e., (dyn {all}).
+            Consider the workaround by defining:
+            `trait Multiple : {all} {{}}`,
+            `impl <T : {all}> Multiple for T {{}}`
+            using `dyn Multiple` instead.",
+        );
+        raise_error!(self, span, "{}", str);
```
N1ark: isn't this checked for by rustc anyways? :o

charon/src/bin/charon-driver/translate/translate_dynamics.rs
```diff
@@ -0,0 +1,235 @@
...
+use itertools::Itertools;
+use tracing::field;
+
+/// Type: `()`
+fn unit_ty() -> Ty {
+    Ty::new(TyKind::Adt(TypeDeclRef {
+        id: TypeId::Tuple,
+        generics: Box::new(GenericArgs::empty()),
+    }))
+}
```
N1ark: You can use `Ty::mk_unit()`!

charon/src/bin/charon-driver/translate/translate_dynamics.rs
```diff
@@ -0,0 +1,235 @@
...
+fn unit_raw_ptr(is_mut: bool) -> Ty {
+    Ty::new(TyKind::RawPtr(
+        unit_ty(),
+        if is_mut {
+            RefKind::Mut
+        } else {
+            RefKind::Shared
+        },
+    ))
+}
```
N1ark: You can use `RefKind::mutable(bool)` to avoid the if-else

charon/src/bin/charon-driver/translate/translate_dynamics.rs
```diff
@@ -0,0 +1,235 @@
...
+
+fn dummy_public_attr_info() -> AttrInfo {
+    AttrInfo {
+        attributes: Vec::new(),
+        inline: None,
+        rename: None,
+        public: true,
+    }
+}
```
N1ark: I think `AttrInfo::default()` would do this :) maybe with `public: false` but I don't think it matters here

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Yes, but I'm not sure if this feature will get back in the future or not. Just a robustness check.

charon/src/bin/charon-driver/translate/translate_dynamics.rs
ssyram: I think having explicit public should help in Eurydice... At least we don't have to treat it specially in a sub-case. Also, having a semantically correct stuff often helps.

charon/src/bin/charon-driver/translate/translate_dynamics.rs
ssyram: Thanks, will fix!

charon/src/bin/charon-driver/translate/translate_dynamics.rs
ssyram: Thanks! Will fix!

charon/src/ast/types.rs
ssyram: Yes, you're right. Will fix!

charon/src/bin/charon-driver/translate/translate_dynamics.rs
Nadrieril: You can still write
```rust
AttrInfo {
    public: true,
    ..AttrInfo::default()
}
```

charon/src/bin/charon-driver/translate/translate_dynamics.rs
Nadrieril, line 1: Would you mind renaming to `translate_trait_objects.rs`? "dynamics" is not a known technical term in rust.

charon/src/ast/types.rs
Nadrieril: Let's not store that here and store it in the `TraitDecl` instead

charon/src/ast/types.rs
```diff
@@ -902,3 +896,37 @@ pub struct FunSig {
...
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub enum DynKind {
+    Dyn,
+    DynStar,
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub struct ExistentialTraitRef {
+    /// This field corresponds to all required info in Rustc
+    pub trait_ref: TraitDeclRef,
```
Nadrieril: That trait ref is missing a `Self` parameter, which makes it invalid. This will cause issue throughout charon and for consumers also.

I see a few options:
1. Store the `TraitDeclId` and `GenericArgs` separately; everything that handles generics will need to add a special case for this;
2. Create a new `TyKind::ExistentialPlaceholder` type, so that we can supply that `Self` argument and the `TraitDeclRef` becomes valid.
3. Make a `Binder<...>` that binds the existential variable.

I think option 2 is the easiest right now. What I had suggested you do was 1. for cases where the trait has no generics, but since you tried the gneeral case let's go for 2.

charon/src/ast/types.rs
```diff
@@ -902,3 +896,37 @@ pub struct FunSig {
...
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub enum TyTerm {
+    Ty(Ty),
+    Const(ConstantExpr),
+}
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub struct ExistentialProjection {
+    pub trait_ref: TraitDeclRef,
+    pub term: TyTerm,
+}
```
Nadrieril: Hax trait resolution cannot resolve these trait refs because they refer to the `dyn` clause itself. One solution is to assert that the `trait_ref` is indeed the same, and store only the `TyTerm`.

charon/src/ast/types.rs
```diff
@@ -902,3 +896,37 @@ pub struct FunSig {
...
+
+/// A predicate of the form `exists<T> where T: Trait`.
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub enum ExistentialPredicate {
+    Trait(ExistentialTraitRef),
+    Projection(ExistentialProjection),
+    AutoTrait(TraitDeclId),
```
Nadrieril: Let's not make this a separate variant, we can handle it as a `Trait`.

charon/src/ast/types.rs
```diff
@@ -902,3 +896,37 @@ pub struct FunSig {
     pub inputs: Vec<Ty>,
     pub output: Ty,
 }
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+pub enum DynKind {
+    Dyn,
+    DynStar,
+}
```
Nadrieril: We will not support `*dyn` until the feature makes progress on the rustc side. Please remove this enum and support only `Dyn`.

charon/src/ast/gast.rs
```diff
@@ -127,6 +127,11 @@ pub enum ItemKind {
         #[drive(skip)]
         reuses_default: bool,
     },
+    /// This is a virtual table struct for a trait
+    VTable {
+        /// which trait is this VTable for
+        trait_id: TraitDeclId,
```
Nadrieril: Please make this a `TraitDeclRef` so that the generics work nicely.

charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -55,6 +55,13 @@ pub enum TransItemSourceKind {
     DropGlueImpl,
     /// The `drop` method for the impl above.
     DropGlueMethod,
+    /// The virtual table struct (ADT), associated DefId should refer to the trait-decl itself.
+    VTable,
+    /// The virtual table global instance for a specific type, corresponding to TraitRef.
+    /// The associated DefId is essentially meaningless
+    VTableInstance(hax::TraitRef),
+    /// The init body of the `VTableInstance`, as per the practice of `Global`
+    VTableInstanceBody(hax::TraitRef),
```
Nadrieril: Please use the `DefId` of the trait impl as the corresponding `DefId`, which removes the need for the `TraitRef` payload. Not sure how to handle built-in vtables though.

charon/src/ast/types.rs
Nadrieril: Well I guess it's easy to support; we can keep it.

charon/src/ast/types.rs
```diff
@@ -788,7 +782,7 @@ pub enum TyKind {
     /// Into<u64>`. The predicate must quantify over a single type and no any regions or constants.
     ///
     /// TODO: we don't translate this properly yet.
-    DynTrait(ExistentialPredicate),
+    DynTrait(Vec<RegionBinder<ExistentialPredicate>>, Region, DynKind),
```
Nadrieril: Could you add a comment documenting the fact that only the first predicate may have methods?

charon/src/ast/types.rs
Nadrieril: With option 2/3 you could also remove the `Region` parameter of `TyKind::Dyn` and replace that with a `TypeOutlives` predicate (I prefer to reuse the predicate kinds we already have).

Nadrieril:
Haven't reviewed everything but this will require a change in how you represent the existential predicates.

charon/src/ast/types.rs
ssyram: So every trait has its own vtable and the vtable is not generated as we encounter `dyn Trait`?

charon/src/bin/charon-driver/translate/translate_dynamics.rs
ssyram: Sure!

charon/src/ast/types.rs
ssyram: Well, I'm only thinking about adding a tag on the AST node... This is straightforward and does not seem to cause extra-trouble?

charon/src/ast/types.rs
ssyram: Sure!

charon/src/bin/charon-driver/translate/translate_dynamics.rs
ssyram: Thanks, fixed.

charon/src/ast/types.rs
ssyram: OK, I will store everything in the `ExistentialTraitRef`, this is more clean.

charon/src/ast/types.rs
ssyram: Thanks, taken 2. But how to define the `TypeOutlives`? What should be the `Ty` parameter?

charon/src/ast/types.rs
ssyram: Will they be the same? The existential projection should have ref that refers to the assoc items inside the trait?

ssyram: Oh, then it is even not a trait ref...

charon/src/ast/gast.rs
ssyram: OK.

charon/src/ast/types.rs
ssyram: If we are to store everything inside a predicate, it would be strange to still use a vector?

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: I'll try to figure out how to get the TraitRef out from the trait impl.

ssyram:
Thanks for all the suggestions from both of you! I'm still not sure about how to refractor the Predicate form. So I will leave it to the discussion in Zulip. Let me fetch & implement the vtable struct & instances first.

charon/src/ast/types.rs
ssyram: Still not sure how, let us discuss in Zulip.

charon/src/ast/types.rs
ssyram: So in the translated `TraitRef`, I should additionally append the `TyKind::ExistentialPlaceholder` at the head of `types` in `GenericArgs`?

charon/src/ast/types.rs
ssyram: Tried to fix with the projection-item name and the args. Please have a look.

charon/src/ast/types.rs
Nadrieril: Good point, but we can do something nice. We can make it so that the vtable is only translated when used by doing `register_id_no_enqueue` in the trait decl (this will give you the `TypeDeclId` but not schedule its translation), and calling `register_and_enqueue_id` when we actually encounter a `dyn Trait`.

That way, all (dyn-compatible) traits will have a `vtable: TypeDeclId`, but that id will only correspond to an item if the vtable is actually used.

charon/src/ast/types.rs
Nadrieril: > Thanks, taken 2. But how to define the `TypeOutlives`? What should be the `Ty` parameter?

It would be the `ExistentialPlaceholder`.

> So in the translated `TraitRef`, I should additionally append the `TyKind::ExistentialPlaceholder` at the head of `types` in `GenericArgs`?

Precisely!

charon/src/ast/types.rs
Nadrieril: Yeah, this was easier than I expected, we can keep it

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: I'm confused, don't you want the opposite?

charon/src/ast/types.rs
Nadrieril: There can still be several trait predicates, no?

Nadrieril: Also please document that it's the first one that has methods (to avoid having to iterate through all of them all the time), this is guaranteed by what rustc gives us.

charon/src/ast/types.rs
```diff
@@ -797,6 +797,8 @@ pub enum TyKind {
     /// variables (those that could appear in a function pointer type like `for<'a> fn(&'a u32)`),
     /// we need to bind them here.
     FnDef(RegionBinder<FnPtr>),
+    /// A placeholder for existential predicates -- the `Self` is erased in such predicates.
```
```suggestion
    /// Fake type used in `dyn Trait` predicates: such a predicate has the `Self` type
    /// existentially quantified. In order for generics to stay consistent, we use this placeholder to
    /// represent the existentially quantified type.
```

charon/src/ast/types.rs
```diff
@@ -917,13 +919,18 @@ pub enum TyTerm {
...
 #[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
 pub struct ExistentialProjection {
-    pub trait_ref: TraitDeclRef,
+    pub trait_item : TraitItemName,
+    pub generics : Box<GenericArgs>,
     pub term: TyTerm,
 }
 
 /// A predicate of the form `exists<T> where T: Trait`.
+/// This predicate itself contains all the predicate info for the stuff
```
Nadrieril: "for the stuff" isn't very precise :)

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -112,18 +110,20 @@ impl ItemTransCtx<'_, '_> {
...
-        let trait_decl_ref =
-            self.translate_trait_decl_ref_from_ex_trait_ref(span, &proj.def_id, &proj.args)?;
-        // self.translate_trait_decl_ref(span, item);
+        let args = self.translate_ex_generics(span, &proj.args)?;
+        let name = self.t_ctx.translate_trait_item_name(&proj.def_id)?;
         let term = match &proj.term {
-            hax_frontend_exporter::Term::Ty(ty) => TyTerm::Ty(self.translate_ty(span, ty)?),
+            hax_frontend_exporter::Term::Ty(ty) => {
+                TyTerm::Ty(self.translate_ty(span, ty)?)
+            }
```
Nadrieril: Please add an assertion that the trait_ref is the one we expect

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -203,11 +203,12 @@ impl ItemTransCtx<'_, '_> {
             };
             fields.push(field);
         }
+        let trait_def_ref = self.translate_trait_decl_ref_from_ex_trait_ref(Span::dummy(), &hax_ex_tref.def_id, &hax_ex_tref.args)?;
```
Nadrieril: The handling of generics isn't correct here. The way you've done it, we'll generate a new item for each generics e.g. `dyn Trait<u32>` will have a different item than `dyn Trait<u64>`. Instead we want the vtable struct to have the same generics as the trait. The way to do it is to remove the payload of `TransItemSourceKind::VTable`, and here we know that the generics are the same so we can use `generics.identity()` as the generics for the `TraitDeclRef`.

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: What might be missing is that when casting something into `dyn` we'll need the right `ImplExpr` to know what impl is being used. This will have to be added on the hax side.

charon/src/ast/types.rs
Nadrieril: Well actually, `DynStar` [just got removed from rustc](https://github.com/rust-lang/rust/pull/143036) so we should remove it also.

ssyram: OK.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: OK, should be done now.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: There's no `trait_decl_ref` now in Projection.

charon/src/ast/types.rs
ssyram: OK, I will try to be more precise when we determined what to do...

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: Seems more complex than expected. I was thinking about fetching info from Rustc like in the `vtable_entries` query impl.

charon/src/ast/types.rs
ssyram: Added.

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: You can't avoid doing trait resolution at some point or other, because when casting into `dyn` you have to figure out the trait impl to point to. I guess `vtable_entries` does a bit of trait resolution (it uses `expect_resolve_for_vtable`) but that wouldn't work in a generic context, e.g.
```rust
fn cast_dyn<T: Debug>(x: &T) -> &dyn Debug { x }
```
In fact, could you add this as a test?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: Yes, that's exactly why you should check that the `trait_ref` you're ignoring is the one you expect. Because someone may write `dyn Debug + Pointee<Metadata=()>` where the projection applies to some trait other than the main one (I don't know if this is allowed today but this could be).

charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -1764,6 +1771,8 @@ impl<C: AstFormatter> FmtWithCtx<C> for Ty {
                 };
                 write!(f, "{value}",)
             }
+            // To make sure this is not definable as ADT or so
+            TyKind::ExistentialPlaceholder => write!(f, "$Exist$"),
```
Nadrieril: `$` looks odd, you could name it `_` I think that'd be clear enough

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -216,10 +255,21 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
-            hax::TyKind::Dynamic(_existential_preds, _region, _) => {
-                // TODO: we don't translate the predicates yet because our machinery can't handle
-                // it.
-                TyKind::DynTrait(ExistentialPredicate)
+            hax::TyKind::Dynamic(preds, region, ..) => {
+                let region = self.translate_region(span, region)?;
+                // This is a robustness check: the current version of Rustc
+                // accepts at most one existential trait ref in a dyn object.
+                // But things may change in the future
+                self.check_at_most_one_existential_trait_ref(span, preds)?;
```
Nadrieril: Note that rustc guarantees that the only method-bearing trait ref is the first one (and a lot of code in rustc relies on that), let's use that to simplify the check.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -42,6 +42,45 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+            return Ok(());
+        }
+        // prepare to print
+        // all: Trait1 + Trait 2 + ...
+        let all = all_ex_trait_ref
+            .iter()
+            .map(|x| {
+                self.t_ctx
+                    .tcx
+                    .def_path_str(x.def_id.as_rust_def_id().unwrap())
```
Nadrieril: We have our own printing system, please translate via charon before doing this check and then use the `fmt_with_ctx` machinery.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -42,6 +42,45 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+            })
+            .collect::<Vec<String>>()
+            .join(" + ");
```
Nadrieril: `Itertools` has a helper for this:
```suggestion
            .format(" + ");
```

charon/src/ast/types.rs
ssyram: So we keep the current structure?

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: Sure, let's test when the infrastructure is about to ready.

charon/src/pretty/fmt_with_ctx.rs
ssyram: Use `_` would make it definable by users? How about "{Exist}"?

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Then we will need no check? I guess the current one is simple enough? I just want to make sure if things changed in the future, we can get a proper panic...

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Will try.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Referring to Pointee like that is not possible. But it is still possible that the item refers to the associated type from super trait. Is there a way to refer to that? Using the item name itself seems not sufficient to resolve this. Shall we change the mechanism?

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Tried, would you like to have a look? I'm not sure if the current version is correct.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: Oh true, could be a supertraits type that's a good example. I propose we error if that happens for now.

charon/src/pretty/fmt_with_ctx.rs
Nadrieril: `_` is a reserved name, you can't name a type `_`

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: I mean sure we can keep the check as is

Nadrieril:
> Maybe we can focus on the implementation for now?

The tests will help me understand what the implementation is doing, I can't just be running the code in my head.

charon/src/bin/charon-driver/translate/translate_generics.rs
```diff
@@ -489,6 +489,8 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
     }
 
     /// Translate the generics and predicates of this item and its parents.
+    /// This function makes the generic environment in `self`, so that the generics and parent clauses
+    /// are available for the analysis afterwards.
```
Nadrieril:
```suggestion
    /// This adds generic parameters and predicates to the current environment (as a binder in `self.binding_levels`).
    /// This is necessary to translate types that depend on these generics (such as `Ty` and `TraitRef`).
    /// The constructed `GenericParams` can be recovered at the end using `self.into_generics()` and stored in the translated item.
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -64,11 +66,7 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
         // all: Trait1 + Trait 2 + ...
         let all = all_ex_trait_ref
             .iter()
-            .map(|x| {
-                self.t_ctx
-                    .tcx
-                    .def_path_str(x.def_id.as_rust_def_id().unwrap())
-            })
+            .map(|x| x.with_ctx(&self.into_fmt()).to_string())
```
Nadrieril: nit: you don't need the `to_string()` I think, and could you move `let fmt = &self.into_fmt()` outside the loop?

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Looks good :+1:

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: Please add a `known-failure` test for that case please

charon/src/ast/types.rs
Nadrieril: Let's replace all the `AutoTrait`s with `Trait`, and document what the projection applies to; after that yeah that's good

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Another good point is that, Rust seems to not allow such specification to have shadowing names, hence the specification by name at the point can uniquely point to an assoc type. Or, it may be just that I'm not familiar with the syntax... At least I don't know how to express:
```Rust
trait Base { type Assoc; }
trait Trait { type Assoc; }
fn main() {
  let x : &dyn Trait<Assoc=(), ???> = ...;
}
```
where `???` denotes the specification of `Base::Assoc`. The Rustc error hints that "consider renaming".

charon/src/pretty/fmt_with_ctx.rs
ssyram: OK.

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: sure, fixed.

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: Added.

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Check ExTraitRef being the first done.

charon/src/ast/types.rs
ssyram: No such text any more...

charon/src/ast/types.rs
ssyram: So you mean we incorporate the AutoTraits in Trait? An ExistentialPredicate should then be: 1. a "Principal" `ExistentialTraitRef` (being a `TraitRef`), 2. a list of projections and 3. a list of auto trait ids? From enum to a struct?

charon/src/ast/types.rs
ssyram: This plan has been implemented.

charon/src/ast/types.rs
Nadrieril: That was split off in https://github.com/AeneasVerif/charon/pull/768.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: Yep I think you're right, rust can't express that today.

ssyram: The current solution is already comprehensive anyway. Thanks for the discussion!

charon/src/bin/charon-driver/translate/translate_closures.rs
```diff
@@ -681,6 +681,8 @@ impl ItemTransCtx<'_, '_> {
             consts: vec![],
             types,
             methods: vec![(call_fn_name, call_fn_binder)],
+            // There cannot be any reference to this closure as `dyn`
+            vtable_instance: None,
```
N1ark: What about this?
```rust
let closure = || {};
let closure_dyn: &dyn Fn() = &closure;
```

ssyram: Oh, you're right. Will fix this.

ssyram: Fixed, thanks!

charon/src/bin/charon-driver/main.rs
```diff
@@ -21,6 +21,7 @@ extern crate rustc_middle;
 extern crate rustc_session;
 extern crate rustc_span;
 extern crate rustc_target;
+extern crate rustc_trait_selection;
```
Nadrieril: This is no longer needed

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -270,10 +359,6 @@ impl ItemTransCtx<'_, '_> {
         };
 
         // Add the basic fields.
-        // Field: `size: usize`
-        mk_field("size".into(), usize_ty());
-        // Field: `align: usize`
-        mk_field("align".into(), usize_ty());
         // Field: `drop: fn(*mut Self)`
```
Nadrieril: Please undo this change, this creates an inconsistency with how the instances are defined

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -222,7 +222,6 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
                 let tref = self.translate_closure_type_ref(span, args)?;
                 TyKind::Adt(tref)
             }
-
```
Nadrieril, line 224: Please remove this unrelated change

charon/src/bin/charon-driver/translate/translate_predicates.rs
```diff
@@ -94,20 +94,31 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                            }
+                            PredicateLocation::Item(item_name) => self
+                                .item_trait_clauses
+                                .entry(item_name.clone())
+                                .or_default(),
+                            PredicateLocation::Parent => unreachable!(),
+                        };
+                        location.push_with(|clause_id| clause(clause_id));
+                    }
+                };
```
Nadrieril: Please remove this change, it is no longer relevant

charon/src/bin/charon-driver/translate/translate_meta.rs
```diff
@@ -301,6 +301,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
         trace!("Computing name for `{def_id:?}`");
 
         let parent_name = if let Some(parent_id) = &def_id.parent {
+            trace!("About to get hax-def for {item:?}");
```
Nadrieril: Please remove this. If you want to trace `hax_def`, the tracing should happen in that function not here (and could use `tracing::instrument` instead of a manual `trace!` call).

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -77,6 +135,18 @@ impl ItemTransCtx<'_, '_> {
...
+        fun_id: FunDeclId,
+        item_meta: ItemMeta,
+        impl_func_def: &hax::FullDef,
+    ) -> Result<FunDecl, Error> {
+        Err(Error {
+            span: item_meta.span,
+            msg: String::from("TODO"),
+        })
+    }
+
```
Nadrieril: Please remove this unrelated change. Also, you may have noticed that I separated the methods of this file into three logical blocks (with a comment on each), and this would not belong here but in the last block (where you can see that there's already a commented-out function for this purpose).

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -1,11 +1,17 @@
...
+use charon_lib::ast::*;
+use hax_frontend_exporter::{self as hax};
+
+use super::translate_crate::TransItemSourceKind;
+use charon_lib::formatter::IntoFormatter;
 use charon_lib::ids::Vector;
+use charon_lib::pretty::FmtWithCtx;
 use charon_lib::ullbc_ast::*;
-use hax_frontend_exporter as hax;
 use itertools::Itertools;
```
Nadrieril: Most of these changes are not needed and result in more confusing imports. Please only change what you need

charon/src/bin/charon-driver/main.rs
ssyram: Fixed.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: I think drop should come first in Rustc?

ssyram: Maybe we can fix also in the instance?

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Empty line? OK...

charon/src/bin/charon-driver/translate/translate_predicates.rs
ssyram: Deleted.

charon/src/bin/charon-driver/translate/translate_meta.rs
ssyram: OK. The place was for determining what was being translated when the error reported yesterday happened

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Deleted.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: This was from the merging... Now should be more clean.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: What rustc does doesn't matter here, I prefer something easy to follow.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -361,8 +461,8 @@ impl ItemTransCtx<'_, '_> {
         let mut generics = self.into_generics();
         {
             dyn_self = dynify(dyn_self, None);
-            generics = dynify(generics, Some(dyn_self.clone()));
-            kind = dynify(kind, Some(dyn_self.clone()));
+            generics = dynify(generics, None);
+            kind = dynify(kind, None);
```
Nadrieril: This is what I was mentioning on Zulip. Please revert that, it's causing your "unexpected Self" errors

ssyram: OK! Thanks for pointing it out!

ssyram:
The original info is too messy, I reset everything and let it be the same as #796 and will start working on it from this point.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -348,17 +348,34 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
     /// Checks whether the given id corresponds to a built-in type.
     pub(crate) fn recognize_builtin_type(
         &mut self,
+        span: Span,
         item: &hax::ItemRef,
     ) -> Result<Option<BuiltinTy>, Error> {
         let def = self.hax_def(item)?;
         let ty = if def.lang_item.as_deref() == Some("owned_box") && !self.t_ctx.options.raw_boxes {
+            // When Box is encountered, enqueue the translation of its drop implementation
+            self.enqueue_box_drop_impl(span, item)?;
```
Nadrieril, line 349: You shouldn't need a hack like that, what caused this?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -306,7 +346,7 @@ impl ItemTransCtx<'_, '_> {
     /// struct TraitVTable<TraitArgs.., AssocTys..> {
     ///   size: usize,
     ///   align: usize,
-    ///   drop: fn(*mut dyn Trait<...>),
+    ///   drop: fn(&mut dyn Trait<...>),
```
Nadrieril: The pointer here was not a mistake, creating `&mut` references has semantic side-effects which I don't fully understand, but I know the `&mut` in `Drop::drop` is a source of trickiness in relationship to `Pin` so we should not create more references than rustc does. Hence the pointer (it's meant to call `drop_in_place` anyway, which takes a pointer).

Nadrieril:
Could you please split this PR into several (or at least make separate clean commits) for the various features it implements? I see at least:
- Remove the redundant `TraitDeclRef`s in `TraitRefKind`;
- Handle `dyn Trait` with associated types;
- Translate `dyn Trait` method calls;
- Support `dyn Fn*` for closures;
- There's probably more.

I can't review it like this with a messy commit history and a bunch of different features implemented at once.

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: I will need the `drop` for `Box` type, to fill the vtable metadata. But this is not usually translated.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Well, but the drop functions are all of type `fn<'0>(&'0 mut Self)`. So I used `&` to provide the region info.

Nadrieril: The `drop` method of the `Drop` trait is not the same thing as the `drop` function we need to put in the vtable. The `drop` in the vtable must be the monomorphization of `std::ptr::drop_in_place` for the correct type, that's the language rules. And `drop_in_place` takes a `*mut T` not `&mut T`.

Nadrieril:
I think this has all been merged now!

PR #766, opened 2025-07-09
Add target information to crate data

### LINKED ISSUE #764: OCaml library doesn't know target platorm
[Here](https://github.com/AeneasVerif/charon/blob/main/charon-ml/src/TypesUtils.ml#L177) the OCaml code assumes that the target platform for the represented crate has the same `usize` width as the one the OCaml code is compiled on. Even though these functions are not used anywhere right now, such information should be available to users of the OCaml API.

firefighterduck:
Adds target information to the crate data to correctly handle pointer sized integers and endianess when a crate is cross-compiled through Charon.

Fixes #764


firefighterduck:
There are a few open problems: 
* first of all, we now have to always have some target information when translating scalars, which is not great, but I don't think we can do much. 
* Second, the newly added tests will fail in CI, since it doesn't have the corresponding targets installed. Unfortunately, I don't know where to add them or how to best disable the tests if a developer doesn't have them installed locally.

charon/src/ast/values_utils.rs
```diff
@@ -117,9 +154,13 @@ impl ScalarValue {
         }
     }
 
-    pub fn int_is_in_bounds(ty: IntegerTy, v: i128) -> bool {
+    pub fn int_is_in_bounds(ptr_size: ByteCount, ty: IntegerTy, v: i128) -> bool {
         match ty {
-            IntegerTy::Isize => v >= (isize::MIN as i128) && v <= (isize::MAX as i128),
+            IntegerTy::Isize => {
+                let min = ptr_size_val_signed!(ptr_size, MIN).unwrap();
+                let max = ptr_size_val_signed!(ptr_size, MAX).unwrap();
```
Nadrieril: I'd prefer to have two helpers with a bit of duplicated code than these macros, if you don't mind

charon/src/ast/values_utils.rs
```diff
@@ -212,15 +244,19 @@ impl ScalarValue {
...
 
-    pub fn from_bits(ty: IntegerTy, bits: u128) -> Self {
-        Self::from_le_bytes(ty, bits.to_le_bytes())
+    pub fn from_bits(target_info: &TargetInfo, ty: IntegerTy, bits: u128) -> Self {
+        if target_info.is_little_endian {
+            Self::from_le_bytes(target_info.target_pointer_size, ty, bits.to_le_bytes())
+        } else {
+            Self::from_be_bytes(target_info.target_pointer_size, ty, bits.to_be_bytes())
+        }
     }
```
Nadrieril: I believe rustc stores these integers as host `u128`s, not target `u128`s (if it did targets, addition on these `u128`s would be all wrong). So we can rely on host operations for this. So the existing code was correct because it went "host integer -> bits -> host integer" in which case the chosen intermediate endianness is irrelevant.

Nadrieril: In fact we could simplify the conversion a lot by doing `u128 as u32` for the unsigned cases; it's only for signed ints that we have to be careful.

charon/src/ast/values_utils.rs
firefighterduck: Sure, I'll change it

charon/src/ast/values_utils.rs
Nadrieril: Actually it's worse: for `hax::DiscriminantValue`, the data comes from `ty::Discr` so what I said is hopefully correct. For the output of `tag_for_variant`, we get a [ScalarInt](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.ScalarInt.html), which says "The first size bytes of data are the value." which means we should use `INT::from_ne_bytes(u128::to_ne_bytes)` for that one I think?

Nadrieril: Conclusion: I propose that we use rustc's `to_uxxx` helpers instead of doing our own bit manipulation. So some of that will have to be done on the hax side.

Nadrieril:
FYI I just updated the rustc pin which is why you're getting weird errors. Plz rebase on top of `main`

Regarding target tests, I'll have to figure out how to install more targets in CI

charon/src/ast/values_utils.rs
firefighterduck: So just to make sure I understand you correctly: We should not do the bit conversion in Charon but in hax (except for the `tag_for_variant` scalar which we kinda need to translate in Charon through `*_ne_bytes`). Is that it?

Nadrieril: Thinking about this more, I feel like the current representation in Charon is not useful. For pretty much zero analyses do I need to have the correct type of integer in a big enum. I propose the following:
```rust
enum ScalarValue {
  Unsigned(IntegerTy, u128),
  Signed(IntegerTy, i128),
}
```
Then:
- for `tag_for_variant`, we use `to_uint` or `to_int`;
- for `hax::DiscriminantValue`, we should do the right conversion even though I'm now confused about how it should work now (doc is [here](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/util/struct.Discr.html), shouldn't be hard to suss out with appropriate examples and access to a big endian target). Can always keep the existing logic and debug later.

Nadrieril: Either way I think you could split the PR that adds `TargetInformation` from the one that fixes our scalar encoding; as you wish.

firefighterduck: > ```rust
> enum ScalarValue {
>   Unsigned(IntegerTy, u128),
>   Signed(IntegerTy, i128),
> }
> ```
 
That's also almost how it is done on the OCaml side, isn't it? (Which reminds me that I wanted to also clean up a few things about how to handle scalars there.)

> Either way I think you could split the PR that adds `TargetInformation` from the one that fixes our scalar encoding; as you wish.

That sounds like a good idea. I'll do that later.

Nadrieril: Yeah, the ocaml side got it right imo ^^ We're just forced to split signed/unsigned on the Rust side because `i128` can't hold `u128::MAX` and vice-versa; tbh eventually I might want to split `IntegerTy` into `IntTy` and `UIntTy` too and propagate the signed/unsigned split to ocaml.

Overall the whole API around scalars doesn't satisfy me quite yet, might change things around in the coming future.

PR #757, opened 2025-07-07
Escherichia/match on char

EschericHya:
Should fix https://github.com/AeneasVerif/charon/issues/546 although I'm not sure about which `int_ty` should be put in the `SwitchInt`

ci: use https://github.com/AeneasVerif/aeneas/pull/590
ci: use https://github.com/AeneasVerif/eurydice/pull/282

PR #771, opened 2025-07-10
Change `IntegerTy` and `ScalarValue` representations

firefighterduck:
Based on https://github.com/AeneasVerif/charon/pull/766#discussion_r2195220005 and https://github.com/AeneasVerif/charon/pull/766#discussion_r2195319340

Changes `ScalarValue` to only differentiate between signed and unsigned and store 128 bit and the integer type either way. Also changes that the `IntegerTy` is now split into two types, `IntTy` and `UIntTy` following the model from `hax`.

I can't say whether this is the best way to do everything, but it at least seems to work (i.e. the test succeed). Will stay a draft until we have decided whether this is the way we want to represent scalar values and integer types.

Will also fail tests in CI until we figured out how to handle alternative targets.

ci: use https://github.com/AeneasVerif/eurydice/pull/247
ci: use https://github.com/AeneasVerif/aeneas/pull/579

### DIFF
--- a/charon-ml/src/Scalars.ml
+++ b/charon-ml/src/Scalars.ml
@@ -93,15 +93,28 @@ let scalar_max ptr_size (int_ty : integer_type) : big_int =
 let check_int_in_range ptr_size (int_ty : integer_type) (i : big_int) : bool =
   Z.leq (scalar_min ptr_size int_ty) i && Z.leq i (scalar_max ptr_size int_ty)
 
+let get_val (scalar : scalar_value) =
+  match scalar with
+  | SignedScalar (_, v) | UnsignedScalar (_, v) -> v
+
+let get_ty (scalar : scalar_value) =
+  match scalar with
+  | SignedScalar (int_ty, _) -> Signed int_ty
+  | UnsignedScalar (uint_ty, _) -> Unsigned uint_ty
+
 (** Check that a scalar value is correct (the integer value it contains is in
     range) *)
 let check_scalar_value_in_range ptr_size (v : scalar_value) : bool =
-  check_int_in_range ptr_size v.int_ty v.value
+  check_int_in_range ptr_size (get_ty v) (get_val v)
 
 (** Make a scalar value, while checking the value is in range *)
 let mk_scalar ptr_size (int_ty : integer_type) (i : big_int) :
     (scalar_value, unit) result =
-  if check_int_in_range ptr_size int_ty i then Ok { value = i; int_ty }
+  if check_int_in_range ptr_size int_ty i then
+    Ok
+      (match int_ty with
+      | Signed int_ty -> SignedScalar (int_ty, i)
+      | Unsigned uint_ty -> UnsignedScalar (uint_ty, i))
   else Error ()
 
 let integer_type_is_signed (int_ty : integer_type) : bool =
--- a/charon-ml/src/Scalars.ml
+++ b/charon-ml/src/Scalars.ml
@@ -61,33 +61,33 @@ let usize_max ptr_size =
 
 let scalar_min ptr_size (int_ty : integer_type) : big_int =
   match int_ty with
-  | Isize -> isize_min ptr_size
-  | I8 -> i8_min
-  | I16 -> i16_min
-  | I32 -> i32_min
-  | I64 -> i64_min
-  | I128 -> i128_min
-  | Usize -> usize_min ptr_size
-  | U8 -> u8_min
-  | U16 -> u16_min
-  | U32 -> u32_min
-  | U64 -> u64_min
-  | U128 -> u128_min
+  | Signed Isize -> isize_min ptr_size
+  | Signed I8 -> i8_min
+  | Signed I16 -> i16_min
+  | Signed I32 -> i32_min
+  | Signed I64 -> i64_min
+  | Signed I128 -> i128_min
+  | Unsigned Usize -> usize_min ptr_size
+  | Unsigned U8 -> u8_min
+  | Unsigned U16 -> u16_min
+  | Unsigned U32 -> u32_min
+  | Unsigned U64 -> u64_min
+  | Unsigned U128 -> u128_min
 
 let scalar_max ptr_size (int_ty : integer_type) : big_int =
   match int_ty with
-  | Isize -> isize_max ptr_size
-  | I8 -> i8_max
-  | I16 -> i16_max
-  | I32 -> i32_max
-  | I64 -> i64_max
-  | I128 -> i128_max
-  | Usize -> usize_max ptr_size
-  | U8 -> u8_max
-  | U16 -> u16_max
-  | U32 -> u32_max
-  | U64 -> u64_max
-  | U128 -> u128_max
+  | Signed Isize -> isize_max ptr_size
+  | Signed I8 -> i8_max
+  | Signed I16 -> i16_max
+  | Signed I32 -> i32_max
+  | Signed I64 -> i64_max
+  | Signed I128 -> i128_max
+  | Unsigned Usize -> usize_max ptr_size
+  | Unsigned U8 -> u8_max
+  | Unsigned U16 -> u16_max
+  | Unsigned U32 -> u32_max
+  | Unsigned U64 -> u64_max
+  | Unsigned U128 -> u128_max
 
 (** Check that an integer value is in range *)
 let check_int_in_range ptr_size (int_ty : integer_type) (i : big_int) : bool =
@@ -106,5 +106,5 @@ let mk_scalar ptr_size (int_ty : integer_type) (i : big_int) :
 
 let integer_type_is_signed (int_ty : integer_type) : bool =
   match int_ty with
-  | Isize | I8 | I16 | I32 | I64 | I128 -> true
-  | Usize | U8 | U16 | U32 | U64 | U128 -> false
+  | Signed _ -> true
+  | Unsigned _ -> false
--- a/charon-ml/src/Scalars.ml
+++ b/charon-ml/src/Scalars.ml
@@ -31,35 +31,58 @@ let u128_max = Z.of_string "340282366920938463463374607431768211455"
     - using the interpreter in *concrete* mode and evaluating operations like
       addition, negation, etc. It is thus ok to not use the precise bounds. *)
 
-let isize_min = i64_min
-let isize_max = i64_max
-let usize_min = u64_min
-let usize_max = u64_max
+let isize_min ptr_size =
+  match ptr_size with
+  | 8 -> i64_min
+  | 4 -> i32_min
+  | 2 -> i16_min
+  | _ -> raise (Failure "Unsupported target pointer size")
 
-let scalar_min (int_ty : integer_type) : big_int =
+let isize_max ptr_size =
+  match ptr_size with
+  | 8 -> i64_max
+  | 4 -> i32_max
+  | 2 -> i16_max
+  | _ -> raise (Failure "Unsupported target pointer size")
+
+let usize_min ptr_size =
+  match ptr_size with
+  | 8 -> u64_min
+  | 4 -> u32_min
+  | 2 -> u16_min
+  | _ -> raise (Failure "Unsupported target pointer size")
+
+let usize_max ptr_size =
+  match ptr_size with
+  | 8 -> u64_max
+  | 4 -> u32_max
+  | 2 -> u16_max
+  | _ -> raise (Failure "Unsupported target pointer size")
+
+let scalar_min ptr_size (int_ty : integer_type) : big_int =
   match int_ty with
-  | Isize -> isize_min
+  | Isize -> isize_min ptr_size
   | I8 -> i8_min
   | I16 -> i16_min
   | I32 -> i32_min
   | I64 -> i64_min
   | I128 -> i128_min
-  | Usize -> usize_min
+  | Usize -> usize_min ptr_size
   | U8 -> u8_min
   | U16 -> u16_min
   | U32 -> u32_min
   | U64 -> u64_min
   | U128 -> u128_min
 
-let scalar_max (int_ty : integer_type) : big_int =
+let scalar_max ptr_size (int_ty : integer_type) : big_int =
   match int_ty with
-  | Isize -> isize_max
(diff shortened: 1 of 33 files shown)


Nadrieril:
As far as I can tell there's no easy way to have more targets in CI today (waiting on [this issue](https://github.com/ipetkov/crane/issues/864)). It's unfortunate but could you `//@ ignore` the target-specific tests? (We in fact already have a target test in `cli.rs` which is likewise ignored)

charon/src/bin/generate-ml/templates/GAstOfJson.ml
```diff
@@ -26,5 +26,6 @@ type id_to_file_map = file FileId.Map.t
 type of_json_ctx = id_to_file_map
 
 let path_buf_of_json = string_of_json
+let target_ptr_size = ref 0
```
Nadrieril: I don't like that, can we just ask for the `target_information` as input for the functions that need it?

charon/src/bin/generate-ml/main.rs
```diff
@@ -1029,7 +1031,8 @@ fn generate_ml(
                 (* Note that we use unbounded integers everywhere.
                    We then harcode the boundaries for the different types.
                  *)
-                { value : big_int; int_ty : integer_type }
+                | SignedScalar of big_int * int_type
+                | UnsignedScalar of big_int * u_int_type
```
Nadrieril: Can we make this auto-generated now, as long as we translate `u128`/`i128` to `big_int`?

charon/src/ast/types.rs
```diff
@@ -520,14 +520,34 @@ pub struct Field {
     Ord,
     PartialOrd,
 )]
-#[charon::rename("IntegerType")]
-pub enum IntegerTy {
+#[charon::rename("IntType")]
```
Nadrieril: Any particular reason not to name it `IntTy` in ocaml as well?

charon/src/ast/types.rs
```diff
@@ -651,7 +693,8 @@ pub struct TypeDeclRef {
 #[charon::rename("LiteralType")]
 #[charon::variants_prefix("T")]
 pub enum LiteralTy {
-    Integer(IntegerTy),
+    Integer(IntTy),
+    UnsignedInteger(UIntTy),
```
Nadrieril: I'd prefer `Int`/`UInt` as variant names.

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -890,13 +908,34 @@ impl BodyTransCtx<'_, '_, '_> {
                 let targets: Vec<(ScalarValue, BlockId)> = targets
                     .iter()
                     .map(|(v, tgt)| {
-                        let v = ScalarValue::from_le_bytes(int_ty, v.data_le_bytes);
+                        let v = ScalarValue::from_bytes(IntegerTy::Signed(int_ty), v.data_le_bytes);
```
Nadrieril: Let's keep the `from_le_bytes` name, a plain `from_bytes` doesn't say how the bytes are to be interpreted.

charon/src/ast/types.rs
firefighterduck: Not really, I just kept the naming scheme from `IntegerTy` to `IntegerType` in OCaml.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -64,26 +64,23 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
-                }))
-            }
+            hax::TyKind::Int(int_ty) => TyKind::Literal(LiteralTy::Integer(match int_ty {
+                hax::IntTy::Isize => IntTy::Isize,
+                hax::IntTy::I8 => IntTy::I8,
+                hax::IntTy::I16 => IntTy::I16,
+                hax::IntTy::I32 => IntTy::I32,
+                hax::IntTy::I64 => IntTy::I64,
+                hax::IntTy::I128 => IntTy::I128,
+            })),
```
Nadrieril: Could you extract a `translate_int_ty` helper to avoid duplicating this match?

charon/src/bin/charon-driver/translate/translate_bodies.rs
firefighterduck: Is the little endian encoding fixed by hax and rustc as well or is it the native endian encoding? I've changed the name and used functions such that hopefully Charon would also work on big endian targets by using the native endian encoding (not sure whether this would ever happen, but it's not too bad I guess).

Nadrieril: This is explicitly the output of `u128::to_le_bytes`. Tbf I'm not clear what the original `u128` was meant to encode; it's the `PU128` in [SwitchTargets](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.SwitchTargets.html) if you're curious.

charon/src/ast/types.rs
Nadrieril: Let's keep the same names when possible then

charon/src/bin/generate-ml/main.rs
firefighterduck: Is there a way to force Charon to translate the `u128`/`i128` to `big_int`? It seems to translate it to just `int`.

Nadrieril: Yep, it's in `type_to_ocaml_name`

charon/src/bin/generate-ml/templates/GAstOfJson.ml
firefighterduck: I've now added it to the ctx since that seemed like the best way.

charon/src/ast/values_utils.rs
```diff
@@ -155,12 +151,12 @@ impl ScalarValue {
...
             ScalarValue::Unsigned(_, v) => v,
-            ScalarValue::Signed(_, v) => u128::from_be_bytes(v.to_ne_bytes()),
+            ScalarValue::Signed(_, v) => u128::from_le_bytes(v.to_le_bytes()),
         }
     }
 
-    pub fn from_bytes(ty: IntegerTy, bytes: [u8; 16]) -> Self {
-        from_ne_bytes!(
+    pub fn from_le_bytes(ty: IntegerTy, bytes: [u8; 16]) -> Self {
+        from_le_bytes!(
```
Nadrieril, line 162: I assume we can't just use `u128::from_le_bytes`/`i128::from_le_bytes`; could you add a comment explaining why not if that's the case?

charon-ml/src/Scalars.ml
```diff
@@ -31,56 +31,93 @@ let u128_max = Z.of_string "340282366920938463463374607431768211455"
...
-let mk_scalar (int_ty : integer_type) (i : big_int) :
+let mk_scalar ptr_size (int_ty : integer_type) (i : big_int) :
     (scalar_value, unit) result =
-  if check_int_in_range int_ty i then Ok { value = i; int_ty } else Error ()
+  if check_int_in_range ptr_size int_ty i then
+    Ok
+      (match int_ty with
+      | Signed int_ty -> SignedScalar (int_ty, i)
+      | Unsigned uint_ty -> UnsignedScalar (uint_ty, i))
+  else Error ()
```
Nadrieril, line 118: I hadn't realized that this was why you needed the pointer size. Let's remove that `check_int_in_range` check, it's fine to trust charon to emit the correct integers. We can keep the `check_int_in_range` function if ocaml users want it though

firefighterduck: Funnily enough, this check helped me find a bug while translating the scalars. Fortunately, we will not need this handwritten translation  anymore, since the serialization/deserialization can now be completely generated (will push in a few minutes).

Nadrieril: Oh nice, I was hoping we'd get there :)

firefighterduck: Maybe it'll take a bit longer. I've just noticed that we somehow translate values with more than 53 bits (the integer bit width for json) / more than 63 bits (the OCaml int bit width). This now fails for some tests and I don't even know why it worked before. I'll try to see how to fix this, but I might not be able to finish it today.

Nadrieril: That's because we unconditionally represented the integers in `ScalarValue` as strings, isn't it? We could possibly switch ot using `serde_json`'s `arbitrary_precision` feature, which uses strings for all numbers.

firefighterduck: Yeah, I've remembered a minute after turning off my computer. Due to the changes to the representation, I was able to reduce the custom impls.

charon/src/ast/values_utils.rs
firefighterduck: I've added the following comment, that hopefully makes it clear:
```rust
/// Translates little endian bytes into a corresponding `ScalarValue`.
/// This needs to do the round-trip to the correct integer type to guarantee
/// that the values are correctly sign-extended (e.g. if the bytes encode -1i8, taking all 16 bytes
/// would lead to the value 255i128 instead of -1i128).
```

charon-ml/src/Scalars.ml
N1ark: > Maybe it'll take a bit longer. I've just noticed that we somehow translate values with more than 53 bits (the integer bit width for json) / more than 63 bits (the OCaml int bit width). This now fails for some tests and I don't even know why it worked before. I'll try to see how to fix this, but I might not be able to finish it today.

Isn't the 53 bits just a limitation from parsing the JSON in JavaScript, as integers are floats? afaik the JSON spec doesn't specify the bit size of integers, since it's just a text format

firefighterduck: It might be, that number was the only thing I found. Nonetheless, due to the 63 bits in OCaml one needs to be careful with any type that could have values that can't be represented as an OCaml `int`.

firefighterduck:
Open TODO's:

- [ ] Is the representation now as it should be?
- [ ] Is the safe translation of `usize`/`isize`/`u128`/`i128`/`u64`/`i64` to `big_int` in OCaml necessary? It requires a lot of changes and might only be required for scalars with their 128 bits.
- [ ] Write fixes for Aeneas/Euridice (probably also kyber?) and use them for the CI here.
- [ ] Squash commits to keep history clean.

charon/src/bin/generate-ml/main.rs
```diff
@@ -175,7 +175,16 @@ fn type_to_ocaml_call(ctx: &GenerateCtx, ty: &Ty) -> String {
     match ty.kind() {
         TyKind::Literal(LiteralTy::Bool) => "bool_of_json".to_string(),
         TyKind::Literal(LiteralTy::Char) => "char_of_json".to_string(),
-        TyKind::Literal(LiteralTy::Integer(_)) => "int_of_json".to_string(),
+        TyKind::Literal(LiteralTy::Int(int_ty)) => match int_ty {
+            // Since OCaml ints are only 63 bits, translate everything that could be too large to big_int
+            IntTy::I128 | IntTy::Isize | IntTy::I64 => "big_int_of_json".to_string(),
+            _ => "int_of_json".to_string(),
+        },
```
Nadrieril: My opinion is to keep `int` for this `isize`/`usize`/`i64`/`u64`. Integers that represent target values should mostly be `u128`/`i128`, whereas the `usize` etc used everywhere most often are things that count from zero and have no reasonable chance of reaching the max representable value of OCaml. If it ever happens we'll get an exception on the ocaml side and we can reconsider this change.

charon/src/ast/types.rs
```diff
@@ -534,6 +552,28 @@ pub enum IntegerTy {
...
+    DriveMut,
+    Hash,
+    Ord,
+    PartialOrd,
+)]
+#[charon::rename("IntegerType")]
+pub enum IntegerTy {
+    Signed(IntTy),
+    Unsigned(UIntTy),
+}
```
Nadrieril, line 560: Where do we use `IntegerTy`? Could we just use `LiteralTy` everywhere instead?

charon/src/ast/values.rs
```diff
@@ -63,25 +61,18 @@ pub enum Literal {
...
-    I128(i128),
-    /// Using u64 to be safe
-    Usize(u64),
-    U8(u8),
-    U16(u16),
-    U32(u32),
-    U64(u64),
-    U128(u128),
+    #[serde(with = "crate::ast::values_utils::scalar_value_ser_de")]
+    Unsigned(UIntTy, u128),
```
Nadrieril: You might even be able to do something like this, no?
```suggestion
    Unsigned(UIntTy, #[serde(with = "crate::ast::values_utils::u128_ser_de")] u128),
```

charon/src/ast/types.rs
firefighterduck: Other than for `SwitchInt`, which get's changed by https://github.com/AeneasVerif/charon/pull/757 anyway, we only use it to make the `ScalarValue` API a bit nicer. However, `LiteralTy` could also work as well. Maybe we then also have scalar values of other literals (such as characters)? Would that make sense?

Nadrieril: Oh right, `ScalarValue` is actually pretty much `IntegerValue`. We already have `Literal` that also represents floats, chars, etc. Let's keep `IntegerTy` for this PR, I want to clean up these types eventually but now sure how yet.

charon/src/ast/values.rs
firefighterduck: Yes, this does work. I tried to figure out whether it would work, but I couldn't find any documentation about adding attributes to fields of tuple variants (for serde). Ideally I'd like to tell serde to translate the types via a string as an attribute without the boilerplate module but this doesn't seem to work.

charon/src/bin/generate-ml/main.rs
firefighterduck: Sounds good, I will revert the changes in the translation.

firefighterduck:
ci: use https://github.com/AeneasVerif/eurydice/pull/247
ci: use https://github.com/AeneasVerif/aeneas/pull/579

PR #773, opened 2025-07-16
Bump `annotate-snippets-rs`

N1ark:
When using Charon as a dependency, Cargo keeps trying to install a newer version of `annotate-snippets-rs`, since they don't seem to fully follow versioning -- I just bumped the dependency for it to be up to date and not break whenever I `cargo update` :)


Nadrieril:
:D Thanks for figuring out the appropriate API changes ^^

Issue #769, opened 2025-07-10
Bug: excessively verbose LLBC for PartialEq implementations


msprotz:
I have this type: https://github.com/AeneasVerif/symcrust/blob/0371b8667c11f44721b7db4c78cbe1d1ee7546db/src/common.rs#L7-L31

The PartialEq implementation for it is:

```
fn symcrust::common::{core::cmp::PartialEq<symcrust::common::Error> for symcrust::common::Error}::eq<0, 1>(self^1 : &0 (symcrust::common::Error), other^2 : &1 (symcrust::common::Error)) -> bool
{
    v@0 : bool;
    self^1 : &'_ (symcrust::common::Error);
    other^2 : &'_ (symcrust::common::Error);
    __self_discr^3 : isize;
    v@4 : &'_ (symcrust::common::Error);
    __arg1_discr^5 : isize;
    v@6 : &'_ (symcrust::common::Error);
    v@7 : isize;
    v@8 : isize;

    storage_live __self_discr^3;
    storage_live v@4;
    v@4 := &*(self^1);
    match (*(v@4)) {
    | 0 => {
        __self_discr^3 := (0: isize : isize)
      }
    | 1 => {
        __self_discr^3 := (32768: isize : isize)
      }
    | 2 => {
        __self_discr^3 := (32769: isize : isize)
      }
    | 3 => {
        __self_discr^3 := (32770: isize : isize)
      }
    | 4 => {
        __self_discr^3 := (32771: isize : isize)
      }
    | 5 => {
        __self_discr^3 := (32772: isize : isize)
      }
    | 6 => {
        __self_discr^3 := (32773: isize : isize)
      }
    | 7 => {
        __self_discr^3 := (32774: isize : isize)
      }
    | 8 => {
        __self_discr^3 := (32775: isize : isize)
      }
    | 9 => {
        __self_discr^3 := (32776: isize : isize)
      }
    | 10 => {
        __self_discr^3 := (32777: isize : isize)
      }
    | 11 => {
        __self_discr^3 := (32778: isize : isize)
      }
    | 12 => {
        __self_discr^3 := (32779: isize : isize)
      }
    | 13 => {
        __self_discr^3 := (32780: isize : isize)
      }
    | 14 => {
        __self_discr^3 := (32781: isize : isize)
      }
    | 15 => {
        __self_discr^3 := (32782: isize : isize)
      }
    | 16 => {
        __self_discr^3 := (32783: isize : isize)
      }
    | 17 => {
[...]

PR #774, opened 2025-07-17
 Don't replace discriminant reads with matches

Nadrieril:
Fixes https://github.com/AeneasVerif/charon/issues/769.


msprotz:
Does this need a deep fix on the Eurydice side? Happy to help

Nadrieril:
I don't know how deep, would you mind having a look?

msprotz:
```diff
diff --git a/lib/AstOfLlbc.ml b/lib/AstOfLlbc.ml
index 940c57d..87a7684 100644
--- a/lib/AstOfLlbc.ml
+++ b/lib/AstOfLlbc.ml
@@ -1806,7 +1806,9 @@ let expression_of_rvalue (env : env) (p : C.rvalue) : K.expr =
   | UnaryOp (op, o1) -> mk_op_app (op_of_unop op) (expression_of_operand env o1) []
   | BinaryOp (op, o1, o2) ->
       mk_op_app (op_of_binop op) (expression_of_operand env o1) [ expression_of_operand env o2 ]
-  | Discriminant _ -> failwith "expression_of_rvalue Discriminant"
+  | Discriminant p ->
+      let e = expression_of_place env p in
+      K.(with_type (TInt PtrdiffT) (ECast (e, TInt PtrdiffT)))
   | Aggregate (AggregatedAdt ({ id = TTuple; _ }, _, None), ops) -> begin
       match ops with
       | [] -> K.with_type TUnit K.EUnit
```

this seems to fix it for me if you agree with the fix?

msprotz:
The plan on record is, just like with Eurydice_array_repeat, to leave in the Ast a call to a special function `Eurydice_discriminant<T,U>(x)` with type T -> U (obviously loosely typed), where T is the type of the argument `x` and U is the expected type of the discriminant. Then, once data types have been compiled and optimized, a dedicated pass will replace this with `(isize)x.tag`, or `(u8)x`, depending on how the enum was compiled down to C.

msprotz:
Commit above should work -- at least it does on my local copy. I'll let you merge both branches simultaneously if it makes sense. Thank you

PR #782, opened 2025-07-22
Update the Substitute.ml file

sonmarcho:


sonmarcho:
Yes that's true, but those comments were really pertinent in Aeneas which provides functions to normalize the types, and in which context one had to be extremely careful to call the wrappers which also normalize types, not the original functions (otherwise the environments would be ill-typed). Many comments referred to the AssociatedTypes file for instance, which doesn't exist in Charon.

Nadrieril:
Fair, I don't have a strong opinion either way.

PR #783, opened 2025-07-24
Represent globals as places

N1ark:
Update of https://github.com/AeneasVerif/charon/pull/649 -- they're practically the same
The main notable difference is that when using the *value* of a global, it is copied, not moved, as moving a global would uninitialise it; e.g. here
```rust
const C = 0;
let a = C;
let b = C; // if C is moved, this is an error
```

ci: use https://github.com/AeneasVerif/aeneas/pull/586
ci: use https://github.com/AeneasVerif/eurydice/pull/251

### DIFF
--- a/charon/src/ast/expressions.rs
+++ b/charon/src/ast/expressions.rs
@@ -27,8 +27,13 @@ pub struct Place {
 )]
 #[charon::variants_prefix("Place")]
 pub enum PlaceKind {
+    /// A local variable in a function body.
     Local(LocalId),
+    /// A subplace of a place.
     Projection(Box<Place>, ProjectionElem),
+    /// A global (const or static).
+    /// Not present in MIR; introduced in [simplify_constants.rs].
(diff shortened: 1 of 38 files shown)


charon/src/ast/expressions.rs
```diff
@@ -461,7 +466,7 @@ impl From<FunDeclRef> for FnPtr {
 /// We later desugar those to regular ADTs, see [regularize_constant_adts.rs].
 ///
 /// [`RawConstantExpr::Global`] case: access to a global variable. We later desugar it to
-/// a separate statement.
+/// a copy of a place global.
```
Nadrieril, line 469: You removed it now :D

N1ark: oops yupppp

Nadrieril:
Looks good! Now to do the aeneas & eurydice changes :"). Feel free to try, otherwise I'll have a try but this will take me a while to get to.

Issue #608, opened 2025-03-14
Better support for unsized types


Nadrieril:
Currently the translation of unsized types requires the user to figure out metadata manipulation themselves: e.g. a cast from `Box<[T; N]>` to `Box<[T]>` makes a fat pointer but nothing in charon makes that clear.

Ideally charon would tell you exactly what metadata to use whenever creating a pointer that happens to be fat.

- [x] Make explicit which metadata is used on a given unsizing coercion;
- [x] Make pointer metadata tracking explicit;
- [ ] Support unsizing behind non-builtin pointers;

PR #797, opened 2025-08-04
Better Unsized Type support

ssyram:
Following #608 , this PR aims to provide the metadata explicitly when a reference / raw pointer is created. Specifically, a new field `Option<Operand>` is added to both `Rvalue::Ref` & `Rvalue::RawPtr`. It is `None` to correspond to the explicit `None` case of the `PtrMetadata`.

This also involves a rework on the data structure of `PtrMetadata`. Now, `PtrMetadata::VTable` refers directly to the virtual table struct generated. Also, a variant `PtrMetadata::InheritFrom` is added to handle the generic case.

Currently still working on adding the Operand obtainment.


Nadrieril:
CI is failing, could you fix it before I have a look?

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -414,7 +414,13 @@ impl BodyTransCtx<'_, '_, '_> {
             hax::Rvalue::Ref(_region, borrow_kind, place) => {
                 let place = self.translate_place(span, place)?;
                 let borrow_kind = translate_borrow_kind(*borrow_kind);
-                Ok(Rvalue::Ref(place, borrow_kind))
+                Ok(Rvalue::Ref {
+                    place,
+                    kind: borrow_kind,
+                    // Use `()` as a placeholder now.
+                    // Will be fixed by the cleanup pass `insert_ptr_metadata`.
+                    ptr_metadata: Operand::mk_const_unit(),
```
Nadrieril, line 427: Why fix it in a pass instead of adding it here (and we can still ask rustc for things)?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -382,7 +382,8 @@ impl ItemTransCtx<'_, '_> {
             },
             kind,
             layout: Some(layout),
-            ptr_metadata: None,
+            // a vtable must not have unsizing metadata itself
```
Nadrieril:
```suggestion
            // A vtable struct is always sized
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -1,4 +1,5 @@
 use super::translate_ctx::*;
+use anstream::panic;
```
Nadrieril: This probably comes from an import auto-added by rust-analyzer
```suggestion
```

charon/src/transform/insert_ptr_metadata.rs
```diff
@@ -0,0 +1,223 @@
...
+        // We will need to keep the derefed type to get the metadata type
+        ProjectionElem::Deref => Some((subplace.clone(), place.ty().clone())),
+        ProjectionElem::Field(proj_kind, field) if is_last_field(ctx, proj_kind, field) => {
+            outmost_deref_at_last_field(ctx, subplace)
+        }
+        // Otherwise, it could be one of the following cases:
+        // 1. It is a field access, but not the last field, so it does not have metadata;
+        // 2. It is an index / sub-slice access, which must return a reference,
+        //    As we are checking for *outmost*, it means no deref is performed,
+        //    so it must remain a reference, which does not have metadata.
```
Nadrieril: I would like to be able to make the pass that removes indexing projections optional. In other words, could you handle these projections too?

charon/src/transform/resolve_sized_ptr_metadata_inherit.rs
Nadrieril, line 1: Instead of resolving by hand like this, we should be asking rustc what the metadata type is (which requires doing this during translation, not after)

charon/src/transform/insert_ptr_metadata.rs
```diff
@@ -0,0 +1,223 @@
...
+            TyKind::Adt(type_decl_ref) => match &type_decl_ref.id {
+                TypeId::Adt(type_decl_id) => {
+                    is_last_field_of_ty_decl_id(ctx, type_decl_id, field)
+                }
+                TypeId::Tuple => type_decl_ref.generics.len() - 1 == field.index(),
+                // Builtin types are atoms
+                TypeId::Builtin(..) => panic!("Trying to get the last field from a built-in type"),
+            },
+            _ => panic!("Type \"{}\" does not have field, projection operation invalid!", ty.with_ctx(&ctx.into_fmt())),
+        },
```
Nadrieril: That should be an error too, alias types are never referenced from other places

charon/src/ast/types_utils.rs
```diff
@@ -630,6 +630,75 @@ impl Ty {
...
+                            // `None` refers to the unit type `()`
+                            None => PtrMetadata::None,
+                            // Otherwise, simply recurse
+                            Some(ty) => ty.get_ptr_metadata(translated),
+                        }
+                    }
+                    // Box simply recurse with its sole argument
+                    TypeId::Builtin(BuiltinTy::Box) => {
+                        ty_ref.generics.types[0].get_ptr_metadata(translated)
+                    }
```
Nadrieril: That's incorrect: `Box` itself has no metadata, it's a pointer

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -374,23 +379,46 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                                    .translate_vtable_struct_ref(span, &trait_predicate.trait_ref)?
+                                    .unwrap())
+                            },
+                        )?
+                        .erase();
+                    PtrMetadata::VTable(vtable)
+                }
+                _ => unreachable!("Unexpected hax type {hax_ty:?} for dynamic type: {ty:?}"),
+            },
+            ty::Param(..) => PtrMetadata::InheritFrom(self.translate_ty(span, hax_ty)?),
```
Nadrieril, line 410: I'm thinking we could check if the parameter is sized here so we don't have to do it in a pass. Would that be enough, or are there other ways to construct a `InheritFrom`?

charon/src/transform/index_to_function_calls.rs
```diff
@@ -22,18 +24,35 @@ struct IndexVisitor<'a> {
...
+    ctx: &'b TransformCtx,
 }
 
-impl<'a> IndexVisitor<'a> {
-    fn fresh_var(&mut self, name: Option<String>, ty: Ty) -> Place {
-        let var = self.locals.new_var(name, ty);
-        let live_kind = RawStatement::StorageLive(var.local_id().unwrap());
-        self.statements.push(Statement::new(self.span, live_kind));
-        var
+impl PtrMetadataComputable for IndexVisitor<'_, '_> {
```
Nadrieril: This trait is cool, could you just move it to `transform/mod.rs` or `transform/ctx.rs` since it's used by multiple passes? And maybe name it `BodyTransformCtx` or something

charon/src/transform/insert_ptr_metadata.rs
Nadrieril: Also, "which must return a reference" is either unclear or entirely incorrect. An indexing projection returns a place, not a reference value. If it's a single index then it works pretty much like a field projection, if it's a range then the resulting place has type slice and the metadata is the size of the range.

charon/tests/ui/simple/pointee_metadata.out
```diff
@@ -247,15 +253,21 @@ fn slice_metadata()
     let @3: *const Array<u32, 2 : usize>; // anonymous local
     let @4: &'_ (Array<u32, 2 : usize>); // anonymous local
     let @5: Array<u32, 2 : usize>; // anonymous local
+    let @6: (); // anonymous local
+    let @7: (); // anonymous local
```
Nadrieril: Could you make sure you always use a unit constant instead of a local when there's no metadata?

charon/src/bin/charon-driver/translate/translate_bodies.rs
ssyram: OK, how can we ask for that? I will have a try, though.

charon/src/transform/insert_ptr_metadata.rs
ssyram: The type of index / slice access is `fn(...) -> &[mut] Self::Output`, so it must be a reference?

charon/tests/ui/simple/pointee_metadata.out
ssyram: Is there a unit constant? That is an `Adt(Tuple), []`?

charon/src/transform/index_to_function_calls.rs
ssyram: OK, will do.

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: Yes, this will be cool. But we could still have that `Box<U : ?Sized>` but we could still pass `i32` to have `Box<i32>`. So this will construct `InheritFrom`.

charon/src/ast/types_utils.rs
ssyram: OK, will fix.

ssyram:
> CI is failing, could you fix it before I have a look?

Sorry? How to fix that? It seems to be the path inconsistency?

charon/src/transform/insert_ptr_metadata.rs
ssyram: So we can ignore this case? Simply error on this?

charon/tests/ui/simple/pointee_metadata.out
ssyram: OK, found, will use that.

ssyram: I used the `Operand::mk_const_unit` function. But then there is a pass that will eliminate the `RawConstantExpr::Adt` the case, hence the result is the same. In this case, I would like to use the original one, which has a more precise storage live operation.

charon/src/transform/index_to_function_calls.rs
ssyram: Trait & its method done. But why `BodyTransformCtx`?

ssyram:
OK, latest changes merged, but the results are the same: the support for assoc-types are required to let the overall process of fetching metadata go smoothly.

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Of course

charon/src/transform/index_to_function_calls.rs
Nadrieril: Because this doesn't fundamentally has anything to do with pointer metadata. It's reusable for any pass that modifies the body, no?

charon/src/transform/insert_ptr_metadata.rs
Nadrieril: That's for non-built-in indexing. `ProjectionElem::Index` is a built-in thing that acts on places of type array or slice in the way I described

charon/src/bin/charon-driver/translate/translate_bodies.rs
Nadrieril: This was a real question, maybe it's more convenient in a pass I don't know. And I don't know what we need to ask rustc, maybe the metadata for a given type?

charon/src/transform/insert_ptr_metadata.rs
Nadrieril: Panic or error yes

charon/tests/ui/simple/pointee_metadata.out
Nadrieril: Hm that's annoying, it causes a lot of churn in test outputs. I honestly prefer changing the pass so it keep the `ConstantExpr::Adt` in that one particular case.

charon/src/bin/charon-driver/translate/translate_bodies.rs
ssyram: OK, then maybe a new PR? The current pass seems to work pretty well?

charon/tests/ui/simple/pointee_metadata.out
ssyram: OK, then it might be a future PR? This does not matter to Eurydice though. Should I simply change it to `Operand::mk_const_unit` for your coming PR on this?

charon/src/transform/index_to_function_calls.rs
ssyram: OK, you're right. Will do.

charon/src/transform/insert_ptr_metadata.rs
ssyram: OK, will do.

charon/src/transform/insert_ptr_metadata.rs
ssyram: Let me study more and try to fix.

charon/src/ast/types_utils.rs
```diff
@@ -630,6 +630,73 @@ impl Ty {
...
+                // there are two cases:
+                // 1. if the declared type has a fixed metadata, just returns it
+                // 2. if it depends on some other types or the generic itself
+                match ty_ref.id {
+                    TypeId::Adt(type_decl_id) => {
+                        let Some(decl) = ty_decls.get(type_decl_id) else {
+                            panic!(
+                                "Internal Error: type decl id not found during getting metadata: {type_decl_id}"
+                            )
+                        };
```
Nadrieril, line 718: Please use `raise_error`, this can happen in normal usage of charon such as with `--opaque`. `panic` should be only used for cases that are theoretically impossible.

Nadrieril: Most of the panics you introduce in this PR are like this in fact, please `raise_error` instead

ssyram: OK, fixed, it is not raised error because the way to this point is too far, I just used the `InheritFrom(TyErr(...))`.

ssyram:
OK, but all tests passed locally and I'm not that sure what is the CI error...

charon/src/transform/insert_ptr_metadata.rs
ssyram: OK, fixed.

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: The `Sized` part is done.

ssyram:
Now the PR itself should be ready. Additionally, the `PtrMetadata<T>` for `Sized<T>` has been resolved to `()`, which greatly alleviated the generic computation.

charon/src/ast/types_utils.rs
```diff
@@ -819,6 +894,112 @@ impl Variant {
...
+            .with_ctx(&translated.into_fmt())
+            .to_string();
+
+        // 1. Get vtable ref from trait declaration
+        //    Provides: 1) final return ID, 2) correct order of associated types
+        // Firstly, get the trait declaration for the vtable ref it stores.
+        let trait_decl = translated
+            .trait_decls
+            .get(self.binder.params.trait_clauses[0].trait_.skip_binder.id)
+            .unwrap();
```
Nadrieril: This `unwrap` is not ok: we must not panic is an item is missing. Please just return `None` instead.

charon/src/ast/types_utils.rs
```diff
@@ -819,6 +894,112 @@ impl Variant {
...
+
+        // Get vtable ref from definition for correct ID.
+        // Generics in vtable ref are placeholders but provide correct order of the associated types.
+        let Some(vtable_ref) = &trait_decl.vtable else {
+            panic!(
+                "Vtable for trait {} is None, meaning the trait is non-dyn-compatible!",
+                trait_name
+            );
```
Nadrieril, line 941: This panic is ok however, because if it happens we (charon developers) made a mistake somewhere.

charon/src/ast/types_utils.rs
```diff
@@ -819,6 +894,112 @@ impl Variant {
...
+            .types
+            .iter()
+            .filter_map(|ty| {
+                if let TyKind::TraitType(tref, name) = &ty.kind() {
+                    Some((tref.trait_decl_ref.skip_binder.id, name.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
```
Nadrieril: note for the future: you may be interested in the `collect_vec` method (from `itertools::Itertools`) instead

charon/src/ast/types_utils.rs
```diff
@@ -819,6 +894,112 @@ impl Variant {
...
+
+        // Find correct type argument from dyn trait's constraints for each assoc type.
+        for (trait_id, assoc_name) in assoc_tys {
+            // Find it
```
Nadrieril, line 984: Are you sure you need to do all this? `vtable_ref` should already work no? It has all the types in the correct order. If what you're trying to do is associated type normalization then we should expose the right methods from `expand_associated_types.rs` instead of doing a special-case here.

Nadrieril: Anyway, I'm ok if you just add a `// TODO: use proper normalization here instead of doing it by hand`

charon/src/bin/charon-driver/translate/translate_bodies.rs
Nadrieril: You still haven't answered my question, but I think I found an answer: it's useful to have all the `compute_ptr_metadata` methods available for other passes anyway, so may as well do this in a pass.

charon/src/ast/types_utils.rs
ssyram: OK, thanks!

charon/src/ast/types_utils.rs
ssyram: Will try.

charon/src/ast/types_utils.rs
ssyram: Cool, thanks!

charon/src/ast/types_utils.rs
ssyram: OK, will fix.

charon/src/transform/insert_ptr_metadata.rs
```diff
@@ -0,0 +1,288 @@
...
+        ProjectionElem::Index { .. } => None,
+        // Subslice must have metadata length, but the key is which place to fetch it from?
+        // I.e., in places like `x[..]`, it returns `{[TypeOf<x>] as Index[Mut]}::Output`
+        ProjectionElem::Subslice { from, to, from_end } => {
+            let to_idx = compute_to_idx(ctx, subplace, *to.clone(), *from_end);
+            let diff_place = ctx.fresh_var(None, Ty::mk_usize());
+            ctx.insert_assn_stmt(
+                diff_place.clone(),
+                // `to` cannot be less than `from` as per Rust rules, so panic
+                Rvalue::BinaryOp(BinOp::Sub(OverflowMode::Panic), to_idx, *from.clone()),
```
Nadrieril: The correct mode is `OverflowMode::UB`; the MIR does not include a panic check here, it should already have been checked beforehand.

charon/src/transform/insert_ptr_metadata.rs
```diff
@@ -0,0 +1,288 @@
...
+            if is_last_field(ctx.get_ctx(), proj_kind, field) =>
+        {
+            outmost_deref_at_last_field(ctx, subplace)
+        }
+        // This is not the last field, so it will never have metadata
+        ProjectionElem::Field(..) => None,
+        // Indexing for array & slice will only result in sized types, hence no metadata
+        ProjectionElem::Index { .. } => None,
+        // Subslice must have metadata length, but the key is which place to fetch it from?
+        // I.e., in places like `x[..]`, it returns `{[TypeOf<x>] as Index[Mut]}::Output`
```
Nadrieril: That comment is incorrect, please remove it

Nadrieril: Actually it's correct but confusing because it makes it sound like the `Index` traits are involved yet they are not. I still don't find it useful tho.

charon/src/transform/insert_ptr_metadata.rs
ssyram: OK.

charon/src/transform/insert_ptr_metadata.rs
ssyram: OK, revised.

Nadrieril:
Could you squash the whole PR into one commit please?

Nadrieril:
> I think there is a merge option of letting the whole PR become a single commit?

Unfortunately not in combination with the GitHub merge queue, unless I wanted to squash _every_ PR which I don't.

ssyram:
OK, done.

ssyram:
OK, a very good news is that AeneasVerif/eurydice#259 is fully ready for this PR. The latest change in that PR merged the latest version of the main official branch of Eurydice and has been tested to be fully compatible with this PR. Hopefully, @protz we can polish AeneasVerif/eurydice#259 to get fully ready next week and when @Nadrieril is back the week after next, we can merge both PR altogether!

PR #800, opened 2025-08-08
Compute layouts on demand

firefighterduck:
### Description
To take another step to having full layout information for as many types as possible, this PR introduces a new `LayoutComputer` (name up for discussion), which can be used to compute memoized layouts (or partial layouts I call `LayoutHint`s) for all kinds of types.

### Goal
This is meant to cater to the need of tools like my own (now in the process of being rewritten in Rust and using `charon_lib`) which want to know layout information about many types, not just those with a type declaration.

### Status

The PR contains a lot of assumptions about when we can get which information and many of these should be discussed in greater detail. Right now it tries to supports the following cases:

*  ADTs with a known layout in the type decl
* ADTs without a known layout in the type decl by trying to instantiate its generic type parameters
* Tuples (precise if all elements are of the same size, otherwise only a hint)
* Builtins (Box as a pointer type, Slice and Str as unsized types)
* Literals
* References and raw pointers (precise if we know whether they need to be wide, otherwise only a hint)
* DynTrait as unsized

All other types are given no layout or hint. This is especially unfortunate for opaque ADTs from `std` for which we could do a lot (e.g. `Result<NonZero<u32>,()>` doesn't get a layout because `NonZero` is opaque). We could probably try to hardcode some layout schema for some of these types.

### DIFF
--- a/charon/src/ast/layout_computer.rs
+++ b/charon/src/ast/layout_computer.rs
@@ -1,0 +1,436 @@
+use std::{cmp::max, collections::HashMap};
+
+use either::Either;
+use serde::Serialize;
+
+use crate::ast::{
+    BuiltinTy, ByteCount, DeBruijnVar, Field, FieldId, Layout, TranslatedCrate, Ty, TyKind,
+    TypeDeclKind, TypeDeclRef, TypeId, TypeVarId, VariantLayout, Vector,
+};
+
+/// A utility to compute/lookup layouts of types from the crate's data.
+/// Uses memoization to not re-compute already requested type layouts.
+/// Should be constructed exactly once and kept around as long as the crate is used.
+///
+/// WARNING: Using this will lead to leaked memory to guarantee that computed layouts stay available.
+/// Since any crate should only have finitely many types, the memory usage is bounded.
+pub struct LayoutComputer<'a> {
+    memoized_layouts: HashMap<Ty, Either<&'a Layout, LayoutHint>>,
+    krate: &'a TranslatedCrate,
+}
+
+/// Minimal information about a type's layout that can be known without
+/// querying the rustc type layout algorithm.
+#[derive(Debug, Clone, Copy, Serialize)]
+pub struct LayoutHint {
+    /// The minimal size of the type.
+    ///
+    /// Includes the minimal sizes of all its fields (if known), but doesn't take
+    /// alignment and possible padding, or wide pointers into consideration.
+    pub min_size: ByteCount,
+    /// The minimal alignment.
+    ///
+    /// Corresponds to the greatest known alignment of the types fields.
+    pub min_alignment: ByteCount,
+    /// Whether thy type might be uninhabited.
+    /// This can be the case if there are type variables as parameters that
+    /// have influence on the layout of the type.
+    ///
+    /// Will only be set to false, if its guaranteed that it is inhabited.
+    pub possibly_uninhabited: bool,
+}
+
+impl<'a> LayoutComputer<'a> {
+    pub fn new<'b: 'a>(krate: &'b TranslatedCrate) -> Self {
+        Self {
+            memoized_layouts: HashMap::new(),
+            krate,
+        }
+    }
+
+    /// Compute the layout of a tuple as precise as possible.
+    ///
+    /// If the elements have different sizes and thus might need padding,
+    /// it returns a hint, otherwise the precise layout.
+    fn get_layout_of_tuple<'c, 't>(
+        &'c mut self,
+        type_decl_ref: &'t TypeDeclRef,
+    ) -> Option<Either<&'a Layout, LayoutHint>> {
+        let mut total_min_size = 0;
+        let mut max_align = 1;
+        let mut uninhabited_part = false;
+        let mut possibly_same_size = None;
+        let mut must_have_same_size = true;
+
+        for ty_arg in type_decl_ref.generics.types.iter() {
+            uninhabited_part |= ty_arg.is_possibly_uninhabited();
+            match self.get_layout_of(ty_arg) {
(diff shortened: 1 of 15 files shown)


firefighterduck:
> So basically this PR wants to compute the generic layout? I.e., the expression of the layout info?

No, the facility is meant to collect more available precise layout information that Charon currently can't use. E.g. for types such as `u32`, `(u32,u32)`, or `&str` Charon currently can't give you a layout even though much about them is obvious to us. For other cases such as `Box<T>`, it at least computes a layout hint, i.e. the minimal known information such as a lower bound on the size (a box can either be a wide or a thin pointer, thus the size of a pointer).

charon/src/ast/layout_computer.rs
```diff
@@ -0,0 +1,538 @@
...
+        Self {
+            memoized_layouts: HashMap::new(),
+            krate,
+        }
+    }
+
+    /// Compute the layout of a tuple as precise as possible.
+    ///
+    /// If the elements have different sizes and thus might need padding,
+    /// it returns a hint, otherwise the precise layout.
```
Nadrieril: Rust gives no guarantees about tuple layout, not even that the fields are in the expected order. Also size isn't the only thing at play, types can have the same size and different alignment (which is likely to cause reorderings). I guess for a tuple with all same size and alignment we can reasonably expect that the fields are in order, but I'd rather ask rustc directly if possible.

charon/src/ast/layout_computer.rs
```diff
@@ -0,0 +1,538 @@
...
+            Some(Either::Right(LayoutHint {
+                min_size: total_min_size,
+                min_alignment: max_align,
+                possibly_uninhabited: uninhabited_part,
+            }))
+        }
+    }
+
+    /// Computes the layout based on the simple algorithm for C types.
+    fn compute_c_layout_from_fields(
```
Nadrieril, line 177: `C` layout doesn't mean "all fields in order in the obvious way". There are weird alignment shenanigans on windows for example, hence this recent RFC: https://github.com/rust-lang/rfcs/pull/3845. We must therefore only provide layout hints for `repr(C)`.

charon/src/ast/layout_computer.rs
```diff
@@ -0,0 +1,538 @@
...
+        &mut self,
+        fields: &Vector<FieldId, Field>,
+        ty_var_map: &HashMap<TypeVarId, Ty>,
+        generic_ctx: GenericCtx<'_>,
+    ) -> Option<Either<&'a Layout, LayoutHint>> {
+        let mut total_min_size = 0;
+        let mut max_align = 1;
+        let mut uninhabited_part = false;
+
+        // If it is a new-type struct, we should be able to get a layout.
```
Nadrieril: There's no guarantee that a newtype struct has the same layout as its contents (unless `repr(transparent)`). In fact it's easy to change that:
```rust
#[repr(align(64))]
struct S(u8);
```

charon/src/ast/layout_computer.rs
```diff
@@ -0,0 +1,538 @@
...
+                        total_min_size += h.min_size;
+                        max_align = max(max_align, h.min_alignment);
+                    }
+                    None => (),
+                }
+            }
+            Some(Either::Right(LayoutHint {
+                min_size: total_min_size,
+                min_alignment: max_align,
+                possibly_uninhabited: uninhabited_part,
```
Nadrieril: For enum variants, if there's an uninhabited field the whole variant may have size zero (https://github.com/rust-lang/unsafe-code-guidelines/issues/443). Hence `min_size` should be zero if `possibly_uninhabited` is true and we're in an enum variant. I don't know the rule for alignment.

Nadrieril: Also, if `repr(packed)` then the alignment can be wrong so `min_alignment` should be `1` I think.

firefighterduck:
> it seems that rustc can figure out that `size_of::<&T>()` is 8 if `T` is sized

I've added a sizedness check already, so that we can also get this information in charon.

> we could get generic layouts to work by asking rustc to evaluate `size_of` and `offset_of` constants and catching errors appropriately.

That's a good point. In my PR, I've tried to add facilities to charon to get layout information **after** the translation to (U)LLBC, thus invoking rustc isnt't really an option. However, we could also think about adding something similar to the translation to also be able to query rustc.

charon/src/ast/types_utils.rs
```diff
@@ -1027,7 +1042,20 @@ pub trait TyVisitable: Sized + AstVisitable {
...
 
     fn substitute_with_self(mut self, generics: &GenericArgs, self_ref: &TraitRefKind) -> Self {
-        let _ = self.visit_vars(&mut SubstVisitor::new(generics, self_ref));
+        let _ = self.visit_vars(&mut SubstVisitor::new(generics, self_ref, false));
+        self
+    }
+
+    fn substitute_frees(self, generics: &GenericArgs) -> Self {
+        self.substitute_with_self_frees(generics, &TraitRefKind::SelfId)
+    }
```
Nadrieril, line 1119: Oh yeah, that reminds me I should make `Free` variables optional, they were added to ease the transition to De Bruijn indices. Could you document this function to explain what it does?

firefighterduck:
> You seem to say you'd want the layout of a given `Ty`?

Exactly, since that is what tools actually need (some more, some less). They need to have information about all types that occur in the program, not just the ones available from definitions. And it's not just/primarily the layout, oftentimes some related properties are of great interest, e.g. whether a type is a ZST, a DST, whether it's FFI-safe, whether it's uninhabited, whether two types can be transmuted into each other, etc. Some of these properties can be approximated from the type alone, but it would be more convenient to have all such information in one place such that all downstream tools can easily access it.

PR #801, opened 2025-08-11
FIX: Use of `annotate-snippets` crate after recent update in upstream

firefighterduck:
[Last week](https://github.com/rust-lang/annotate-snippets-rs/commit/177666be65c9a1f2392908e194c6ced14ec7b3b2) an update to the `annotate-snippets-rs` crate that Charon uses removed a function that is called in Charon. To make sure that Charon can be compiled with up-to date upstream crate versions, I've removed the use of this function.

PS: As a small drive-by I've also fixed a doc problem.

### DIFF
--- a/charon/Cargo.lock
+++ b/charon/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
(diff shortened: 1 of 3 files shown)


Nadrieril:
Ty!

PR #814, opened 2025-09-15
Translate `dyn Trait` method shims

ssyram:
This PR is split from #762 . It contains the part of the implementation for supporting vtable instances. For each *concrete* casting from `&T` to `&dyn Trait`, a global `static {impl Trait for T}::{vtable}` will be generated, which initialises the method fields with the method shims --- notably, when one calls a `method` from `&dyn Trait`, the receiver is `&dyn Trait` instead of the concrete type `&T`. This requires a bridging method that uses a `Concretize` cast to cast back from `&dyn Trait` to `&T` and then call the correct `method` within the user-defined `impl Trait for T`.

Note that this PR does not handle the virtual implementations, e.g., closures --- there is no explicit `impl Fn for CLOSURE` that is defined by the user. This will be fixed in the up-coming PR split from #762 .

Also, in this PR, the metadata fields of the vtable instances, i.e., the `size`, `align` and `drop` fields are not translated and are currently `Opaque`. This will also be partly alleviated in the up-coming PR from #762 .

ci: use https://github.com/AeneasVerif/aeneas/pull/619
ci: use https://github.com/AeneasVerif/eurydice/pull/294

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_crate.rs
+++ b/charon/src/bin/charon-driver/translate/translate_crate.rs
@@ -504,6 +504,7 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
         &mut self,
         span: Span,
         item: &hax::ItemRef,
+        kind: TransItemSourceKind,
     ) -> Result<MaybeBuiltinFunDeclRef, Error> {
         match self.recognize_builtin_fun(item)? {
             Some(id) => {
@@ -514,7 +515,7 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
                     generics: Box::new(generics),
                 })
             }
-            None => self.translate_item(span, item, TransItemSourceKind::Fun),
+            None => self.translate_item(span, item, kind),
         }
     }
 
@@ -525,8 +526,9 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
         &mut self,
         span: Span,
         item: &hax::ItemRef,
+        kind: TransItemSourceKind,
     ) -> Result<RegionBinder<FnPtr>, Error> {
-        let fun_item = self.translate_fun_item(span, item)?;
+        let fun_item = self.translate_fun_item(span, item, kind)?;
         let late_bound = match self.hax_def(item)?.kind() {
             hax::FullDefKind::Fn { sig, .. } | hax::FullDefKind::AssocFn { sig, .. } => {
                 Some(sig.as_ref().rebind(()))
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -1,10 +1,13 @@
+use itertools::Itertools;
+use std::mem;
+
 use super::{
     translate_crate::TransItemSourceKind, translate_ctx::*, translate_generics::BindingLevel,
 };
-
+use charon_lib::formatter::IntoFormatter;
 use charon_lib::ids::Vector;
+use charon_lib::pretty::FmtWithCtx;
 use charon_lib::ullbc_ast::*;
-use itertools::Itertools;
 
 fn dummy_public_attr_info() -> AttrInfo {
     AttrInfo {
@@ -317,11 +320,23 @@ impl ItemTransCtx<'_, '_> {
         Ok(fields)
     }
 
+    /// This is a temporary check until we support `dyn Trait` with `--monomorphize`.
+    pub(crate) fn check_no_monomorphize(&self, span: Span) -> Result<(), Error> {
+        if self.monomorphize() {
+            raise_error!(
+                self,
+                span,
+                "`dyn Trait` is not yet supported with `--monomorphize`; \
+                use `--monomorphize-conservative` instead"
+            )
+        }
+        Ok(())
+    }
+
     /// Construct the type of the vtable for this trait.
     ///
     /// It's a struct that has for generics the generics of the trait + one parameter for each
     /// associated type of the trait and its parents.
-    /// TODO(dyn): add the associated types.
     ///
     /// struct TraitVTable<TraitArgs.., AssocTys..> {
     ///   size: usize,
@@ -347,6 +362,7 @@ impl ItemTransCtx<'_, '_> {
                 for a non-dyn-compatible trait"
             );
         }
+        self.check_no_monomorphize(span)?;
 
         self.translate_def_generics(span, trait_def)?;
 
@@ -503,6 +519,7 @@ impl ItemTransCtx<'_, '_> {
     ) -> Result<GlobalDecl, Error> {
         let span = item_meta.span;
         self.translate_def_generics(span, impl_def)?;
+        self.check_no_monomorphize(span)?;
 
         let (impl_ref, _, vtable_struct_ref) =
             self.get_vtable_instance_info(span, impl_def, impl_kind)?;
@@ -547,14 +564,15 @@ impl ItemTransCtx<'_, '_> {
                 ..
             } => {
                 // The method is vtable safe so it has no generics, hence we can reuse the impl
-                // generics.
+                // generics -- the lifetime binder will be added as `Erased` in `translate_fn_ptr`.
                 let item_ref = impl_def.this().with_def_id(self.hax_state(), item_def_id);
-                let shim_ref =
-                    self.translate_item(span, &item_ref, TransItemSourceKind::VTableMethod)?;
+                let shim_ref = self
+                    .translate_fn_ptr(span, &item_ref, TransItemSourceKind::VTableMethod)?
+                    .erase();
                 ConstantExprKind::FnPtr(shim_ref)
             }
             hax::ImplAssocItemValue::DefaultedFn { .. } => ConstantExprKind::Opaque(
-                "shim for provided methods \
+                "shim for default methods \
                     aren't yet supported"
                     .to_string(),
             ),
@@ -721,6 +739,26 @@ impl ItemTransCtx<'_, '_> {
         }))
     }
 
+    fn generate_concretization(
+        &mut self,
+        span: Span,
+        statements: &mut Vec<Statement>,
+        shim_self: &Place,
+        target_self: &Place,
+    ) -> Result<(), Error> {
+        let rval = Rvalue::UnaryOp(
+            UnOp::Cast(CastKind::Concretize(
+                shim_self.ty().clone(),
+                target_self.ty().clone(),
+            )),
+            Operand::Move(shim_self.clone()),
+        );
+        let stmt = StatementKind::Assign(target_self.clone(), rval);
+        statements.push(Statement::new(span, stmt));
+
+        Ok(())
+    }
+
     pub(crate) fn translate_vtable_instance_init(
         mut self,
         init_func_id: FunDeclId,
@@ -730,6 +768,7 @@ impl ItemTransCtx<'_, '_> {
     ) -> Result<FunDecl, Error> {
         let span = item_meta.span;
         self.translate_def_generics(span, impl_def)?;
+        self.check_no_monomorphize(span)?;
 
         let (impl_ref, _, vtable_struct_ref) =
             self.get_vtable_instance_info(span, impl_def, impl_kind)?;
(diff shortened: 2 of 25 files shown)


charon-ml/src/PrintTypes.ml
```diff
@@ -340,7 +340,22 @@ and trait_instance_id_to_string (env : 'a fmt_env)
...
+  | ParentClause (tref, clause_id) ->
+      let inst_id = trait_ref_to_string env tref in
+      let clause_id = trait_clause_id_to_string env clause_id in
+      "parent(" ^ inst_id ^ ")::" ^ clause_id
+  | Dyn ->
+      let trait =
+        region_binder_to_string trait_decl_ref_to_string env tr.trait_decl_ref
+      in
+      "dyn(" ^ trait ^ ")"
+  | UnknownTrait msg -> "UNKNOWN(" ^ msg ^ ")"
```
Nadrieril: Looks like you reverted some changes I made. Please re-read your PR before submitting it to avoid such issues.

charon/src/ast/types/vars.rs
```diff
@@ -137,7 +141,7 @@ pub struct ConstGenericVar {
 
 /// A trait predicate in a signature, of the form `Type: Trait<Args>`. This functions like a
 /// variable binder, to which variables of the form `TraitRefKind::Clause` can refer to.
-#[derive(Debug, Clone, Serialize, Deserialize, Drive, DriveMut)]
+#[derive(Debug, Clone, Serialize, Deserialize, Drive, DriveMut, PartialOrd, Ord)]
```
Nadrieril: Where does the need for all these `PartialOrd` impls come from? It doesn't seem meaningful to order trait clauses or types etc

charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -69,7 +69,10 @@ pub enum TransItemSourceKind {
     /// Shim function to store a method in a vtable; give a method with `self: Ptr<Self>` argument,
     /// this takes a `Ptr<dyn Trait>` and forwards to the method. The `DefId` refers to the method
     /// implementation.
-    VTableMethod,
+    ///
+    /// For technical reasons, it takes the `self_type` and `dyn_self_type`:
+    /// the former is the type being implemented now while the latter is the `dyn Trait<...>` type.
+    VTableMethod(Ty, Ty),
```
Nadrieril: The type being implemented is already available from the `DefId` of the impl. I think it's a mistake to include it here, particularly since we only want to define the most generic possible version of this method shim.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -529,8 +531,11 @@ impl ItemTransCtx<'_, '_> {
         &mut self,
         span: Span,
         impl_def: &hax::FullDef,
+        self_ty: &Ty,
+        dyn_self: &Ty,
         item: &hax::ImplAssocItem,
-        mut mk_field: impl FnMut(ConstantExprKind),
+        mut next_ty: impl FnMut() -> Ty,
+        mut mk_field: impl FnMut(ConstantExprKind, Ty),
```
Nadrieril: Please document what `next_ty` is meant to do

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: Please remove these two arguments. I'd rather recompute them than risk duplicating shims or getting generics wrong.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -540,27 +545,45 @@ impl ItemTransCtx<'_, '_> {
...
+                let mut shim_ref: FnPtr = self.translate_item(
+                    span,
+                    &item_ref,
+                    TransItemSourceKind::VTableMethod(self_ty.clone(), dyn_self.clone()),
+                )?;
+                let ty = next_ty();
+                match ty.kind() {
+                    TyKind::FnPtr(sig) => {
+                        let regions = shim_ref.generics.regions;
+                        // Add a bunch of erased regions to avoid the type mismatch.
```
Nadrieril: That's a terrible explanation for why we're doing this. I presume the issue is that late-bound regions are missing from `shim_ref` and we're adding them here. If that's the case you should be using `translate_fn_ptr` instead (I'll add a line of documentation to explain that in my next PR). You'll just have to change it to take a `TransItemSourceKind` as input and maybe fix it if it doesn't find the signature.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -724,6 +769,64 @@ impl ItemTransCtx<'_, '_> {
         }))
     }
 
+    fn check_concretization_ty_match(
+        &self,
+        span: Span,
+        src_ty: &Ty,
+        tar_ty: &Ty,
+    ) -> Result<(), Error> {
```
Nadrieril: Could you move this to the `check_generics` pass so we get legible errors?

charon/tests/ui/monomorphization/dyn-trait.out
```diff
@@ -88,32 +88,18 @@ note: the error occurred when translating `core::marker::MetaSized::<type_error(
...
-note: the error occurred when translating `core::fmt::Display::{vtable}::<alloc::string::String>`, which is (transitively) used at the following location(s):
-  --> tests/ui/monomorphization/dyn-trait.rs:12:27
-   |
-12 |     let _ = dyn_to_string(&str);
-   |                           ----
 
-thread 'rustc' panicked at src/bin/charon-driver/translate/translate_trait_objects.rs:375:17:
-incorrect `dyn_self`
+thread 'rustc' panicked at /Users/ssyram/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/index_vec-0.1.4/src/indexing.rs:128:10:
+index out of bounds: the len is 0 but the index is 0
```
Nadrieril: Could you add a check to make sure we don't try to translate vtables if `--monomorphize` is on? Having a clear error is important.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: After my other comments I think you'll be able to remove it actually.

charon-ml/src/PrintTypes.ml
ssyram: Yes, it was for Eurydice compatibility. Will resolve.

charon/src/ast/types/vars.rs
ssyram: It was added by need, I will re-check.

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: But how can we get the `dyn_self_type`? I will have a re-check.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: OK, will do.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Well, actually this part is something that I'm not fully understanding. This may be the late-bound issue as you discussed... So, concretely how should I fix that?

charon/src/ast/types/vars.rs
Nadrieril: If it's possible to instead remove some existing `derive(PartialOrd)` clauses, I'm ok with that.

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: With a bunch of substitutions probably. We get it from the appropriate trait decl and substitute.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: I just explained exactly what you have to do in the previous message

ssyram: Not yet concrete enough to understand... But maybe wait for the next version in main to see what to do.

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: To rebuild that is too complex and error-prone... Especially for assoc-type? That was why I want to keep the info there. Also, this should not lead to duplication of shims? Because it is the case that we should construct a vtable-shim-method for a specific pair of concrete type `T` and a dyn-self-type `dyn Trait<...>`. Also, the `DefId` is for the *function* ID which implements for concrete type `T`. So the types are providing the correct info, right?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: Ok, here's more detailed:
- add a `kind: TransItemSourceKind` parameter to `translate_fn_ptr` that it will pass to `translate_fun_item` then to `translate_item`. update call locations of `translate_fn_ptr` to preserve current behavior.
- use `translate_fn_ptr` instead of `translate_item`;
- see if it works.

charon/tests/ui/monomorphization/dyn-trait.out
ssyram: OK, will do.

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: > Because it is the case that we should construct a vtable-shim-method for a specific pair of concrete type T and a dyn-self-type dyn Trait<...>.

Not exactly no, we should build one shim for each method implementation, because we want one vtable value for each trait implementation. This does mean that if we have a `dyn Trait + Send` we should first cast it to `dyn Trait` before we can call a method shim on it. But that's ok, it's a no-op.

> Also, this should not lead to duplication of shims?

No because we'll have the minimum amount of shims required to build the vtable values.

> Also, the DefId is for the function ID which implements for concrete type T.

Yes exactly, it's for a method implementation coming from a particular trait impl.

> So the types are providing the correct info, right?

Well, they are not correct right now. You're using `self_ty` and `dyn_self` which are types valid within the `TraitImpl` and using them within a method, which does not have the same generic parameters (see e.g. https://github.com/AeneasVerif/charon/issues/513). The rule in charon is that when you move a value between two items with different generic parameters, you must do a substitution.

Anyway, the easiest solution to "this substitution is too complicated" is to ask rustc to do it for us. Should be easy to change `vtable_safe: bool` into `vtable_sig: Option<PolyFnSig>` in `hax::FullDef::AssocFn`.

charon-ml/src/PrintTypes.ml
ssyram: Synchronized.

charon/src/ast/types/vars.rs
ssyram: OK, I've checked that this is related to the `Ty` parameters for `VTableMethod` below.

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: OK, so this place should be changed after `Hax` has the change?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: It works, thanks.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Fixed.

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: Well someone needs to do the hax change (and I'm away starting tomorrow), but yes

charon/tests/ui/monomorphization/dyn-trait.out
ssyram: Done.

charon/src/bin/charon-driver/translate/translate_crate.rs
ssyram: Added in AeneasVerif/hax#4 .

charon/src/ast/types/vars.rs
ssyram: OK.

Nadrieril:
I renamed `dyn_sig` to `vtable_sig` since, and am merging that small change here: https://github.com/AeneasVerif/charon/pull/829. Once that is merged please update your PR

ssyram:
OK, updated.

PR #820, opened 2025-09-23
Mono Name-Matcher

ssyram:
This PR enhanced the name-matcher mechanism to work also for the mono codes. The key is that when the newly added `match_mono` switch is on, the program now tries to extract out the generics stored within the name and match against this generics. This is a key component to ensure the migration to mono LLBC as in AeneasVerif/eurydice#275 .
 
Notable issue: now there is a discrepency between the parsers in Rust and OCaml. More concretely, the string used in Eurydice, that: `core::slice::index::{core::ops::index::Index<[@T], @I>}::index<'_, @, core::ops::range::Range<usize>, [@]>` cannot be recognised by `Pattern::parse` in Rust.

### DIFF
--- a/charon/rustfmt.toml
+++ b/charon/rustfmt.toml
@@ -1,0 +1,1 @@
+style_edition = "2024"


Nadrieril:
Please remove that option and make the name matchers always work wirh monomorphized types please

charon/tests/ui/ml-name-matcher-tests.rs
```diff
@@ -79,3 +79,21 @@ fn funs_with_disambiguator(b : bool) ->u32 {
...
+    fn create(item: T) -> Self {
+        Self { item }
+    }
+}
+
+#[pattern::pass("test_crate::mono_usage")]
+fn mono_usage() {
+    let _container1 = MonoContainer::create(42i32);
+    let _container2 = MonoContainer::create("test");
+}
```
Nadrieril, line 102: These name matcher tests are not currently run with `--monomorphize` on. Could you make `rust-mono-mane-matcher-tests.rs` and `ml-mono-name-matcher-tests.rs` that contain the subset of these files that work in mono mode? If by miracle everything works in mono mode then we can keep one file and run it in both modes.

ssyram: OK. Do you have any suggestions on how to configure Charon to run on mono mode? I didn't see the switches to enable mono translation.

Nadrieril: In `charon/tests/test-rust-name-matcher.rs` you can see a call to `translate_rust_text`. You can pass it a different file and the `--monomorphize` option. On the ocaml side we're relying on the ui suite to generate the llbc file, so at the top of `ml-mono-name-matcher-tests.rs` you'd just add `//@ charon-args --monomorphize`, and in `Tests.ml` you add a new `Test_NameMatcher.run_tests` line.

ssyram: Done.

PR #836, opened 2025-10-05
Insert `StorageLive`s after all transforms

N1ark:
In #828 the insertion of `StorageLive`s was moved to one of the first passes, which meant some locals added in further transforms (I am not sure which, quite frankly) were missing a `StorageLive`.
I moved the transform back to the end -- from the diff of the tests, some locals now seem to have a wider scope than before -- this is sound, though suboptimal, but I'm not sure I understand how that is the case :) if you have any idea I'm happy to fix it, but otherwise I think the current state of things is also ok


N1ark:
Aha! So the diff on the tests is basically the inverse as what happens in 31fc2af800eb1fcf5675fb8e8dab91812c618a91, so it's some ordering that doesn't work, but I'm quite unsure what would be the correct ordering :p

Nadrieril:
Yeah no I definitely missed something, sorry about that. I'll look into that tomorrow.

charon/src/transform/simplify_output/inline_promoted_consts.rs
```diff
@@ -47,6 +47,16 @@ impl UllbcPass for Transform {
...
 
+                            // The init function of a global assumes the return place is live;
+                            // this is not the case once we inline it
+                            inner_body.body[0].statements.insert(
+                                0,
+                                Statement::new(
+                                    Span::dummy(),
+                                    StatementKind::StorageLive(LocalId::ZERO),
+                                ),
+                            );
```
Nadrieril, line 58: Oh wow, well spotted

Nadrieril:
Amazing! Yes I'd love such a check. Thanks for the PR!

PR #837, opened 2025-10-06
Add `#[repr(...)]` options to `TypeDecl`

firefighterduck:
Adds information about the representation options to type declarations. As a drive by, also fixes some doc warnings.

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -402,10 +402,25 @@ pub enum PtrMetadata {
     InheritFrom(Ty),
 }
 
+/// Describes which layout algorithm is used for representing the corresponding type.
+/// Depends on the `#[repr(...)]` used.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum ReprAlgorithm {
+    /// The default layout algorithm. Used without an explicit `ŗepr` or for `repr(Rust)`.
+    Rust,
+    /// The C layout algorithm as enforced by `repr(C)`.
+    C,
+}
+
+/// Describes modifiers to the alignment and packing of the corresponding type.
+/// Represents `repr(align(n))` and `repr(packed(n))`.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum AlignmentModifier {
+    Align(ByteCount),
+    Pack(ByteCount),
+}
+
 /// The representation options as annotated by the user.
-///
-/// If all are false/None, then this is equivalent to `#[repr(Rust)]`.
-/// Some combinations are ruled out by the compiler, e.g. align and pack.
 ///
 /// NOTE: This does not include less common/unstable representations such as `#[repr(simd)]`
 /// or the compiler internal `#[repr(linear)]`. Similarly, enum discriminant representations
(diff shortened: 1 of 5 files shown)


charon/src/ast/types.rs
```diff
@@ -403,6 +402,24 @@ pub enum PtrMetadata {
...
+/// or the compiler internal `#[repr(linear)]`. Similarly, enum discriminant representations
+/// are encoded in [`Variant::discriminant`] and [`DiscriminantLayout`] instead.
+/// This only stores whether the discriminant type was derived from an explicit annotation.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ReprOptions {
+    pub align: Option<ByteCount>,
+    pub pack: Option<ByteCount>,
+    pub c: bool,
+    pub transparent: bool,
+    pub explicit_discr_type: bool,
```
Nadrieril, line 434: What's the use of knowing whether the discriminant type was explicit? I guess it's that this makes some casts legal? Could you add a doc comment explaining what to make of this field?

Nadrieril: Actually the whole of `ReprOptions` is about what kinds of casts/transmutes are legal isn't it

charon/src/ast/types_utils.rs
```diff
@@ -1311,6 +1315,16 @@ impl Layout {
...
 }
 
+impl ReprOptions {
+    /// Whether this representation options guarantee a fixed
+    /// field ordering for the type.
+    ///
+    /// Cf. `rustc_abi::ReprOptions::inhibit_struct_field_reordering`.
+    pub fn guarantees_fixed_field_order(&self) -> bool {
+        self.c || self.explicit_discr_type
+    }
```
Nadrieril, line 1332: Instead of the internal docs, would you mind referencing https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.c.struct and https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.primitive.adt .

charon/src/ast/types.rs
N1ark, line 434: Maybe I'm bikeshedding, but some of these are incompatible with others -- it would be nice to have these constrained by the type.
```rust
enum Repr {
    Rust,
    C,
}
enum Modifier {
    Pack(ByteCount),
    Align(ByteCount),
}
struct ReprOptions {
    repr: Repr,
    modifier: Option<Modifier>,
    transparent: bool,
    explicit_discr_type: bool,
}
```

firefighterduck: It's about casts/transmutes, but also about layout guarantees (e.g. `align` and `packed` provide a lower and upper bound, respectively, on the actual alignment). In this context, knowing that the discriminant type was fixed via a `repr(...)` implies that the enum's representation and layout are fixed (especially for enums without fields), regardless of the target platform or compiler version.

charon/src/ast/types_utils.rs
firefighterduck: In this case it certainly makes sense to refer to the reference. Additionally, we might want to add some remark that this must also consider `repr(simd)` and `repr(linear)` if we add these later. This bit of information is unfortunately not described in the reference (yet).

Edit: Done.

charon/src/ast/types.rs
firefighterduck: Done

Nadrieril: Ok interesting I didn't know this also affected layout

PR #838, opened 2025-10-08
Fix `NameMatcherMap` for `PWild`

N1ark:
`NameMatcherMap` didn't handle wildcard `_` patterns correctly. 
e.g. a map containing `foo::_::bar` and `foo::_::baz` wouldn't match `foo::thing::bar` or `foo::{...}::bar`.

sidenote: we really should redo the map to just use `pattern_elem` in the nodes rather than `pattern` :')


Nadrieril:
Ah yes, the `PWild` implementation was thrown together a bit fast x) Ty!

Issue #819, opened 2025-09-17
Bug: Charon confuses drop glue and `Drop` implementations


Nadrieril:
Here's how dropping happens in Rust:
- For every value that goes out of scope without being moved out, the (post-drop-elaboration) MIR contains a `TerminatorKind::Drop(place)`;
- The meaning of this terminator is to call `core::ptr::drop_in_place<T>(&raw mut place)`;
- `drop_in_place` first calls the `Drop` impl of `T` if there is one;
- Then it runs the "drop glue", which is MIR code that drops (using `TerminatorKind::Drop`) all the fields of `T`.

Note how a `Drop::drop` impl must therefore _not_ drop any of its fields (that requires unsafe anyway). When I wrote the code that handle drops I had confused these things and put the drop glue inside the `Drop::drop` impls I generate. That's incorrect; however I still want to use the trait system to track which code to run on drop for generic types.

My current idea would be to add a fake `drop_in_place(*mut self)` method to the `Drop` trait and put each type's drop glue in there. Then the MIR `drop` would just mean "call `Drop::drop_in_place` on that place". That may make things easier for `dyn` vtables too.

PR #839, opened 2025-10-08
Dyn method dispatch

ssyram:
Following #814 , this PR is further split from #762, which includes the part of vtable support for the correct dispatch of virtual methods. Now, calling methods of &dyn Trait will be translated to finding the corresponding function pointer from the vtable (fetched by ptr_metadata operation) and then call the function pointer.

Specifically, calling virtual methods is turned from `<dyn Trait as Trait>::method(receiver, args...)` to something like:
```
let vtable = receiver.ptr_metadata;
let method_fn_ptr = vtable.method_index;
method_fn_ptr(receiver, args...)
```

This is reached by adding a new pass: `normalize::transform_dyn_trait_calls`.


charon/src/ast/expressions_utils.rs
```diff
@@ -244,3 +237,12 @@ impl UnOp {
...
 }
+
+impl Operand {
+    pub fn ty(&self) -> &Ty {
+        match self {
+            Operand::Copy(place) | Operand::Move(place) => &place.ty,
+            Operand::Const(c) => &c.ty,
+        }
+    }
+}
```
Nadrieril: why did you move this? mind leaving it where it is?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -297,14 +297,21 @@ impl ItemTransCtx<'_, '_> {
         mk_field("size".into(), usize_ty());
         // Field: `align: usize`
         mk_field("align".into(), usize_ty());
-        // Field: `drop: fn(*mut Self)`
+        // Field: `drop: fn<'0>(&'0 mut Self)` -- `Self` is just a placeholder, will be dynified below.
```
Nadrieril: As I've said before, I insist that this must take a raw pointer, because this is a call to `ptr::drop_in_place` which takes a raw pointer.

charon/src/transform/mod.rs
```diff
@@ -112,6 +113,9 @@ pub static INITIAL_CLEANUP_PASSES: &[Pass] = &[
 
 /// Body cleanup passes on the ullbc.
 pub static ULLBC_PASSES: &[Pass] = &[
+    // Transform dyn trait method calls to vtable function pointer calls
+    // This should be early to handle the calls before other transformations
+    UnstructuredBody(&normalize::transform_dyn_trait_calls::Transform),
```
Nadrieril: nit: I would put it just below `skip_trait_refs_when_known` because it feels very similar

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: To be fair I made that mistake myself before, and I still need to fix that: https://github.com/AeneasVerif/charon/issues/819.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+//! is transformed from:
+//! ```text
+//! @0 := call TraitMethod::method(x, args)
+//! ```
+//!
+//! to:
+//! ```text
+//! vtable@9 := ptr_metadata(move (@receiver))              // Extract vtable pointer
+//! method_ptr@8 := copy (((*vtable@9).method_check))       // Get method from vtable
+//! @0 := (move method_ptr@8)(move (@receiver), move (@args)) // Call through function pointer
```
Nadrieril: Would you mind fixing this comment to use the new `ptr_metadata` representation?

charon/src/ast/expressions_utils.rs
ssyram: Sure. If you're curious about the reason: I implemented one such function myself before in the original PR. So there became two duplicated definitions when I rebase.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+        let FnOperand::Regular(fn_ptr) = &call.func else {
+            return None; // Not a regular function call
+        };
+
+        let FnPtrKind::Trait(trait_ref, method_name, _) = &fn_ptr.kind else {
+            return None; // Not a trait method call
+        };
+
+        match &trait_ref.kind {
+            TraitRefKind::Dyn => Some((trait_ref.clone(), method_name.clone())),
```
Nadrieril: nit: you could return references here, why clone?

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+                        );
+                    }
+                }
+            }
+        }
+    }
+
+    fn name(&self) -> &str {
+        "TransformDynTraitCalls"
+    }
```
Nadrieril: That's not needed, the method is implemented automatically

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: OK.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+                    Ok(None) => {
+                        trace!("No transformation needed for this call");
+                    }
+                    Err(e) => {
+                        register_error!(
+                            ctx,
+                            span,
+                            "Failed to transform dynamic trait call: {:?}",
+                            e
+                        );
```
Nadrieril: That would show the error twice, since the error was obtained from a `raise_error!` already; please don't. In fact please remove the whole match, we don't need to have "no transformation needed" emitted for every single function call.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+}
+
+pub struct Transform;
+
+impl UllbcPass for Transform {
+    fn transform_body(&self, ctx: &mut TransformCtx, body: &mut ExprBody) {
+        trace!(
+            "TransformDynTraitCalls: Processing body with {} blocks",
+            body.body.iter().count()
+        );
```
Nadrieril: I don't see the point of displaying the number of blocks here.

charon/src/transform/mod.rs
ssyram: Sure.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+
+        for (block_id, block) in body.body.iter_mut().enumerate() {
+            let block_id = BlockId::new(block_id);
```
Nadrieril:
```suggestion
        for (block_id, block) in body.body.iter_indexed() {
```

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: Of course.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+                field_id,
+            ));
+
+        // 6. Transform the original call to use the function pointer
+        call.func = FnOperand::Move(method_ptr_place);
+
+        trace!(
+            "Generated {} new statements for vtable dispatch",
+            self.statements.len()
+        );
```
Nadrieril: Who needs this? Please ask you AI to not add traces that wouldn't help anyone trying to debug this.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+        if call.args.is_empty() {
+            raise_error!(self.ctx, self.span, "Dyn trait call has no arguments!");
+        }
+        let dyn_trait_place = match &call.args[0] {
+            Operand::Copy(place) => place,
+            Operand::Move(place) => place,
+            Operand::Const(_) => {
+                // Constants should not be dyn trait objects in practice
+                return Ok(None);
+            }
```
Nadrieril, line 98: Please emit an error in this case since this isn't supposed to happen.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: Will delete.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: Will delete.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK, that was for debugging before. But seems outdated now.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+};
+
+/// Transformer for dynamic trait calls that holds common state and provides methods
+/// for the transformation operations.
+struct DynTraitCallTransformer<'a> {
+    ctx: &'a TransformCtx,
+    span: Span,
+    statements: &'a mut Vec<Statement>,
+    locals: &'a mut Locals,
+}
```
Nadrieril: nit: I'd prefer to use a `UllbcStatementTransformCtx` here instead of making your own

charon/src/ast/expressions_utils.rs
Nadrieril: oh, fun :D

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: Will try -- when the PR was made, I think the `UllbcStatementTransformCtx` was not there.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,474 @@
...
+            };
+            // Push it
+            generics.types.push(assoc_ty);
+        }
+
+        // 5. Return vtable ref's ID with correct generics
+        Ok(TypeDeclRef {
+            id: vtable_ref.id,
+            generics,
+        })
```
Nadrieril: You should never need to manipulate things by hand like this. Charon is structured so that substitutions do the right thing:
```suggestion
        let trait_pred = trait_ref.trait_decl_ref.clone().erase();
        let vtable_ref = vtable_ref
            .clone()
            .substitute_with_self(&trait_pred.generics, &trait_ref.kind);
        Ok(vtable_ref)
```
There is one problem with that is that you'll get unnormalized types. So we should 1. move the `expand_associated_types` pass after this one (it's the one that normalized associated types), and 2. I should fix it to work with `dyn` because it doesn't yet oops

ssyram: OK.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
Nadrieril: Yeah, I added it recently to make better use of your `BodyTransformCtx` idea

ssyram: OK, done.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,297 @@
...
+            kind: StatementKind::StorageLive(vtable_place.as_local().unwrap()),
+            comments_before: vec![],
+        });
+
+        let method_ptr_place = self.locals.new_var(None, method_ptr_ty.clone());
+        self.statements.push(Statement {
+            span: self.span,
+            kind: StatementKind::StorageLive(method_ptr_place.as_local().unwrap()),
+            comments_before: vec![],
+        });
```
Nadrieril: The main point of `UllbcStatementTransformCtx` is to simplify this:
```suggestion
        let vtable_place = self.fresh_var(None, vtable_ty);
        let method_ptr_place = self.fresh_var(None, method_ptr_ty.clone());
```

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,297 @@
...
+            .clone()
+            .project(ProjectionElem::PtrMetadata, vtable_place.ty().clone());
+        Statement {
+            span: self.span,
+            kind: StatementKind::Assign(
+                vtable_place.clone(),
+                Rvalue::Use(Operand::Copy(ptr_metadata_place)),
+            ),
+            comments_before: vec!["Extract vtable pointer from dyn trait object".to_string()],
+        }
```
Nadrieril: Same here, you can use `self.insert_assn_stmt` instead of constructing the `Statement` by hand. Btw this comment should be removed, I only want user comments in there.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn generate_method_pointer_extraction(
+        &self,
+        method_ptr_place: &Place,
+        vtable_place: &Place,
+        field_id: FieldId,
+    ) -> Statement {
+        // Create vtable dereference: *vtable
+        let vtable_deref_place = Place {
+            kind: PlaceKind::Projection(Box::new(vtable_place.clone()), ProjectionElem::Deref),
+            ty: Ty::mk_unit(), // placeholder type
```
Nadrieril: This incorrect type is not ok, please add the real type.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -0,0 +1,297 @@
...
+        };
+
+        // Create method field projection: (*vtable).method_field
+        let method_field_place = Place {
+            kind: PlaceKind::Projection(
+                Box::new(vtable_deref_place),
+                ProjectionElem::Field(
+                    FieldProjKind::Adt(TypeDeclId::new(0), None), // Placeholder vtable type ID
```
Nadrieril: Same here, please don't leave incorrect information in the AST? If this is your AI model again, please review more thoroughly its output. While you're at it it's nicer to use the helper method `.project` instead of building the place by hand.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK, thanks for pointing out.

charon/src/transform/normalize/transform_dyn_trait_calls.rs
ssyram: OK, fixed. Sorry about that.

PR #842, opened 2025-10-09
Translate All involved Vtable Instances & Structs

ssyram:
This issue is found in Eurydice. Now, when translating the vtable structs / instances, the involved content (involved super vtable struct / instance) may not be translated and hence when translating the struct / instance itself in Eurydice, it triggers the definition not found problem.

This PR fixes the problem by enforcing the translation of all super trait structs / instances.

Please merge #841 first as this will trigger the double translation described there.

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -525,7 +525,7 @@ impl ItemTransCtx<'_, '_> {
             _ => unreachable!(),
         };
         let vtable_struct_ref = self
-            .translate_vtable_struct_ref_no_enqueue(span, implemented_trait)?
+            .translate_vtable_struct_ref(span, implemented_trait)?
             .expect("trait should be dyn-compatible");
         let implemented_trait = self.translate_trait_decl_ref(span, implemented_trait)?;
         let impl_ref = self.translate_item(
@@ -655,7 +655,7 @@ impl ItemTransCtx<'_, '_> {
 
             let vtable_def_ref = self
                 .translate_region_binder(span, &impl_expr.r#trait, |ctx, tref| {
-                    ctx.translate_vtable_struct_ref_no_enqueue(span, tref)
+                    ctx.translate_vtable_struct_ref(span, tref)
                 })?
                 .erase()
                 .expect("parent trait should be dyn compatible");
--- a/charon/src/bin/charon-driver/translate/translate_types.rs
+++ b/charon/src/bin/charon-driver/translate/translate_types.rs
@@ -393,11 +393,7 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
                                     unreachable!()
                                 };
                                 Ok(ctx
-                                    .translate_vtable_struct_ref_maybe_enqueue(
-                                        false,
-                                        span,
-                                        &trait_predicate.trait_ref,
-                                    )?
+                                    .translate_vtable_struct_ref(span, &trait_predicate.trait_ref)?
                                     .unwrap())
                             },
                         )?


charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -185,6 +185,7 @@ impl ItemTransCtx<'_, '_> {
     /// Given a trait ref, return a reference to its vtable struct, if it is dyn compatible.
     pub fn translate_vtable_struct_ref(
         &mut self,
+        enforce_enqueue: bool,
         span: Span,
         tref: &hax::TraitRef,
     ) -> Result<Option<TypeDeclRef>, Error> {
```
Nadrieril, line 207: I agree with the idea of this change. Instead of a boolean though, I'd prefer if you renamed the original function to `translate_vtable_struct_ref_no_enqueue` and make a `translate_vtable_struct_ref` that enqueues (because not enqueuing should be the exception to avoid exactly the error you're fixing). And same for the other function of course.

ssyram: This produces duplications in code? Maybe I should add two more functions one calls this abstract one with `true` one `false`?

Nadrieril: The enqueing one will calll the `no_enqueue` one of course, no duplication needed

ssyram: But the difference between the two is in `translate_item_no_enqueue` or `translate_item`, or, more precisely `translate_item_maybe_enqueue`. How can the enqueue one calls the opposite?

Nadrieril: Hm ok maybe that's hard. Let's do your thing with a common function that takes a boolean then!

ssyram: OK.

ssyram: Done.

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -393,7 +393,11 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
                                     unreachable!()
                                 };
                                 Ok(ctx
-                                    .translate_vtable_struct_ref(span, &trait_predicate.trait_ref)?
+                                    .translate_vtable_struct_ref_maybe_enqueue(
+                                        false,
+                                        span,
+                                        &trait_predicate.trait_ref,
+                                    )?
```
Nadrieril: You missed this one

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -488,7 +525,7 @@ impl ItemTransCtx<'_, '_> {
             _ => unreachable!(),
         };
         let vtable_struct_ref = self
-            .translate_vtable_struct_ref(span, implemented_trait)?
+            .translate_vtable_struct_ref_no_enqueue(span, implemented_trait)?
```
Nadrieril: This one should enqueue too I think

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -618,7 +655,7 @@ impl ItemTransCtx<'_, '_> {
 
             let vtable_def_ref = self
                 .translate_region_binder(span, &impl_expr.r#trait, |ctx, tref| {
-                    ctx.translate_vtable_struct_ref(span, tref)
+                    ctx.translate_vtable_struct_ref_no_enqueue(span, tref)
```
Nadrieril: This one should enqueue too

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: This one should enqueue too, no?

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: Done.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
ssyram: OK.

charon/src/bin/charon-driver/translate/translate_types.rs
ssyram: OK, thanks for pointing out.

PR #848, opened 2025-10-14
Add a fake `Drop::drop_in_place(*mut self)` method

### LINKED ISSUE #819: Bug: Charon confuses drop glue and `Drop` implementations

Nadrieril:
Fixes #819 by distinguishing `fn drop(&mut self)` and `fn drop_in_place(*mut self)`.

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -756,16 +756,40 @@ impl ItemTransCtx<'_, '_> {
         };
 
         // Build a reference to `std::ptr::drop_in_place<T>`.
-        let drop_in_place: hax::ItemRef = {
-            let s = self.hax_state_with_id();
-            let drop_in_place = self.tcx.lang_items().drop_in_place_fn().unwrap();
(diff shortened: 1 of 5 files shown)


ssyram:
So ultimately, what are the differences?

Issue #824, opened 2025-09-30
Add a pass that monomorphizes types instantiated with types that contain mutable borrows


Nadrieril:
That's for aeneas, which currently doesn't have a way to handle instanciating a generic type with a type that contains mutable borrows.

PR #852, opened 2025-10-16
Partially implement `&mut` partial monomorphization feature

Nadrieril:
Implements the bulk of #824. Missing features:
- const generics; today `Type<N, &mut bool>` causes errors;
- method generics; today `self.method::<&mut bool>()` causes erreurs;
- fn pointers `fn(&mut bool) -> ...`;
- testing both name matchers with the feature on.

### DIFF
--- a/charon/src/transform/normalize/partial_monomorphization.rs
+++ b/charon/src/transform/normalize/partial_monomorphization.rs
@@ -339,7 +339,14 @@ impl<'a> PartialMonomorphizer<'a> {
                     tref.generics.types.iter().any(|ty| self.is_infected(ty))
                 }
             },
-            TyKind::FnDef(_region_binder) => false, // TODO
+            TyKind::FnDef(..) | TyKind::FnPtr(..) => {
+                register_error!(
+                    self.ctx,
+                    self.span,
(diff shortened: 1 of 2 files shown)


charon/tests/ui/monomorphize-mut.out
```diff
@@ -0,0 +1,464 @@
...
+}
+
+#[lang_item("Option")]
+pub enum core::option::Option::<&_ mut (&_ mut (_))><'_0, '_1, T0>
+where
+    [@TraitClause0]: Sized<&'_0 mut (&'_1 mut (T0))>,
+{
+  None,
+  Some(&'_0 mut (&'_1 mut (T0))),
+}
```
sonmarcho, line 255: This is duplicated?

charon/tests/ui/monomorphize-mut.out
```diff
@@ -0,0 +1,464 @@
...
```
sonmarcho, line 214: This is duplicated?

Nadrieril: Well spotted, I'll look into it

PR #872, opened 2025-10-27
Supporting vtables for closures

ssyram:
> This PR is a re-submission of #871 , as @Sam-Ni does not have the permission of running the CI, I resubmit for him to view the CI results.

This PR adds support for vtable instantiation for closures.
The implementation involves two main steps:

1. Matching a closure implementation

2. Instantiating the corresponding vtable

Step 1 is implemented in `translate_body.rs`, while Step 2 is implemented in `translate_trait_objects.rs`.
The modifications in each file are described below.

1. translate_body.rs

- `translate_rvalue`
Supports registering closure trait implementations. For example, in the test file `dyn-fn.rs`, the closure `|counter| { ... }` is matched in the `hax::ImplExprAtom::Builtin` case and registered as a `VTableInstance`.

2. translate_trait_objects.rs

- `add_method_to_vtable_def`
Adds additional checks for `FnOnce` when constructing the method field in the `FnOnce` vtable, due to the incompatibility of `FnOnce::call_once` with dyn-compatibility.

- `get_vtable_instance_info`
Supports retrieving the vtable struct reference and the corresponding trait declaration reference for closures, based on their kind (Fn, FnMut, or FnOnce).

- `add_supertraits_to_vtable_value`
Supports filling vtable fields with built-in supertraits. For closures supertraits, they are translated as VTableInstance with `TraitImplSource::Closure(kind)`, generating a vtable according to the closure `kind`. For other built-in supertraits, they are translated as VTableInstance with `TraitImplSource::Normal`, whose trait reference itself is used as the impl reference (e.g. `@2 := &core::marker::MetaSized::{vtable}`).

- `gen_vtable_instance_init_body`
Supports generating the vtable body for closures. Closure implementations are treated as trait implementations with no associated items and a single specialized closure method (`call` or `call_mut`). `call_once` is ignored due to the incompatibility mentioned above. Based on this design, the function retrieves `(trait_pred, items)` for closures and uses `mk_field` to add the specialized closure method.

- `generate_closure_method_shim_ref`
Implements the generation of the closure method shim reference used in the `mk_field` in `gen_vtable_instance_init_body`.


R1kM:
Done

charon/tests/ui/vtables.rs
```diff
@@ -28,6 +28,21 @@ fn to_dyn_obj<T: NoParam>(arg: &T) -> &dyn NoParam {
...
+// }
+
+// fn ok_clone<T: Clone>(arg: &T) -> T {
+//     arg.clone()
+// }
+// fn use_ok_clone(arg: &String) -> String {
+//     ok_clone(arg);
+//     let s = MyStruct { a: 10, b: String::from("Hello") };
+//     ok_clone(&s).b
+// }
```
Nadrieril: Why is this commented out? This does not feel useful for vtables, please remove

charon/tests/ui/simple/dyn-fn.out
```diff
@@ -1,254 +1,11 @@
...
+thread 'rustc' panicked at src/bin/charon-driver/translate/translate_trait_objects.rs:941:17:
+internal error: entered unreachable code
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+error: Thread panicked when extracting item `test_crate::gives_fn::{closure#0}`.
+ --> tests/ui/simple/dyn-fn.rs:8:15
+  |
+8 |     takes_fn(&|counter| {
+  |               ^^^^^^^^^
 
+ERROR Charon failed to translate this code (1 errors)
```
Nadrieril: This is not acceptable in a PR. If a PR regresses a test like this you must indicate it in the PR description and explain why you think it is justified to accept this regression. In this case it's a panic, which at the very least should be changed to emit e rpoper error message (but most likely should be fixed instead).

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -523,22 +537,49 @@ impl ItemTransCtx<'_, '_> {
...
+                    }
+                    TraitImplSource::Closure(ClosureKind::Fn) => {
+                        fn_impl.as_ref().expect("Fn impl should exist")
+                    }
+                    _ => unreachable!("Expected closure trait impl source"),
+                };
+                let target_kind = match impl_kind {
+                    TraitImplSource::Closure(kind) => *kind,
+                    _ => unreachable!(),
+                };
```
Nadrieril: nit: do the `let target_kind` first and match on that instead of matching twice on `impl_kind`

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -679,16 +741,126 @@ impl ItemTransCtx<'_, '_> {
...
+                        // The challenge is that we need an impl_ref, but builtin traits don't have concrete impls
+                        // Let's try using the trait reference itself as the impl reference
+                        // A simple case would be that they are all marker traits like `core::marker::MetaSized`
+                        let trait_ref = impl_expr.r#trait.hax_skip_binder_ref();
+                        let mut vtable_instance_ref: GlobalDeclRef = self
+                            .translate_item_no_enqueue(
+                                span,
+                                trait_ref,
+                                TransItemSourceKind::VTableInstance(
+                                    TraitImplSource::Normal, // Builtin traits are normal impls
```
Nadrieril: Built-in traits are not normal impls. In fact you're translating a trait reference here, which won't work in poly mode because in poly mode we'd only try to generate that impl once for the fully generic trait (e.g. a single `impl<T> Sized for T` or something, which is not what we want). The way to generate built-in impls is like we did for closures: the base item must be the ADT for which we generate a built-in impl. That's another project, for now please remove this branch.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -679,16 +741,126 @@ impl ItemTransCtx<'_, '_> {
...
+                            .generics
+                            .types
+                            .remove_and_shift_ids(TypeVarId::ZERO);
+                        let global = Box::new(ConstantExpr {
+                            kind: ConstantExprKind::Global(vtable_instance_ref),
+                            ty: fn_ptr_ty,
+                        });
+                        ConstantExprKind::Ref(global)
+                    } else {
+                        // For non-dyn-compatible builtin traits, we don't need vtable instances
```
Nadrieril: Please emit an error here, this shuold not be happening.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -797,12 +1011,23 @@ impl ItemTransCtx<'_, '_> {
         };
         mk_field(ConstantExprKind::FnPtr(fn_ptr));
 
+        // Add usual method
```
Nadrieril:
```suggestion
        // Add the normal methods
```

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -699,13 +871,55 @@ impl ItemTransCtx<'_, '_> {
...
+                    };
+                    // Get the appropriate trait implementation for this closure kind
+                    let trait_impl = match closure_kind {
+                        ClosureKind::FnOnce => fn_once_impl,
+                        ClosureKind::FnMut => {
+                            fn_mut_impl.as_ref().expect("FnMut impl should exist")
+                        }
+                        ClosureKind::Fn => fn_impl.as_ref().expect("Fn impl should exist"),
+                    };
+                    (&trait_impl.trait_pred, &vec![])
```
Nadrieril: instead of returning an empty set of items which is confusing, I'd prefer if you added the items directly in the other branch

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -823,6 +1048,45 @@ impl ItemTransCtx<'_, '_> {
         Ok(Body::Unstructured(builder.build()))
     }
 
+    fn generate_closure_method_shim_ref(
+        &mut self,
+        span: Span,
+        impl_def: &hax::FullDef,
+        closure_kind: ClosureKind,
+    ) -> Result<ConstantExprKind, Error> {
```
Nadrieril: `translate_fn_ptr` doesn't work here?

Nadrieril: I'd prefer this method to be in translate_closures also

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -823,6 +1048,45 @@ impl ItemTransCtx<'_, '_> {
...
+        // The signature can only contain region binders
+        let sig = &args.fn_sig;
+        for _ in &sig.bound_vars {
+            // The late-bound region binders are at last, for the whole closure, as per `translate_closures`, use push
+            generics.regions.push(Region::Erased);
+        }
+        // Finally, one more region for the receiver, this is at first, as it is the region of the function itself, use insert at head
+        generics
+            .regions
+            .insert_and_shift_ids(RegionId::ZERO, Region::Erased);
```
Nadrieril: I don't understand this, could you add an example to show which regions come from where?

charon/tests/ui/vtables.rs
Sam-Ni: removed

charon/tests/ui/simple/dyn-fn.out
Sam-Ni: Sorry for that. This panic occurs because Fn* Traits are builtin traits, which are marked opaque and not translated by default. I fixed this and now charon can translate closures.

Sam-Ni: However, a new panic appears when translating dyn-fn.rs. The error message is as follows:

```
error: Trying to generate a vtable shim for a non-vtable-safe method
  --> tests/ui/simple/dyn-fn.rs:8:15
   |
 8 |       takes_fn(&|counter| {
   |  _______________^
 9 | |         *counter += 1;
10 | |         true
11 | |     })
   | |_____^
````

This panic happened in function `translate_vtable_shim`, which assumes that the input `impl_func_def` is `FullDefKind::AssocFun`. However, when translating a closure, `impl_func_def` as a `FullDefKind::Closure` is passed. So the new problem is that `FullDefKind::Closure` does not have fields `vtable_sig` and `sig`, which are offered by `FullDefKind::AssocFun`. Should we fcollect these information in charon, or in hax?

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 3 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: fixed

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 4 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: fixed

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 5 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: fixed

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 6 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: fixed

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 7 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: fixed

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 8 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: > `translate_fn_ptr` doesn't work here?

We tried `translate_fn_ptr` and pass the impl_def of closure. However, error occurs when executing ``translate_fn_ptr` -> `translate_fun_item` -> `translate_generic_args`, reporting “Unexpected error: could not find" type variables `closurekind` and `upvars`.

Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: > I'd prefer this method to be in translate_closures also

moved into translate_closures

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs (hunk 9 above)
Comment (Sam-Ni) [reply charon/src/bin/charon-driver/translate/translate_trait_objects.rs]: An example `dyn-closure-region` is added

File: charon/tests/ui/simple/dyn-fn.out (hunk 2 above)
Comment (Nadrieril) [reply charon/tests/ui/simple/dyn-fn.out]: I see, this should be added in hax

File: charon/src/bin/charon-driver/translate/translate_trait_objects.rs [hunk 10]
```diff
@@ -679,16 +744,98 @@ impl ItemTransCtx<'_, '_> {
...
+                                closure_kind,
+                            )),
+                        )?;
+                        let global = Box::new(ConstantExpr {
+                            kind: ConstantExprKind::Global(vtable_instance_ref),
+                            ty: fn_ptr_ty,
+                        });
+                        ConstantExprKind::Ref(global)
+                    } else if self
+                        .translate_vtable_struct_ref(span, impl_expr.r#trait.hax_skip_binder_ref())?
```
Nadrieril, line 808: You can't skip binders like that, this will result in errors if the binder is not trivial.. Please use `translate_region_binder` then erase the resulting `charon::RegionBinder`.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Nadrieril: I looked into it, that's an inconsistency in hax actually. I fixed it in latest hax,  `translate_fn_ptr` looks like it works now (the generics still need to be fixed up though of course).

charon/src/bin/charon-driver/translate/translate_closures.rs
```diff
@@ -745,4 +745,43 @@ impl ItemTransCtx<'_, '_> {
...
+        closure_kind: ClosureKind,
+    ) -> Result<ConstantExprKind, Error> {
+        // Register the closure method shim
+        let shim_id: FunDeclId = self.register_item(
+            span,
+            impl_def.this(),
+            TransItemSourceKind::VTableMethod(TraitImplSource::Closure(closure_kind)),
+        );
+
+        let mut generics = Box::new(self.the_only_binder().params.identity_args());
```
Nadrieril, line 762: Please use the appropriate `translate_closure_bound_ref_etc` method here instead of most of this function

Nadrieril:
That's fixed now, right?

PR #895, opened 2025-11-17
Add an optional pass to desugar Drop to Call in ULLBC

Hua-hana:
This PR adds an optional transform pass which transforms `TerminatorKind::Drop(tref, place)` to `TerminatorKind::Call(Method(tref, "drop_in_place"), &raw mut place)` to make Drop an actual function call. It would be activated when we add `--desugar-drops` option. This pass is useful for further code generation, e.g., for the C code generation in Eurydice.

### DIFF
--- a/charon/src/transform/normalize/desugar_drops.rs
+++ b/charon/src/transform/normalize/desugar_drops.rs
@@ -1,7 +1,5 @@
 use super::super::ctx::UllbcPass;
 use crate::{
-    errors::Error,
-    raise_error, register_error,
     transform::{
         TransformCtx,
         ctx::{BodyTransformCtx, UllbcStatementTransformCtx},
@@ -21,7 +19,8 @@ fn is_noop_destruct(tref: &TraitRef) -> bool {
 
 impl<'a> UllbcStatementTransformCtx<'a> {
     /// Transform a Drop to a Call that calls the drop_in_place method.
-    fn transform_drop_to_call(&mut self, term: &mut Terminator) -> Result<(), Error> {
+    /// If we cannot desugar this drop, we just leave it unchanged.
+    fn transform_drop_to_call(&mut self, term: &mut Terminator) {
         if let TerminatorKind::Drop {
             place,
             tref,
@@ -34,7 +33,7 @@ impl<'a> UllbcStatementTransformCtx<'a> {
                 term.kind = TerminatorKind::Goto {
                     target: target.clone(),
                 };
-                return Ok(());
+                return;
             }
 
             let ref_drop_arg = TyKind::RawPtr(place.ty().clone(), RefKind::Mut).into_ty();
@@ -53,17 +52,12 @@ impl<'a> UllbcStatementTransformCtx<'a> {
             // Get the declaration id of drop_in_place from tref
             let trait_id = tref.trait_decl_ref.skip_binder.id;
             let Some(tdecl) = self.ctx.translated.trait_decls.get(trait_id) else {
-                return Ok(());
+                return;
             };
             let method_name = TraitItemName("drop_in_place".into());
             let Some(bound_method) = tdecl.methods.iter().find(|m| m.name() == method_name) else {
-                raise_error!(
-                    self.ctx,
-                    self.span,
-                    "Could not find a method with name \
-                    `{method_name}` in trait `{:?}`",
-                    trait_id,
-                )
+                // skip this drop if we cannot find its method id
+                return;
             };
             let method_decl_id = bound_method.skip_binder.item.id;
 
@@ -82,7 +76,6 @@ impl<'a> UllbcStatementTransformCtx<'a> {
                 on_unwind: on_unwind.clone(),
             };
         }
-        Ok(())
     }
 }
 
@@ -94,7 +87,7 @@ impl UllbcPass for Transform {
             return;
         }
         decl.transform_ullbc_terminators(ctx, |ctx, term| {
-            let _ = ctx.transform_drop_to_call(term);
+            ctx.transform_drop_to_call(term);
         });
     }
 }


charon/src/options.rs
```diff
@@ -266,6 +266,13 @@ pub struct CliOpts {
     #[clap(long = "preset")]
     #[arg(value_enum)]
     pub preset: Option<Preset>,
+
+    #[clap(
+        long = "desugar-drops",
+        help = "If activated, transform `Drop(p)` to `Call drop_in_place(&mut p)` for ULLBC"
```
Nadrieril:
```suggestion
        help = "If activated, transform `Drop(p)` to `Call drop_in_place(&raw mut p)`"
```

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+            let drop_arg = self.fresh_var(Some("drop_arg".into()), ref_drop_arg);
+            let drop_ret = self.fresh_var(Some("drop_ret".into()), Ty::mk_unit());
+
+            let unit_metadata = self.fresh_var(None, Ty::mk_unit());
+            let rval = Rvalue::RawPtr {
+                place: place.clone(),
+                kind: RefKind::Mut,
+                ptr_metadata: Operand::Move(unit_metadata),
+            };
+            // assign &mut place to drop_arg
```
Nadrieril:
```suggestion
            // assign &raw mut place to drop_arg
```

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+            tref,
+            target,
+            on_unwind,
+        } = &mut term.kind
+        {
+            let ref_drop_arg = TyKind::RawPtr(place.ty().clone(), RefKind::Mut).into_ty();
+            let drop_arg = self.fresh_var(Some("drop_arg".into()), ref_drop_arg);
+            let drop_ret = self.fresh_var(Some("drop_ret".into()), Ty::mk_unit());
+
+            let unit_metadata = self.fresh_var(None, Ty::mk_unit());
```
Nadrieril: Prefer `self.compute_place_metadata` as this can avoid creating a bunch of `()` locals everywhere.
```suggestion
            let ptr_metadata = self.compute_place_metadata(&place);
```

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+                place: place.clone(),
+                kind: RefKind::Mut,
+                ptr_metadata: Operand::Move(unit_metadata),
+            };
+            // assign &mut place to drop_arg
+            self.statements.push(Statement {
+                span: self.span,
+                kind: StatementKind::Assign(drop_arg.clone(), rval),
+                comments_before: Vec::new(),
+            });
```
Nadrieril: `BodyTransformCtx` has a method for that:
```suggestion
            self.insert_assn_stmt(drop_arg.clone(), rval);
```

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+            let item_name = TraitItemName("drop_in_place".into());
+
+            match &tref.kind {
+                TraitRefKind::TraitImpl(impl_ref) => {
+                    let Some(item) = self.ctx.translated.get_item(impl_ref.id) else {
+                        raise_error!(
+                            self.ctx,
+                            self.span,
+                            "Could not find TraitImpl item for Destruct trait."
+                        );
```
Nadrieril: I propose that this pass should never emit an error and instead should just leave the `Drop` terminator as-is when we can't desugar it.

charon/src/transform/mod.rs
```diff
@@ -110,6 +111,8 @@ pub static INITIAL_CLEANUP_PASSES: &[Pass] = &[
     NonBody(&normalize::expand_associated_types::Transform),
     // `--remove-adt-clauses`: Remove all trait clauses from type declarations.
     NonBody(&simplify_output::remove_adt_clauses::Transform),
+    // Transform Drops into Calls to drop_in_place.
+    UnstructuredBody(&normalize::desugar_drops::Transform),
```
Nadrieril: Please move this pass before `skip_trait_refs_when_known` so that we can benefit from its normalization of trait method calls.

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+                            self.span,
+                            "Could not find drop_in_place method in Destruct trait impl."
+                        );
+                    };
+                    let method_id = item_binder.1.skip_binder.id;
+
+                    let fn_ptr = FnPtr::new(
+                        FnPtrKind::Trait(tref.clone(), item_name, method_id),
+                        GenericArgs::empty(),
+                    );
```
Nadrieril: This is tricky but the `method_id` here must be the id of the method _declaration_, not implementation.

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+                                self.span,
+                                "desugar_drops: Expected NoopDestruct or UntrackedDestruct kind in BuiltinOrAuto."
+                            );
+                        }
+                    }
+                }
+                TraitRefKind::Clause(_) => {
+                    // Call the trait method
+                    let fn_ptr = FnPtr::new(
+                        FnPtrKind::Trait(tref.clone(), item_name, FunDeclId::new(0)),
```
Nadrieril: It is never ok to create a fake `FunDeclId`. This is an important mistake, we must not generate known-incorrect ASTs in Charon. If you did this yourself you should have instead asked me for help. If this is LLM-generated code you should review its code better. I dislike the feeling that I can't trust the author of a PR to have thought about such invariants.

As mentioned above, the `FunDeclId` required here is the one of the method declaration, which you can find by looking up the trait.

charon/src/transform/normalize/desugar_drops.rs
```diff
@@ -0,0 +1,147 @@
...
+            self.statements.push(Statement {
+                span: self.span,
+                kind: StatementKind::Assign(drop_arg.clone(), rval),
+                comments_before: Vec::new(),
+            });
+
+            let item_name = TraitItemName("drop_in_place".into());
+
+            match &tref.kind {
+                TraitRefKind::TraitImpl(impl_ref) => {
```
Nadrieril: You don't actually need to inspect the trait ref here beyond checking if it's a noop. You can build the `FnPtr` unconditionally otherwise.

charon/src/transform/normalize/desugar_drops.rs
Hua-hana: Sorry about that. It is my mistake. I thought fn_ptr should receive an implementation id. I cannot find a way to generate the id for the trait clause so I did it mistakenly. It is not LLM-generated code. I would ask you next time if I find some problem I cannot solve.

Nadrieril: Thank you, this is just to avoid very-hard-to-debug errors in the future

Nadrieril:
Looks great! Thanks for this :)

PR #900, opened 2025-11-19
Generate drop shim function for vtable instance

Hua-hana:
This PR adds code to generate drop shim for the vtable instance generated for some `impl Trait for T`. For now it uses Hax to get the implemetation of `T: Destruct` in the charon side. Maybe in the future we should record this information directly in the hax::FullDefKind::TraitImpl.

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -915,8 +915,7 @@ impl ItemTransCtx<'_, '_> {
     /// local target_self@2 : TargetReceiverTy;
     /// // perform some conversion to cast / re-box the drop_shim receiver to the target receiver
     /// target_self@2 := concretize_cast<ShimReceiverTy, TargetReceiverTy>(shim_self@1);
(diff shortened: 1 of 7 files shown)


charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -937,6 +906,132 @@ impl ItemTransCtx<'_, '_> {
...
+                GenericArgs::empty(),
+            )
+        };
+
+        // call drop_in_place
+        builder.call(Call {
+            func: FnOperand::Regular(fn_ptr),
+            args: Vec::from([Operand::Move(target_self.clone())]),
+            dest: drop_ret_place,
+        });
```
Nadrieril: Instead of generating a `Call`, would you generate a `TerminatorKind::Drop(*target_self)`? This is for users that don't use `--desugar-drops` and want to identify where drops happen. I recommend adding a `drop` method to `BodyBuilder` that looks like its `call` method.

Hua-hana: It is done

Issue #917, opened 2025-11-27
Bug: const clauses + const lose generics


N1ark:
I believe there is some error when translating consts, that means generics get lost.

In particular, take this code:

```rust
//@ charon-args=--monomorphize

#![feature(const_trait_impl)]

pub const trait HasFn<O> {
    fn get_fn() -> fn() -> O;
}

struct Bar;

fn bar() {}

impl const HasFn<()> for Bar {
    fn get_fn() -> fn() {
        bar
    }
}

fn foo<T, O>() -> O
where
    T: const HasFn<O>,
{
    let f: fn() -> O = const { T::get_fn() };
    f()
}

pub fn main() {
    foo::<Bar, ()>();
}
```

First, Charon gets an error because of an unhandled `ClauseKind::HostEffect` (due to the `const HasFn<O>` in `foo`). If we chose to just ignore these clauses (which should be ok),  we then get the following error:

```
error: Unexpected error: could not find the type variable O
  --> tests/ui/const-clause.rs:23:30
   |
23 |     let f: fn() -> O = const { T::get_fn() };
   |                              ^^^^^^^^^^^^^^^

error: Unexpected error: could not find the type variable O
  --> tests/ui/const-clause.rs:23:30
   |
23 |     let f: fn() -> O = const { T::get_fn() };
   |                              ^^^^^^^^^^^^^^^

ERROR Charon failed to translate this code (2 errors)
```

I am not sure what happens here; in particular, having `let f = T::get_fn()` (without the `const`) doesn't raise this error. I believe somewhere in the global translation we lose generics of the outer body?

Not sure because I don't fully grasp generics, just wanted to report this. I ran into this when wanting to tweak hax/Charon to translate the following piece of code, which fails for similar reasons [here](https://doc.rust-lang.org/src/alloc/vec/in_place_collect.rs.html#227-248).
```rust
let _: Vec<u8> = vec![0u8].into_iter().filter(|_| true).collect();
```

Nadrieril:
The weird thing is that this shows up with `--monomorphize`. We shouldn't be translating generics when this flag is passed. In fact your example works in poly mode.

PR #920, opened 2025-11-27
Add `ConstExprKind::FnPtr`, rename `FnPtr->FnDef`

N1ark:
That naming was horrendous 🫠 Now a `FnDef` is a proper `FnDef`, and `FnPtr` is a proper `FnPtr`.

I unfortunately couldn't get a simple example where a hax `FnPtr` gets generated. I can only say confidently that it occurs in this code:
```rust
let _: Vec<u8> = vec![0u8].into_iter().filter(|_| true).collect();
```
However this code still doesn't compile because somewhere inside it does a function call on a `const` operand, which hax crashes on. Ideally we should probably change `FnOperand` to just be `Static(FunDeclRef) | Dynamic(Operand)`, but trying to do that change also brought errors because of #917 (now fixed). I'll give it another try later :p

(the eurydice PR is fine i just closed one by accident and had to change it so CI ran w the wrong one)

ci: use https://github.com/AeneasVerif/aeneas/pull/660
ci: use https://github.com/AeneasVerif/eurydice/pull/361

### DIFF
--- a/charon/src/ast/expressions.rs
+++ b/charon/src/ast/expressions.rs
@@ -562,7 +562,12 @@ pub enum ConstantExprKind {
     Ptr(RefKind, Box<ConstantExpr>),
     /// A const generic var
     Var(ConstGenericDbVar),
-    /// Function pointer
+    /// Function definition -- this is a ZST constant
+    FnDef(FnPtr),
+    /// A function pointer to a function item; this is an actual pointer to that function item.
+    ///
+    /// We eliminate this case in a micro-pass.
+    #[charon::opaque]
     FnPtr(FnPtr),
     /// A pointer with no provenance (e.g. 0 for the null pointer)
     ///
--- a/charon/src/transform/simplify_output/simplify_constants.rs
+++ b/charon/src/transform/simplify_output/simplify_constants.rs
@@ -32,7 +32,7 @@ fn transform_constant_expr(
         | ConstantExprKind::Var(_)
         | ConstantExprKind::RawMemory(..)
         | ConstantExprKind::TraitConst(..)
-        | ConstantExprKind::FnPtr(..)
+        | ConstantExprKind::FnDef(..)
         | ConstantExprKind::Opaque(_) => {
             // Nothing to do
             // TODO: for trait const: might come from a top-level impl, so we might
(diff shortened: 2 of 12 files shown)


charon/src/ast/expressions.rs
```diff
@@ -562,7 +562,12 @@ pub enum ConstantExprKind {
     Ptr(RefKind, Box<ConstantExpr>),
     /// A const generic var
     Var(ConstGenericDbVar),
-    /// Function pointer
+    /// Function definition -- this is a ZST constant
+    FnDef(FnPtr),
+    /// A function pointer; this is an actual pointer to that function
+    ///
+    /// We eliminate this case in a micro-pass.
+    #[charon::opaque]
```
Nadrieril:
```suggestion
    /// A function pointer to a function item; this is an actual pointer to that function item.
    ///
    /// We eliminate this case in a micro-pass.
    #[charon::opaque]
```

charon/src/transform/simplify_output/simplify_constants.rs
```diff
@@ -123,6 +123,21 @@ fn transform_constant_expr(
             let ty = tref.generics.types[0].clone();
             Rvalue::Aggregate(AggregateKind::Array(ty, len), fields)
         }
+        ConstantExprKind::FnPtr(fptr) => {
+            let TyKind::FnPtr(sig) = val.ty.kind() else {
+                unreachable!("FnPtr constant must have FnPtr type");
+            };
+            let from_ty = TyKind::FnDef(sig.clone().map(|_| fptr.clone())).into_ty();
```
Nadrieril: Something feels wrong with these binders. The `fptr: FnPtr` may refer to lifetimes in the current context. I think the `TyKind::FnDef` corresponding to a given function item should be unique, so we should quantify its lifetimes always the same way. I don't know how to retrieve that though. Maybe nevermind for now.

Either way, there's a binder level mismatch:
```suggestion
            let from_ty = TyKind::FnDef(sig.clone().map(|_| fptr.clone().move_under_binder())).into_ty();
```

N1ark: hmmm i see -- im happy to leave it as is for now and maybe if we run into incorrect generics later we change it. im unsure in general of how frequently a fn pointer is encountered in a const context

PR #921, opened 2025-11-27
Add `ConstantExprKind::Slice`

N1ark:
Properly distinguish array and slice constants, and then properly desugar slice constants into an unsize! Previously, we used to store an array constant expr into a `[T]` typed location, which isn't possible since that is unsized.

This is a similar weird thing as with string constants where the constant actually hides the presence of a reference. This really makes me think we're doing something wrong around these :s I need to bump #785 too this reminds me

### DIFF
--- a/charon-ml/src/PrintTypes.ml
+++ b/charon-ml/src/PrintTypes.ml
@@ -196,6 +196,11 @@ and fun_decl_id_to_string (env : 'a fmt_env) (id : FunDeclId.id) : string =
   | None -> fun_decl_id_to_pretty_string id
   | Some def -> name_to_string env def.item_meta.name
 
+and fun_decl_ref_to_string (env : 'a fmt_env) (fn : fun_decl_ref) : string =
+  let fun_id = fun_decl_id_to_string env fn.id in
+  let generics = generic_args_to_string env fn.generics in
+  fun_id ^ generics
+
 and global_decl_id_to_string env def_id =
   match GlobalDeclId.Map.find_opt def_id env.crate.global_decls with
   | None -> global_decl_id_to_pretty_string def_id


charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -1152,6 +1152,10 @@ impl<C: AstFormatter> FmtWithCtx<C> for ConstantExprKind {
                 let values = values.iter().map(|v| v.with_ctx(ctx)).format(", ");
                 write!(f, "[{}]", values)
             }
+            ConstantExprKind::Slice(values) => {
+                let values = values.iter().map(|v| v.with_ctx(ctx)).format(", ");
+                write!(f, "&[{}]", values)
```
Nadrieril: That doesn't seem ok, the type should be `[T]`, not `&[T]` right?

charon/src/transform/simplify_output/simplify_constants.rs
```diff
@@ -63,9 +62,24 @@ fn transform_constant_expr(
...
+                    let bval_ty = bval.ty().clone();
+                    let adt = bval_ty.as_adt().expect("Non-adt slice sub-constant");
+                    assert_eq!(adt.id, TypeId::Builtin(BuiltinTy::Array));
+
+                    let len = adt.generics.const_generics[0].clone();
+                    let metadata = UnsizingMetadata::Length(len);
+                    break 'ref_const Rvalue::UnaryOp(
+                        UnOp::Cast(CastKind::Unsize(bval_ty, val.ty.clone(), metadata)),
+                        bval,
+                    );
```
Nadrieril: Unsizing casts today only apply to pointers; making it by-value would mean we could have unsized values, but then where do we store the metadata? I do think we should have the option of unsized constants/statics (by adding a metadata tield to GlobalDecl`), but I don't know how we'd produce such a value.

charon/src/transform/simplify_output/simplify_constants.rs
```diff
@@ -115,12 +129,15 @@ fn transform_constant_expr(
...
-            let ty = tref.generics.types[0].clone();
+            let mut tref = val.ty.kind().as_adt().unwrap().clone();
+            let ty = tref.generics.types.pop().unwrap();
+            match tref.id.as_builtin().unwrap() {
+                BuiltinTy::Array => {}
+                BuiltinTy::Slice => val.ty = Ty::mk_array(ty.clone(), len.clone()),
+                _ => {
+                    unreachable!("Unpexected builtin type in array/slice constant")
+                }
+            }
```
Nadrieril, line 152: So you're basically transforming the `ConstantExpr::Slice` to `ConstantExpr::Array` first, then desugaring the array? I'd prefer if you did that explicitly at the start of the function, but also doesn't that create a type error?

charon/tests/ui/simple/const-subslice.out
```diff
@@ -19,15 +19,15 @@ fn main()
...
+    let @6: Array<u8, 3 : usize>; // anonymous local
     let @7: &'_ (Slice<u8>); // anonymous local
 
     @0 := ()
     storage_live(y@1)
     storage_live(@6)
     @6 := [const (0 : u8), const (1 : u8), const (2 : u8)]
     storage_live(@7)
-    @7 := &@6
+    @7 := unsize_cast<Array<u8, 3 : usize>, &'_ (Slice<u8>), 3 : usize>(move (@6))
```
Nadrieril: This is casting an array to a reference, something has gone wrong

charon/src/pretty/fmt_with_ctx.rs
N1ark: hmmm true true -- wanted to differentiate it from an array constant but maybe its fine as is

charon/tests/ui/simple/const-subslice.out
N1ark: ah good catch !! woooops

charon/src/transform/simplify_output/simplify_constants.rs
N1ark: hmmm why would it be a type error?

charon/src/transform/simplify_output/simplify_constants.rs
N1ark: that was a mistake indeed the unsize should be on a reference here.
i agree this is something we want (for slices and strings, mainly), and i have no idea how we would do it; rather I think i'd want all globals to be sized, but that when we desugar the `Global` places, we can have a metadata that doesn't match the global's type. So you could have a global of type `[u8; 10]`, but when we desugar the constant, it would be `&GLOBAL with_metadata const 10`. not sure how that fits in with the rest though

Nadrieril: that part fits well: you're basically doing the unsizing cast whenever you reference the constant. the issue is str then, since it doesn't have a sized equivalent.

charon/src/transform/simplify_output/simplify_constants.rs
Nadrieril: Because if e.g. we're in a `Assign(place, Operand::Const(c))` then `place.ty() == c.ty`; you can't just change c.ty.

charon/src/ast/expressions.rs
```diff
@@ -517,6 +517,8 @@ pub enum ConstantExprKind {
     Adt(Option<VariantId>, Vec<ConstantExpr>),
     #[charon::opaque]
     Array(Vec<ConstantExpr>),
+    #[charon::opaque]
+    Slice(Vec<ConstantExpr>),
```
Nadrieril, line 521: I see `ConstantExpr`s as more like "constructors" than anything, so I don't mind using `Array` for slices too.

charon/src/transform/simplify_output/simplify_constants.rs
Nadrieril: Oh wait, the only side-effect of this assignment is that `val.ty` is read at the end of the function? That's ugly come on x) Plz `let mut rval_ty = val.ty` at the start of the function so we can follow what's up.

charon/src/transform/simplify_output/simplify_constants.rs
```diff
@@ -63,6 +62,32 @@ fn transform_constant_expr(
                 })),
             )
         }
+        ConstantExprKind::Ref(bval) if bval.ty.is_slice() => {
+            // We need to special-case slices; we don't take a reference to the slice,
+            // but an unsizing cast
+            // We generate the following code:
+            // let array: [T; N] = [...];
+            // let array_ref: &[T; N] = &array;
+            // return unsize::<&[T; N], &[T]>(array_ref);
```
Nadrieril, line 72: Yeah I like that :+1:

charon/src/transform/simplify_output/simplify_constants.rs
N1ark: i mean the sized equivalent is `[u8; N]` with an additional well-formed invariant; but I think if we are the ones generating the string globals we can ensure the strings are well-formed, so implicitly transmuting (via the global's type) a `&[u8]` into a `&str` is fine

PR #923, opened 2025-11-28
Make const-eval bytes more semantically precise

N1ark:
The last of my `ConstantExprKind` PRs :) 
I tried hard to find an example where this occurs and crashes hax, but I couldn't, again I wonder if it's a matter of MIR settings? I know that in my Charon fork I run into union constants (notably [here](https://doc.rust-lang.org/src/core/array/mod.rs.html#151)), but somehow hax/Charon compiles this into an aggregate assignment.

I still think it's a nice thing to have for clients (and I would need it..), but if you'd rather not add this I understand.

### DIFF
--- a/charon/src/ast/expressions.rs
+++ b/charon/src/ast/expressions.rs
@@ -523,6 +523,27 @@ impl From<FunDeclRef> for FnPtr {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, SerializeState, DeserializeState, Drive, DriveMut, Hash)]
+pub enum Provenance {
+    Global(GlobalDeclRef),
+    Function(FunDeclRef),
+    Unknown,
+}
+
+/// A byte, in the MiniRust sense: it can either be uninitialized, a concrete u8 value,
+/// or part of a pointer with provenance (e.g. to a global or a function)
+#[derive(Debug, PartialEq, Eq, Clone, SerializeState, DeserializeState, Drive, DriveMut, Hash)]
+pub enum Byte {
+    /// An uninitialized byte
+    Uninit,
+    /// A concrete byte value
+    Value(u8),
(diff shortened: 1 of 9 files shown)


charon/src/ast/expressions.rs
```diff
@@ -515,6 +515,10 @@ pub enum ConstantExprKind {
     /// We eliminate this case in a micro-pass.
     #[charon::opaque]
     Adt(Option<VariantId>, Vec<ConstantExpr>),
+    /// A union constant, with the field to which we write, and the inner value.
+    /// We eliminate this case in a micro-pass.
+    #[charon::opaque]
+    Union(FieldId, Box<ConstantExpr>),
```
Nadrieril: I don't mind having these if you need them, but we can't ever get one of these from const-eval, can we? The active field is not a well-defined notion, recall the classic:
```rust
union Foo {
  a: (bool, u8),
  b: (u8, bool),
}
let mut x = Foo { a: (true, 42) };
x.b.0 = 42;
```

N1ark: yes i knew it was a little unsound but this example makes it a lot more clearly unsound 🫠 
would you prefer for me to opt into having `Union(Vec<CharonByte>)` with the following?
```rust
enum Provenance {
    Global(GlobalDeclRef),
    Function(FunDeclRef),
    Unknown
}

enum CharonByte {
    Uninit,
    Byte(u8),
    Provenance(Provenance),
}
```
It can always be a first step until we get closer to MiniRust values :)

Nadrieril: Yep, I'd like that :) Actually `ConstantExprKind::RawMemory` should use that too, and at that point you don't need a `Union` variant.

N1ark: Sure! will do 🫡

PR #926, opened 2025-11-29
Distinguish VTable unsizes, add metadata

N1ark:
Based on https://github.com/AeneasVerif/hax/pull/6

Rename `UnsizingMetadata::VTablePtr` -> `VTableDirect`, and add to it an optional `GlobalDeclRef` for the associated VTable. Note I don't do anything about the global's generics right now, which I believe is wrong. There are also cases where the global is not known because the impl atom isn't `ImplExprAtom::Concrete`; I have no idea how to fix this, maybe @ssyram knows, but for now it's already better than nothing :) 

Add `UnsizingMetadata::VTableNested`, for `dyn`-`dyn` unsizes. It also carries a `TraitRef`, along with an _optional_ `FieldId`, for the field in the current VTable that contains the sub-vtable to unsize to. Note here that the field id is optional, in the case of no-op unsizes (e.g. `dyn T + 'static` -> `dyn T`).

I believe we should distinguish these two cases, since fundamentally they operate differently: the former "creates" metadata from a statically known VTable address, the latter fetches a VTable dynamically, from the current VTable in the metadata.

To obtain the `FieldId` in `VTableNested` I had to slightly reorganise the way fields are added to the vtable structure (commit 07a71ed13e08ffada3f7e5ca2f943dcdc9086b73)

(disregard the branch name, this is a first step towards mono `dyn` but I didn't get there)


charon/src/ast/expressions.rs
```diff
@@ -222,7 +222,8 @@ pub enum CastKind {
 #[charon::variants_prefix("Meta")]
 pub enum UnsizingMetadata {
     Length(ConstGeneric),
-    VTablePtr(TraitRef),
+    VTableDirect(TraitRef, Option<GlobalDeclRef>),
+    VTableNested(TraitRef, Option<FieldId>),
```
Nadrieril: This is looking more and more like `UnsizingMetadata` is a defunctionalization x) What I'd actually want instead of `ptr := unsize(ptr, UnsizingMetadata::VTableNested("field1"))` is `ptr := ptr with metadata (copy ptr.metadata.field1)`. I'm not saying you have to do that in this PR though, and we should be careful not to introduce a reborrow that wasn't there.

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -566,28 +566,42 @@ impl BodyTransCtx<'_, '_, '_> {
...
                                             tref,
                                             TransItemSourceKind::VTableInstance(
                                                 TraitImplSource::Normal,
                                             ),
                                         );
+                                        Some(GlobalDeclRef {
+                                            // FIXME: this is wrong, I believe
+                                            generics: Box::new(GenericArgs::empty()),
+                                            id,
+                                        })
```
Nadrieril: You should be able to get the right generics using `translate_item` instead of `register_item`.

charon/tests/ui/simple/unsize-dyn-dyn.rs
Nadrieril, line 1: nit: calling this unsizing feels wrong because we're not going from sized to unsized. I'd prefer `dyn-cast-to-supertrait.rs`

charon/src/ast/expressions.rs
Nadrieril: Also the `TraitRef` is basically useless once we have the `GlobalDeclRef`, so I propose we add `ConstantExprKind::Vtable` and change `VTableDirect` to take only a `ConstantExpr`.

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -566,28 +566,42 @@ impl BodyTransCtx<'_, '_, '_> {
...
                                     }
                                     // TODO(dyn): more ways of registering vtable instance?
                                     _ => {
                                         trace!(
                                             "impl_expr not triggering registering vtable: {:?}",
                                             impl_expr
-                                        )
+                                        );
+                                        None
```
Nadrieril: You'll notice that even if we return `ConstantExprKind::VTable` in this branch, we can't register the vtable instance since we don't know the impl. That's a fundamental incompleteness in the choice I made to translate vtable instances lazily: it won't work for:
```rust
fn foo<T: Trait>(x: &T) -> &dyn Trait { x }
fn main() {
    let _ = foo::<SomeType>(...);
}
```
You might want a `--always-translate-vtable-instances` for your purposes. Or maybe we should do a static analysis at the end of translation to register all the ones that might be used indirectly like that :thinking:. Or maybe if we see any `dyn Trait` anywhere then we just register all the vtable instances for that trait, kinda like how we deal with lazily translating methods. Yeah that's saner.

charon/src/ast/expressions.rs
Nadrieril: Ah, something else that can happen: `dyn Trait -> dyn Sized`, despite the fact that `Sized` is not a supertrait of `Trait`. We might need to store a vtable pointer to `dyn Sized` in every vtable.

N1ark:
> Thanks for this PR! But I also totally agree with @Nadrieril that we in fact cannot always have that explicit field of `GlobalRef`, except monomorphised codes... Especially with:
> 
> ```rust
> fn to_dyn<T : Trait>(t : &T) -> &dyn Trait { t }
> ```

This is not a `VTableNested` unsize, since it's a non-dyn to dyn unsize. In all possible `dyn`-`dyn` unsizes I believe you always necessarily know what the field is? Regardless I'm not sure how a client is expected to perform the unsize without having the field

charon/src/ast/expressions.rs
Nadrieril: I decided to ignore that for now and just keep the same vtable.

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -566,28 +569,118 @@ impl BodyTransCtx<'_, '_, '_> {
...
+                                else {
+                                    raise_error!(
+                                        self,
+                                        span,
+                                        "Nested vtable unsizing only supported for dyn traits, got {:?}",
+                                        src_hax_ty
+                                    );
+                                };
+
+                                // Add a binder that contains the existentially quantified type.
```
Nadrieril: Can you use `translate_dyn_binder` here instead of managing the binder levels yourself?

PR #927, opened 2025-11-29
Change `FnOperand::Move` -> `Dynamic`

N1ark:
Remove `FnOperand::Move(Place)`, replaced by `FnOperand::Dynamic(Operand)`. 

One of the reasons I wanted to do this is that VTable calls were using a `move`, which is a bit awkward given you can't really move out from the `'static` vtable.

I also believe this may help us accept more code.

ci: use https://github.com/AeneasVerif/aeneas/pull/662
ci: use https://github.com/AeneasVerif/eurydice/pull/362


Nadrieril (approved):
Yeah that's much better, ty

PR #950, opened 2025-12-15
Use copy_files instead of symlink for dune to be happy

giltho:
I want to use the dune "package management" (which is kind of a replacement for using opam separately). So far, [it fails](https://github.com/soteria-tools/soteria/pull/15), and it seems to be because of [this issue](https://github.com/ocaml/dune/issues/9873), where dune has issues dealing with symlinks.

This PR updates charon to avoid requiring a symlink for the tests, and uses dune's `copy_files` instead

### DIFF
--- a/nix/charon-ml.nix
+++ b/nix/charon-ml.nix
@@ -19,13 +19,14 @@ let
     buildInputs = [ ocamlPackages.calendar ];
   };
 
-  # We need both `charon-ml` and the `dune-project` file.
+  # We need `charon-ml`, the `dune-project` file, and files that are used in tests.
   src = lib.cleanSourceWith {
     src = ./..;
     filter =
       path: type:
       (lib.hasPrefix (toString ../charon-ml) path)
-      || (lib.hasPrefix (toString ../dune-project) path);
+      || (lib.hasPrefix (toString ../dune-project) path)
+      || (lib.hasPrefix (toString ../charon/tests/ui) path);
   };
 
   charon-name_matcher_parser =


nix/charon-ml.nix
```diff
@@ -72,11 +72,6 @@ let
       inherit src;
 
       OCAMLPARAM = "_,warn-error=+A"; # Turn all warnings into errors.
-      preCheck =
-        if doCheck then ''
-          ln -sf ${charon}/tests-llbc charon-ml/tests/test-outputs
-        '' else
-          "";
```
Nadrieril, line 79: Something like this is needed because in the nix env the test outputs are otherwise not accessible

Nadrieril: You cuold `rm -r test-outputs` first, that should work

giltho: Hmm, not sure what that means
Why does it need to be available for nix? Isn't dune taking care of all of this?

Nadrieril: In nix we build the rust side and ocaml sides separately. So when we get to building the ocaml side, the `llbc` files emitted by running the rust tests are not there. We must put them there ourselves, which is what this does.

nix/charon-ml.nix
```diff
@@ -19,13 +19,14 @@ let
...
-  # We need both `charon-ml` and the `dune-project` file.
+  # We need `charon-ml`, the `dune-project` file, and files that are used in tests.
   src = lib.cleanSourceWith {
     src = ./..;
     filter =
       path: type:
       (lib.hasPrefix (toString ../charon-ml) path)
-      || (lib.hasPrefix (toString ../dune-project) path);
+      || (lib.hasPrefix (toString ../dune-project) path)
+      || (lib.hasPrefix (toString ../charon/tests/ui) path);
```
Nadrieril, line 29: That's not needed: ocaml only needs `.llbc` files, none of which are part of the git repo.

nix/charon-ml.nix
Nadrieril: Anyway I can do the nix bits, but I'm not sure I want two copies to happen on the ocaml side

giltho: Aaaaah, got it
Closing this anyway until it works one the dune side

PR #965, opened 2026-01-10
Make minor updates

sonmarcho:
This PR introduces a minor change which is worth mentioning by updating the formatting of body regions in Charon-ML. Instead of displaying, e.g., `&'0 ...`, we display `&°0 ...`. I'm not super happy with the use of the symbol `°` and it can be changed: the only important thing is that I need to use a different symbol for the bound regions and for the body regions, otherwise the logs can be quite confusing when debugging.

ci: use https://github.com/AeneasVerif/eurydice/pull/367
ci: use https://github.com/AeneasVerif/aeneas/pull/691

### DIFF
--- a/charon-ml/src/NameMatcher.ml
+++ b/charon-ml/src/NameMatcher.ml
@@ -798,7 +798,7 @@ let match_fn_ptr (ctx : 'fun_body ctx) (c : match_config) (p : pattern)
         | TraitImplItem (_, trait_ref, method_name, _)
           when c.match_with_trait_decl_refs ->
             let subst =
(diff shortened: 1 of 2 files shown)


sonmarcho:
> FYI they were already different: `'_42` is an anonymous bound region whereas `'42` is a body region (the idea being that `_42` is a valid rust identifier but `42` isn't). Did you not notice or was that not clear enough still?

I see, thanks for this! I found the issue in a case where there were only body regions and it was unclear to me whether they were body regions or anonymous regions. Had there been both in the context I would probably have noticed the difference. I'm still in favour of having a more discriminative formatting.

PR #1000, opened 2026-02-04
Fix local assignment being removed when removing asserts

N1ark:
Compare `invalid-reconstruct-assert.out` to see the difference; before the fix, the assignment to `_39` was removed, despite the local being used later. This pattern occurs in `slice::binary_search_by`, I'm unsure how to cause it manually so the repr is a bit big, sorry!

### DIFF
--- a/charon/src/transform/resugar/reconstruct_fallible_operations.rs
+++ b/charon/src/transform/resugar/reconstruct_fallible_operations.rs
@@ -4,13 +4,64 @@
 //! compiling for release). In our case, we take this into account in the semantics of our
 //! array/slice manipulation and arithmetic functions, on the verification side.
 
+use std::collections::HashSet;
+
 use derive_generic_visitor::*;
 
 use crate::ast::*;
+use crate::ids::IndexVec;
 use crate::transform::TransformCtx;
-use crate::ullbc_ast::{ExprBody, Statement, StatementKind};
+use crate::ullbc_ast::{BlockId, ExprBody, Statement, StatementKind};
 
 use crate::transform::ctx::UllbcPass;
+
+type LocalUses = IndexVec<BlockId, HashSet<LocalId>>;
+
+/// Compute for each block the locals that are assumed to have been initialized/defined before entering it.
+fn compute_uses(body: &ExprBody) -> LocalUses {
+    #[derive(Visitor)]
+    struct UsedLocalsVisitor<'a>(&'a mut HashSet<LocalId>);
+
+    impl VisitBody for UsedLocalsVisitor<'_> {
+        fn visit_place(&mut self, x: &Place) -> ::std::ops::ControlFlow<Self::Break> {
+            if let Some(local_id) = x.as_local() {
+                self.0.insert(local_id);
+            }
+            self.visit_inner(x)
+        }
+    }
+
+    body.body.map_ref(|block| {
+        let mut uses = HashSet::new();
+        let mut visitor = UsedLocalsVisitor(&mut uses);
+
+        // do a simple live variable analysis by walking the block backwards
+        for statement in block.statements.iter().rev() {
+            match &statement.kind {
+                StatementKind::Assign(place, rval) => {
+                    // We clear the assigned place, but it may be re-added
+                    // if it's used in rval
+                    if let Some(local_id) = place.as_local() {
+                        visitor.0.remove(&local_id);
+                    }
+                    let _ = rval.drive_body(&mut visitor);
+                }
+                StatementKind::StorageLive(local) | StatementKind::StorageDead(local) => {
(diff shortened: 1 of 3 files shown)


charon/src/transform/resugar/reconstruct_fallible_operations.rs
```diff
@@ -4,14 +4,42 @@
...
 
+type LocalUses = IndexVec<BlockId, HashSet<LocalId>>;
+
+fn compute_uses(body: &ExprBody) -> LocalUses {
+    enum Never {}
+    struct UsedLocalsVisitor(HashSet<LocalId>);
+
+    impl Visitor for UsedLocalsVisitor {
+        type Break = Never;
+    }
```
Nadrieril: FYI `#[derive(Visitor)]` does exactly this, using `std::convert::Infallible` as the empty type

charon/src/transform/resugar/reconstruct_fallible_operations.rs
```diff
@@ -84,9 +112,21 @@ fn equiv_op(op_l: &Operand, op_r: &Operand) -> bool {
...
     _ctx: &mut TransformCtx,
+    uses: &LocalUses,
+    block_id: BlockId,
     locals: &mut Locals,
     statements: &mut [Statement],
 ) {
+    let used_elsewhere = |p: &Place| {
+        if let Some(local_id) = p.as_local() {
+            uses.iter_indexed()
+                .any(|(bid, used)| bid != block_id && used.contains(&local_id))
```
Nadrieril: I don't think that's what we want to check: the local could be used in the same block, inside `rest`

N1ark: true :) was being lazy, fixed

charon/src/transform/resugar/reconstruct_fallible_operations.rs
```diff
@@ -470,7 +535,20 @@ fn remove_dynamic_checks(
...
+        if let StatementKind::Assign(place, _) = &statements[i].kind
+            && let Some(local) = place.as_local()
+            && let mut statements_to_keep = statements[removed_len..].as_ref().iter()
+            && let mut other_blocks = uses.iter_indexed().filter(|(bid, _)| *bid != block_id)
+            && (other_blocks.any(|(_, used)| used.contains(&local))
+                || statements_to_keep.any(|st| uses_local(st, local)))
+        {
+            continue;
+        };
         statements[i].kind = StatementKind::Nop;
```
Nadrieril, line 552: Ok that's really nice, that will indeed avoid a ton of issues.

I'm ok to merge this, but I'm worried about perf: in a block with lots of arithmetic operations, we'll be traversing the statements quadratically I believe.

How about we count the number of uses of each local in the whole body. Then whenever we remove a statement we decrease the use-count of any locals it mentions, and when we get to removing the initial assignment, we only remove it if there are no uses of that local left.

N1ark: hmmm im not entirely sure i fully understand your solution; ill try implementing it tomorrow and you tell me if its what you had in mind

N1ark: so I did this here: https://github.com/soteria-tools/charon/commit/e3d3c814dad9b4080422daa7094dc021b13cb5a1
the issue is that for the example i had, the same locals are used for the length check on two different blocks, and just naively counting the uses will mean that when handling one of the two blocks, the check local will be considered "used" elsewhere, even though it's used in an assert we will remove (see line 286 of the test output diff)

N1ark: if that can serve as (slight) reassurance:
i made a file that is:
```rust
fn main() {
    let a = 0;
    let b = 1;
    let a = a + b;
    let b = b * 2;
    // ... 500 more lines
    let a = a + b;
}
```
here are the perf results :) these changes seem to have basically no impact on perf
|  | with reconstruct | without |
| ------------- | ------------- | -------- |
| `main`  | 0.629s  | 0.260s |
| this branch  | 0.627s  |  0.258s |

Nadrieril: > the issue is that for the example i had, the same locals are used for the length check on two different blocks, and just naively counting the uses will mean that when handling one of the two blocks, the check local will be considered "used" elsewhere, even though it's used in an assert we will remove (see line 286 of the test output diff)

Oh, and your current branch handles it correctly because of the little `StorageDead/Live` analysis... Interesting.

Fwiw I want charon to eventually use miri's std built without optimizations, so I don't care much about supporting optimized MIR. No problem merging this if that helps you in the meantime though. Thanks for the perf measurement, that does convince me.

PR #1001, opened 2026-02-04
Add metadata to `ConstantExprKind::Ref,RawPtr`

N1ark:
- Add an `Option<UnsizingMetadata>` to `Ref` and `RawPtr` constant expressions. This will, eventually, allow us to nicely represent globals/statics that are unsized (e.g. `&'static dyn _`). In `simplify_constants`, these then get lowered with an unsize if there is some metadata.
- Remove `ConstantExprKind::Slice` and make use of the a ref with unsizing metadata to an array.
- Add `--unsized-strings`, that allows representing strings as references to `[u8; N]` backing the data.

I'm not entirely sure `--unsized-strings` is sound however since if it returns a `&'static str` I guess the local gets uninitialised at the end of the scope?


Nadrieril (approved):
Very nice, that's one tricky issue fixed

PR #1009, opened 2026-02-07
Make assert information more precise

### LINKED ISSUE #868: Support more unwind nodes
PRs #682 and #700 respectively make `call` a terminator and add a reference to the unwinding block in ULLBC.

This process needs to be done again for the following ULLBC statements which are terminators in MIR and have unwind blocks:
- [ ] Drop
- [ ] Assert

(I realised I changed context and never noted anywhere that this wasn't finished.)

N1ark:
(waiting for a review before taking care of Aeneas / Eurydice)

As mentioned in #1000, relying on `reconstruct_fallible_operations` is risky and error-prone. However not having this pass enabled currently provided too little information to provide satisfactory user feedback for assertion failures.

We first translate the message associated with each MIR assert, to allow tools to explain what a failed assert may correspond to. This will also allow us to detect what asserts correspond to fallible operations.

Next we translate terminator asserts from MIR into a new `TerminatorKind::Assert`. We also remove `on_failure` from `Assert`; instead, `StatementKind::Assert` is now `{ assert: Assert, on_failure: AbortKind }`, and `TerminatorKind::Assert` is now `{ assert: Assert, target, on_unwind }`. 

Currently, we initially only translate `StatementKind::Assert` for the `Assume` non-diverging intrinsic; later passes then may lower a `TerminatorKind::Assert` into a `StatementKind`, e.g. for dynamic checks if `--reconstruct-fallible-operations` is set.

This mostly closes #868! The last couple cases where we lose edges right now is when converting calls to the language items `panic`, `panic_fmt`, `begin_panic` or to the functions `core::panicking::assert_failed` and `core::panicking::panic_explicit` into a `TerminatorKind::Abort(AbortKind::Panic)`. The same happens in `inline_local_panic_functions.rs`. These function calls have an unwind edge, that is needed e.g. to drop any remaining locals after starting the panic. An option is adding an optional `on_unwind` to `TerminatorKind::Abort`.

Small sidenote but I also find it weird we always convert some of these function calls to aborts in `translate_bodies`; supposedly this should be done as a micro-pass along with `inline_local_panic_functions.rs` if a specific flag is set (something like `--simplify-panics`). Technically the only 'magic' associated with panics should be in `__rust_start_panic`

ci: use https://github.com/AeneasVerif/aeneas/pull/745
ci: use https://github.com/AeneasVerif/eurydice/pull/376

### DIFF
--- a/charon/src/ast/ullbc_ast.rs
+++ b/charon/src/ast/ullbc_ast.rs
@@ -122,6 +122,7 @@ pub enum TerminatorKind {
     },
     /// Assert that the given condition holds, and if not, unwind to the given block. This is used for
     /// bounds checks, overflow checks, etc.
+    #[charon::rename("TAssert")]
     Assert {
         assert: Assert,
         target: BlockId,
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.166"
+let supported_charon_version = "0.1.167"
--- a/charon-ml/src/PrintGAst.ml
+++ b/charon-ml/src/PrintGAst.ml
@@ -28,11 +28,16 @@ let call_to_string (env : 'a fmt_env) (indent : string) (call : call) : string =
   let dest = place_to_string env call.dest in
   indent ^ dest ^ " := move " ^ func ^ args
 
-let assertion_to_string (env : 'a fmt_env) (indent : string) (a : assertion) :
-    string =
+let assertion_to_string (env : 'a fmt_env) (a : assertion) : string =
   let cond = operand_to_string env a.cond in
-  if a.expected then indent ^ "assert(" ^ cond ^ ")"
-  else indent ^ "assert(¬" ^ cond ^ ")"
+  if a.expected then "assert(" ^ cond ^ ")" else "assert(¬" ^ cond ^ ")"
+
+let abort_kind_to_string (env : 'a fmt_env) (a : abort_kind) : string =
+  match a with
+  | Panic None -> "panic"
+  | Panic (Some name) -> "panic(" ^ name_to_string env name ^ ")"
+  | UndefinedBehavior -> "undefined_behavior"
+  | UnwindTerminate -> "unwind_terminate"
 
 (** Small helper *)
 let fun_sig_with_name_to_string (env : 'a fmt_env) (indent : string)
--- a/charon-ml/src/PrintLlbcAst.ml
+++ b/charon-ml/src/PrintLlbcAst.ml
@@ -37,7 +37,11 @@ module Ast = struct
         indent ^ "storage_dead " ^ local_id_to_string env var_id
     | Deinit p -> indent ^ "deinit " ^ place_to_string env p
     | Drop (p, _, _) -> indent ^ "drop " ^ place_to_string env p
-    | Assert a -> assertion_to_string env indent a
+    | Assert (asrt, abort_kind) ->
+        indent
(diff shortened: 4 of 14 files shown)


charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -938,18 +939,21 @@ impl<'tcx> BodyTransCtx<'tcx, '_, '_> {
                 expected,
                 msg,
                 target,
-                unwind: _, // We model unwinding as an effet, we don't represent it in control flow
+                unwind, // We model unwinding as an effet, we don't represent it in control flow
```
Nadrieril:
```suggestion
                unwind,
```

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -1105,6 +1110,7 @@ impl<'tcx> BodyTransCtx<'tcx, '_, '_> {
                     // I don't know in which other cases it can be `None`.
                     assert!(target.is_none());
                     // We ignore the arguments
+                    // TODO: shouldn't we do something with the unwind edge?
```
Nadrieril, line 1113: Yeah, `Abort` should have an unwind edge actually. I'm really not sure that this particular reconstruction we're doing here is useful honestly, this is just a function call that can be caught by tools.

charon/src/transform/control_flow/ullbc_to_llbc.rs
```diff
@@ -1013,6 +1015,31 @@ impl<'a> ReconstructCtx<'a> {
...
+                let then_block = self.translate_jump(terminator.span, then_tgt);
+                let else_block = self.translate_jump(terminator.span, else_tgt);
+                let switch = tgt::Switch::If(cond.clone(), then_block, else_block);
+
+                // Return
+                let span = tgt::combine_switch_targets_span(&switch);
+                let span = combine_span(&src_span, &span);
+                let st = tgt::StatementKind::Switch(switch);
+                tgt::Statement::new(span, st).into_block()
+            }
```
Nadrieril, line 1031: I want to use `StatementKind::Assert` here too. That loses the unwind info for now, like it does with the rest of LLBC.

charon/src/ast/ullbc_ast.rs
```diff
@@ -45,10 +45,15 @@ pub enum StatementKind {
...
-    Assert(Assert),
+    /// A non-diverging runtime check for a condition. This can be either:
+    /// - Emitted for inlined "assumes" (which cause UB on failure)
+    /// - Reconstructed from `if b { panic() }` if `--reconstruct-asserts` is set.
+    /// This statement comes with the effect that happens when the check fails
+    /// (rather than representing it as an unwinding edge).
+    Assert {
+        assert: Assert,
+        on_failure: AbortKind,
+    },
```
Nadrieril, line 56: Hm, having `assert` at all felt like a hack because it's a glorified `if`. Having it both as a statement and a terminator feels even less motivated. Long-term I'd like to remove both and just use `if` instead honestly. Not asking you to do this here tho.

charon/src/bin/charon-driver/translate/translate_bodies.rs
Nadrieril: Let's keep it for now, `Abort` is a useful primitive for when we'll turn `Assert` into a plain `if`.

charon/src/transform/control_flow/ullbc_to_llbc.rs
N1ark: Fixed :) LLBC is now I think exactly the same as before

N1ark:
sounds good! and yeah agreed having both `StatementKind::Assert` and `TerminatorKind::Assert` is not ideal ^^' fixed your comments, should be ready to merge

edit: forgot about aeneas and eurydice :P will do these tomorrow

N1ark:
oops, rebased

Issue #503, opened 2024-12-20
Revert the NameMatcher's hack


sonmarcho:
Once [this issue](https://github.com/AeneasVerif/charon/issues/495) is solved, we should revert [this PR](https://github.com/AeneasVerif/charon/pull/502).

Nadrieril:
#495 has been solved, but unfortunately there are still mismatches in function generics around closures. This is tracked in #194.

PR #1019, opened 2026-02-12
Fix an issue in the Charon-ML name matcher

sonmarcho:
The logic to match generic arguments was plain wrong. I don't remember why I did this, but it seems related to https://github.com/AeneasVerif/charon/issues/503 somehow.

There are still quite a few issues with the name matcher, I thus opened a corresponding issue: https://github.com/AeneasVerif/charon/issues/1024

### DIFF
--- a/charon/tests/ui/ml-mono-name-matcher-tests.rs
+++ b/charon/tests/ui/ml-mono-name-matcher-tests.rs
@@ -53,17 +53,16 @@ impl<T, U> Trait<Box<T>> for Option<U> {
 
 // `call[i]` is a hack to be able to refer to `Call` statements inside the function body.
 #[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<'_, i32>")]
-// Regions can be omitted
-#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<i32>")]
-#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<@T>")]
-// Generic arguments are not required.
-#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some")]
+// TODO: this should pass
+#[pattern::fail(call[0], "core::option::{core::option::Option<@T>}::is_some<_, i32>")]
+#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<'_, @T>")]
 #[pattern::pass(call[1], "ArrayToSliceShared<'_, bool, 1>")]
 // This is a trait instance call.
 #[pattern::pass(call[2], "core::ops::index::Index<[bool], core::ops::range::RangeFrom<usize>>::index")]
 #[pattern::pass(call[2], "core::ops::index::Index<[@T], @I>::index")]
 // We can reference the method directly.
-#[pattern::pass(call[2], "core::slice::index::{core::ops::index::Index<[@T], @I>}::index<bool, core::ops::range::RangeFrom<usize>>")]
+// TODO: this should pass
+#[pattern::fail(call[2], "core::slice::index::{core::ops::index::Index<[@T], @I>}::index<bool, core::ops::range::RangeFrom<usize>>")]
 fn foo() {
     let _ = Some(0).is_some();
     let slice: &[bool] = &[false];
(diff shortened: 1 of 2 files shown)


charon/tests/ui/ml-mono-name-matcher-tests.rs
```diff
@@ -53,17 +53,16 @@ impl<T, U> Trait<Box<T>> for Option<U> {
 
 // `call[i]` is a hack to be able to refer to `Call` statements inside the function body.
 #[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<'_, i32>")]
-// Regions can be omitted
-#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<i32>")]
-#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<@T>")]
-// Generic arguments are not required.
-#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some")]
```
Nadrieril, line 60: Could you leave these tests, even if you have to turn them into failures? The last one in particular is very important to keep passing

charon/tests/ui/ml-mono-name-matcher-tests.rs
```diff
@@ -53,17 +53,16 @@ impl<T, U> Trait<Box<T>> for Option<U> {
...
+#[pattern::fail(call[0], "core::option::{core::option::Option<@T>}::is_some<_, i32>")]
+#[pattern::pass(call[0], "core::option::{core::option::Option<@T>}::is_some<'_, @T>")]
 #[pattern::pass(call[1], "ArrayToSliceShared<'_, bool, 1>")]
 // This is a trait instance call.
 #[pattern::pass(call[2], "core::ops::index::Index<[bool], core::ops::range::RangeFrom<usize>>::index")]
 #[pattern::pass(call[2], "core::ops::index::Index<[@T], @I>::index")]
 // We can reference the method directly.
-#[pattern::pass(call[2], "core::slice::index::{core::ops::index::Index<[@T], @I>}::index<bool, core::ops::range::RangeFrom<usize>>")]
+// TODO: this should pass
+#[pattern::fail(call[2], "core::slice::index::{core::ops::index::Index<[@T], @I>}::index<bool, core::ops::range::RangeFrom<usize>>")]
```
Nadrieril, line 65: Do you understand what changed here? That looks like a pretty major regression no?

charon/tests/ui/ml-mono-name-matcher-tests.rs
sonmarcho: The last one currently fails because now we always take into account the generic arguments of functions. Of course we need to be able to refer to functions without giving their generic arguments, but I do not think we should do this at the level of the name matcher but rather by instrumenting the way we call the name matcher (this is what Aeneas does): otherwise we easily end up with a dubious logic like the one that was currently implemented, where it was possible to ignore generic arguments altogether, causing other kinds of issues.
But I can put them back by turning them to failures if you really want.

charon-ml/src/NameMatcher.ml
```diff
@@ -685,27 +685,9 @@ and match_generic_args (ctx : 'fun_body ctx) (c : match_config) (m : maps)
         List.map (fun x -> MCg x) generics.const_generics;
       ]
   in
-  if List.length pgenerics = 0 then true (* Generics can be omitted *)
```
Nadrieril, line 688: Why change that? That seem important to me. Maybe we should distinguish `function<>` (matches function if it have no generics) and `function` (matches function regardless of generics)?

charon/tests/ui/ml-mono-name-matcher-tests.rs
Nadrieril: I do really want the tests even if they're failures yes, and I also want a way to match ignoring generics, however that's implemented

charon-ml/src/NameMatcher.ml
sonmarcho: See my first comment: this case shouldn't be handled by the name matcher because it is completely ad hoc and leads to bugs. If we want to ignore arguments, we should call the name matcher without giving it arguments (this is what Aeneas does).

charon/tests/ui/ml-mono-name-matcher-tests.rs
sonmarcho: I'm really annoyed with this kind of tests which I think should not be baked into the logic of the name matcher. The problem is that sometimes when matching a pattern against a trait impl we want to match the pattern against the trait decl implemented by the impl, not the impl itself. This is perfectly fine when matching against *impl blocks* but I'm not convinced by the logic of the name matcher when matching a pattern against a *function* name. First of all, the handling of the generic arguments given to the function was wrong (and still is): the only reason it worked is that we actually completely ignored them, while we shouldn't (see my other comments). This led to bugs in Aeneas. Moreover, I think it is more natural (and less ad-hoc) to instrument the calls to the name matchers whenever we want to check whether a pattern matches a function (or method) (cf my other comments).

Issue #874, opened 2025-10-30
Support `StatementKind::PlaceMention`


N1ark:
Though a comment in  Charon's driver seems to imply [`StatementKind::PlaceMention`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/enum.StatementKind.html#variant.PlaceMention) is there "only to make borrow-checking accept less code", which is not entirely true I think.
In particular it requires the place to be valid, so one of the projections may be invalid within and this UB is lost if we compile this statement away.
In [MiniRust](https://github.com/minirust/minirust/blob/master/spec/lang/step/statements.md#mentioning-a-place-without-accessing-it) this statement is mentioned so I'm not sure we can avoid it.

This is an exemple of UB code that we would miss:
```rust
let p = {
    let b = Box::new(42);
    &*b as *const i32 as *const (u8, u8, u8, u8)
};
unsafe {
    let _ = (*p).1; //~ ERROR: in-bounds pointer arithmetic failed
}
```

There are a few Miri tests that I believe check for exactly this, i.e. a tool using Charon cannot pass them:
- https://github.com/rust-lang/miri/blob/fa8b9658497810f5ee5db69b437b062fbc5e527c/tests/fail/dangling_pointers/dangling_pointer_project_underscore_let.rs
- https://github.com/rust-lang/miri/blob/fa8b9658497810f5ee5db69b437b062fbc5e527c/tests/fail/dangling_pointers/dangling_pointer_project_underscore_let_type_annotation.rs
- https://github.com/rust-lang/miri/blob/fa8b9658497810f5ee5db69b437b062fbc5e527c/tests/fail/dangling_pointers/dangling_pointer_project_underscore_match.rs

PR #1023, opened 2026-02-14
Add `PlaceMention`

N1ark:
Fixes https://github.com/AeneasVerif/charon/issues/874

Remove `StatementKind::Deinit`, which wasn't used anywhere anymore
Add `StatementKind::PlaceMention(Place)`, for when a place is mentioned (see [documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.StatementKind.html#variant.PlaceMention)). 

Although it may also serve other purposes when borrow checking (it's not clear to me), my main use of it is ensuring UB in code such as `let _ = <some invalid place>` gets caught. As such, I have disabled translation of it for cases where the place is just a local, as that created quite a bit of noise, and to my knowledge doesn't really enable anything more analysis-wise? Maybe I'm mistaken.

ci: use https://github.com/AeneasVerif/aeneas/pull/771
ci: use https://github.com/AeneasVerif/eurydice/pull/379

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_bodies.rs
+++ b/charon/src/bin/charon-driver/translate/translate_bodies.rs
@@ -867,6 +867,8 @@ impl<'tcx> BodyTransCtx<'tcx, '_, '_> {
             }
             mir::StatementKind::PlaceMention(place) => {
                 let place = self.translate_place(span, place)?;
+                // We only translate this for places with projections, as
+                // no UB can arise from simply mentioning a local variable.
                 if place.is_local() {
                     None
                 } else {


charon/tests/ui/assert-kinds-reconstruct-fallible.out
```diff
@@ -6,47 +6,44 @@ fn main()
...
+        storage_live(_4);
         _0 = ();
         storage_live(_1);
         _1 = [const 1 : i32, const 2 : i32, const 3 : i32];
         storage_live(_2);
         _2 = const 0 : usize;
-        _3 = copy _2 < const 3 : usize;
-        assert(move _3 == true) (bounds_check) else panic;
+        _ = _1[copy _2];
         storage_dead(_2);
```
N1ark, line 25: a nice thing with this addition is that because the place is found to be used later, the bound check is properly removed here :) (which makes sense since it is done implicitly by the access)

charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -867,7 +867,11 @@ impl<'tcx> BodyTransCtx<'tcx, '_, '_> {
             }
             mir::StatementKind::PlaceMention(place) => {
                 let place = self.translate_place(span, place)?;
-                Some(StatementKind::PlaceMention(place))
+                if place.is_local() {
+                    None
```
Nadrieril, line 873: Could you add a comment justifying why we can skip that case?

Nadrieril: (I suggest adding a new commit instead of rebasing, otherwise you'll have to update the eurydice and aeneas PRs too)

N1ark: done :)

PR #1030, opened 2026-02-22
Bump rustc to 2026-02-19

N1ark:
- Bump the rust toolchain version to 2026-02-19!

In smaller bits, ordered chronologically
- Slight change to filename parsing
  https://github.com/rust-lang/rust/pull/149709
- Slightly change due to a change to `TyCtx::destructure_const`
  https://github.com/rust-lang/rust/pull/150411
- Remove the runtime check `NullOp`s, add `Operand::RuntimeCheck`
  https://github.com/rust-lang/rust/pull/148766
- Manually ignore `Destruct` parent traits for `dyn Trait`
  https://github.com/rust-lang/rust/pull/148637
- Tweaked `get_mir`, now that `is_ctfe_mir_available` is gone; amusingly i think this change means more functions are now translated (e.g. see the diff for `slice-index-range.out`, where `core::slice::index::slice_index_fail::do_panic#3::runtime`'s body is not missing anymore)
  https://github.com/rust-lang/rust/pull/152096
- ~~Remove `transform/resugar/reconstruct_boxes.rs`, thanks to the `Box::new` and `vec!` rework (see below)~~
  https://github.com/rust-lang/rust/pull/148190
  
  **After discussion this last bump is kept for a separate PR as it also involves changes to the transformation pass to reconstruct the `vec!` into something safe**

The main thing i need some attention for in the review is the change in `translate-traits`, to ignore `Destruct`: if you have a better fix for it let me know

ci: use https://github.com/AeneasVerif/aeneas/pull/796
ci: use https://github.com/AeneasVerif/eurydice/pull/381

### DIFF
--- a/charon/hax-frontend/src/types/mir.rs
+++ b/charon/hax-frontend/src/types/mir.rs
@@ -75,8 +75,7 @@ pub mod mir_kinds {
                 id: DefId,
                 f: impl FnOnce(&Body<'tcx>) -> T,
             ) -> Option<T> {
-                tcx.is_ctfe_mir_available(id)
-                    .then(|| f(tcx.mir_for_ctfe(id)))
+                (!tcx.is_trivial_const(id)).then(|| f(tcx.mir_for_ctfe(id)))
             }
         }
 
--- a/charon/src/bin/charon-driver/translate/get_mir.rs
+++ b/charon/src/bin/charon-driver/translate/get_mir.rs
@@ -99,17 +99,17 @@ fn get_mir_for_def_id_and_level<'tcx>(
                     | hax::DefKind::InlineConst
             );
             let mir_available = tcx.is_mir_available(rust_def_id);
-            let ctfe_mir_available = tcx.is_ctfe_mir_available(rust_def_id);
-            // For globals, prefer ctfe_mir when both are available.
-            let body = if mir_available && !(is_global && ctfe_mir_available) {
-                tcx.optimized_mir(rust_def_id).clone()
-            } else if ctfe_mir_available {
-                tcx.mir_for_ctfe(rust_def_id).clone()
+
+            if mir_available && !is_global {
+                Some(tcx.optimized_mir(rust_def_id).clone())
+            } else if (is_global && !tcx.is_trivial_const(rust_def_id))
+                || tcx.is_const_fn(rust_def_id)
+            {
+                Some(tcx.mir_for_ctfe(rust_def_id).clone())
             } else {
-                trace!("no mir available");
-                return None;
-            };
-            Some(body)
+                trace!("mir not available for {:?}", rust_def_id);
+                None
+            }
         }
         hax::DefIdBase::Promoted(rust_def_id, promoted_id) => {
             Some(hax::get_promoted_mir(tcx, rust_def_id, promoted_id))
--- a/charon/tests/ui/simple/gat-default.out
+++ b/charon/tests/ui/simple/gat-default.out
@@ -1,5 +1,5 @@
 
-thread 'rustc' panicked at /rustc-dev/eda76d9d1d133effbf7facb28168fd78d75fd434/compiler/rustc_type_ir/src/binder.rs:784:9:
+thread 'rustc' panicked at /rustc-dev/efc9e1b50cbf2cede7ebe25f0a1fc64fd8b3e942/compiler/rustc_type_ir/src/binder.rs:793:9:
 type parameter `U/#1` (U/#1/1) out of range when instantiating, args=[()]
 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 error: Hax panicked when translating `test_crate::{impl#0}`.
@@ -9,7 +9,7 @@ error: Hax panicked when translating `test_crate::{impl#0}`.
   | ^^^^^^^^^^^^^^^^^^^^^^
 
 
-thread 'rustc' panicked at /rustc-dev/eda76d9d1d133effbf7facb28168fd78d75fd434/compiler/rustc_type_ir/src/binder.rs:784:9:
+thread 'rustc' panicked at /rustc-dev/efc9e1b50cbf2cede7ebe25f0a1fc64fd8b3e942/compiler/rustc_type_ir/src/binder.rs:793:9:
 type parameter `U/#1` (U/#1/1) out of range when instantiating, args=[()]
 error: Hax panicked when translating `test_crate::{impl#0}`.
  --> tests/ui/simple/gat-default.rs:9:1
--- a/charon/hax-frontend/src/constant_utils/uneval.rs
+++ b/charon/hax-frontend/src/constant_utils/uneval.rs
@@ -114,9 +114,8 @@ pub fn eval_ty_constant<'tcx, S: UnderOwnerState<'tcx>>(
 impl<'tcx, S: UnderOwnerState<'tcx>> SInto<S, ConstantExpr> for ty::Const<'tcx> {
     #[tracing::instrument(level = "trace", skip(s))]
     fn sinto(&self, s: &S) -> ConstantExpr {
-        use rustc_middle::query::Key;
         let tcx = s.base().tcx;
-        let span = self.default_span(tcx);
+        let span = rustc_span::DUMMY_SP;
         match self.kind() {
             ty::ConstKind::Param(p) => {
                 let ty = p.find_const_ty_from_env(s.param_env());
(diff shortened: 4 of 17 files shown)


charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -284,6 +284,16 @@ impl ItemTransCtx<'_, '_> {
...
 
+    fn assert_is_destruct(&self, tref: &hax::TraitRef) {
+        assert!(
+            tref.def_id
+                .as_rust_def_id()
+                .is_some_and(|id| self.tcx.is_lang_item(id, rustc_hir::LangItem::Destruct)),
+            "unexpected non-dyn compatible supertrait: {:?}",
+            tref.def_id
+        );
+    }
```
N1ark, line 300: had to add this check, because a bunch of `dyn Trait`s had a `Destruct` parent clause (i think?) [which is not dyn-compatible anymore](https://github.com/rust-lang/rust/pull/148637). The right fix is probably to remove whatever we do in hax that adds this trait bound, I just couldn't locate it so let me know

Nadrieril: It's fine to lie about dyn-compatibility in Charon though, I think that would be easier

Nadrieril: Oh wait that's exactly what you did. Sounds good to me, tho please add a comment explaining what's happening

charon/tests/cargo/error-dependencies.out
```diff
@@ -5,7 +5,7 @@ error: Error parsing attribute: Unrecognized attribute: `charon::error`
   | ^^^^^^^^^^^^^^^^
 
 error: Error parsing attribute: Unrecognized attribute: `charon::error`
- --> ./tests/cargo/error-dependencies/error-crate/src/lib.rs:4:1
+ --> error-crate/src/lib.rs:4:1
```
Nadrieril, line 9: Nicee, that path was wrong

charon/tests/ui/simple/gat-default.out
```diff
@@ -1,5 +1,5 @@
 
-thread 'rustc' panicked at /rustc-dev/94b49fd998d6723e0a9240a7cff5f9df37b84dd8/compiler/rustc_type_ir/src/binder.rs:781:9:
+thread 'rustc' panicked at compiler/rustc_type_ir/src/binder.rs:782:9:
```
Nadrieril: That one I don't like, this path looks relative to the current dir but isn't. Is there a different `RemapPathScopeComponents` that gives us the old path or at least not a a mistakeable path?

charon/src/ast/expressions.rs
```diff
@@ -368,6 +372,8 @@ pub enum Operand {
     /// Constant value (including constant and static variables)
     #[charon::rename("Constant")]
     Const(Box<ConstantExpr>),
+    /// Whether a particular runtime check should be performed or not.
+    RuntimeCheck(RuntimeCheck),
 }
```
Nadrieril, line 371: Adding a new operand is a big burden on downstream users since they'll all need to handle them. I would like these to be constants instead, unless there's a reason not to?

Nadrieril: Ah wait no:
> It seems slightly odd to call these things "constants" when they have a different value inside and outside of const blocks...

N1ark: yeah that's why i chose to follow rust's choice, i dont mind having them as constants but it is *odd*

charon/tests/ui/simple/gat-default.out
N1ark: ive tried every `RemapPathScopeComponents`, from `::empty()` to `::all()` and nothing change this :') it might be that what i do is wrong

charon/src/ast/expressions.rs
Nadrieril: I think I'd rather keep emitting `NullOp`s

charon/tests/ui/simple/gat-default.out
Nadrieril: I think we should use `::empty()`, and possibly use the working directory returned from `embeddable_name` somehow? At least we return `Virtual` so this won't cause consumers to assume the path is real.

charon/src/ast/expressions.rs
N1ark: done :)

charon/tests/ui/simple/gat-default.out
Nadrieril: Somehow later bumps fix this, weird but ok

charon/tests/ui/issue-393-shallowinitbox.out
```diff
@@ -1,10 +1,161 @@
...
+        return
+    }
+    _5 = const 42 : u8
+    storage_dead(_6)
+    ((((*_4)).1).0).0 = [move _5]
+    storage_dead(_5)
+    _3 = move _4
+    conditional_drop[{impl Destruct for alloc::boxed::Box<T>[@TraitClause0, @TraitClause1]}<MaybeUninit<[u8; 1 : usize]>[{built_in impl Sized for [u8; 1 : usize]}], Global>[{built_in impl MetaSized for MaybeUninit<[u8; 1 : usize]>[{built_in impl Sized for [u8; 1 : usize]}]}, {built_in impl Sized for Global}]] _4
+    storage_dead(_4)
+    vec_2 = box_assume_init_into_vec_unsafe<u8, 1 : usize>[{built_in impl Sized for u8}](move _3)
```
Nadrieril: It's important that `vec![]` expands to safe code that aeneas can translate. So we'll still need a reconstruction pass, hopefully much simpler now. The only difference is that we should use `Box::<MaybeUninit<_>>::write` instead of writing by hand then calling `box_assume_init_into_vec_unsafe`.

charon/hax-frontend/src/types/mir.rs
```diff
@@ -75,8 +75,7 @@ pub mod mir_kinds {
                 id: DefId,
                 f: impl FnOnce(&Body<'tcx>) -> T,
             ) -> Option<T> {
-                tcx.is_ctfe_mir_available(id)
-                    .then(|| f(tcx.mir_for_ctfe(id)))
+                Some(f(tcx.mir_for_ctfe(id)))
```
Nadrieril: I'm worried about this causing ICEs... I don't know how to access the query tables ourselves to check that without the method. Maybe it's fine to leave is like that, we only call it for inline consts anyway.

N1ark: a safer check might be this, i just realised:
```rust
(!tcx.is_trivial_const(id)).then(|| f(tcx.mir_for_ctfe(id)))
```
although the current version doesn't seem to ICE so maybe it's ok and it never gets called in problematic ways

charon/src/bin/charon-driver/translate/get_mir.rs
```diff
@@ -99,17 +99,17 @@ fn get_mir_for_def_id_and_level<'tcx>(
...
-                tcx.optimized_mir(rust_def_id).clone()
-            } else if ctfe_mir_available {
-                tcx.mir_for_ctfe(rust_def_id).clone()
+
+            if mir_available && !is_global {
+                Some(tcx.optimized_mir(rust_def_id).clone())
+            } else if (is_global && !tcx.is_trivial_const(rust_def_id))
+                || tcx.is_const_fn(rust_def_id)
+            {
+                Some(tcx.mir_for_ctfe(rust_def_id).clone())
```
Nadrieril, line 108: Oof I'll have to think hard about this one. That was subtle logic to avoid ICEs, I'm not sure how to do it right without the method.

Nadrieril: I think I recall globals that had `optimized_mir` but not `ctfe_mir` for example... Tbf if that's not in the test suite it's also ok to pretend they don't exist and see

Nadrieril: Maybe you understand this better than me

N1ark: im not sure i do tbh ^^' i reached this code from trial-and-error on the test suite + looking around the rust codebases for uses of the functions; i dont think i encountered globals with an optimised body and no CTFE, so we'll see.. once this PR is stable and i start using it if i see any ICEs ill report/fix them :) 
i had tried having a `catch_unwind` around the `mir_for_ctfe` to "check" if it's going to ICE but that didn't work :P

Nadrieril: There's already a `catch_unwind` around `translate_body` so that's handled, but any compiler ICE poisons the compiler session so we can't fully recover.

From reading the rustc code the `trivial_const` check might be enough. I agree with merging and seeing what happens in practice

charon/hax-frontend/src/types/mir.rs
Nadrieril: Yes please

N1ark: done !

charon/tests/ui/issue-393-shallowinitbox.out
N1ark: this will be left for a separate PR; to have it written down somewhere:
the goal is to have a reconstruction of the following form (assuming the shape above):
```
_T = [move _5]
_3 = Box::write(move _4, move _T)
_B = unsize<Box<[T; N]>, Box<[T]>> _3
vec_2 = (((((*_B)).1).0).0).into_vec()
```

Issue #856, opened 2025-10-17
Feature request: Support `dyn Trait` with `--monomorphize`


Nadrieril:
This is unsupported currently because of the difficulty of representing partially-monomorphized traits (with fixed values for their type parameters and associated types, but with a `Self` parameter nonetheless). As a first step, @N1ark and I came up with the following proposal:

- when translating a vtable definition in mono mode, we use `poly_hax_def` for the trait but don't add any generic parameters to the vtable definition (this is subtle and will require an abundance of comments);
- in mono mode, all function fields of vtables are `*const ()` instead of `fn(..)`;
- therefore when creating the vtable we cast into `fn(..)` to `*const ()` and when using the vtable we cast the other way;
- when translating a trait decl, we skip all its methods.

If I'm not mistaken that may be enough to make `dyn` work in mono. Note that we must be careful to never translate a trait impl in mono mode (we should error if that happens); if we do it right, only the vtables and method impls will be translated.

This is not the easiest issue, come discuss in Zulip if you plan to tackle this.

PR #1042, opened 2026-03-01
Support `dyn Trait` with `--monomorphize`

Sam-Ni:
This is a draft PR implementing dynamic traits in mono mode. The implementation follows the proposal described in #856.

Currently, user-defined traits with generics and associated types are supported.

Some special cases still need to be handled to pass all test cases, including built-in traits (e.g., closures) and traits with default methods.

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_meta.rs
+++ b/charon/src/bin/charon-driver/translate/translate_meta.rs
@@ -79,7 +79,7 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
                         // We use the virtual name because it is always available.
                         // That name normally starts with `/rustc/<hash>/`. For our purposes we hide
                         // the hash.
-                        let virtual_name = name.path(RemapPathScopeComponents::empty());
+                        let virtual_name = name.path(RemapPathScopeComponents::MACRO);
                         let mut components_iter = virtual_name.components();
                         if let Some(
                             [
@@ -641,6 +641,8 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
                 }
                 Attribute::Unknown(raw_attr.clone())
             }
+            // `#[verify::test]`: mark a function for test extraction
+            "test" if args.is_none() => Attribute::Unknown(raw_attr.clone()),
             _ => return Ok(None),
         };
         Ok(Some(parsed))
--- a/charon/tests/ui/regressions/issue-1067-filename-ice.out
+++ b/charon/tests/ui/regressions/issue-1067-filename-ice.out
@@ -1,0 +1,169 @@
+# Final LLBC before serialization:
+
+pub struct core::fmt::Error {}
+
+// Full name: core::marker::MetaSized
+#[lang_item("meta_sized")]
+pub trait MetaSized<Self>
+
+// Full name: core::marker::Sized
+#[lang_item("sized")]
+pub trait Sized<Self>
+{
+    parent_clause0 : [@TraitClause0]: MetaSized<Self>
+    non-dyn-compatible
+}
+
+// Full name: core::result::Result
+#[lang_item("Result")]
+pub enum Result<T, E>
+where
+    [@TraitClause0]: Sized<T>,
+    [@TraitClause1]: Sized<E>,
+{
+  Ok(T),
+  Err(E),
+}
+
+// Full name: core::fmt::Debug
+#[lang_item("Debug")]
+pub trait Debug<Self>
+{
+    fn fmt<'_0_1, '_1_1, '_2_1> = core::fmt::Debug::fmt<'_0_1, '_1_1, '_2_1, Self>[Self]
+    vtable: core::fmt::Debug::{vtable}
+}
+
+pub fn core::fmt::Debug::fmt<'_0, '_1, '_2, Self>(_1: &'_0 Self, _2: &'_1 mut Formatter<'_2>) -> Result<(), core::fmt::Error>[{built_in impl Sized for ()}, {built_in impl Sized for core::fmt::Error}]
+where
+    [@TraitClause0]: Debug<Self>,
+= <opaque>
+
+// Full name: core::marker::Destruct
+#[lang_item("destruct")]
+pub trait Destruct<Self>
+{
+    fn drop_in_place = core::marker::Destruct::drop_in_place<Self>
+    non-dyn-compatible
+}
+
+unsafe fn core::marker::Destruct::drop_in_place<Self>(_1: *mut Self)
+= <opaque>
+
+// Full name: core::result::{Result<T, E>[@TraitClause0, @TraitClause1]}::unwrap
+pub fn unwrap<T, E>(_1: Result<T, E>[@TraitClause0, @TraitClause1]) -> T
+where
+    [@TraitClause0]: Sized<T>,
+    [@TraitClause1]: Sized<E>,
+    [@TraitClause2]: Debug<E>,
+= <opaque>
+
+pub opaque type std::io::error::Error
+
+// Full name: std::io::error::{impl Debug for std::io::error::Error}::fmt
+pub fn {impl Debug for std::io::error::Error}::fmt<'_0, '_1, '_2>(_1: &'_0 std::io::error::Error, _2: &'_1 mut Formatter<'_2>) -> Result<(), core::fmt::Error>[{built_in impl Sized for ()}, {built_in impl Sized for core::fmt::Error}]
+= <opaque>
+
+// Full name: std::io::error::{impl Debug for std::io::error::Error}
+impl Debug for std::io::error::Error {
+    fn fmt<'_0_1, '_1_1, '_2_1> = {impl Debug for std::io::error::Error}::fmt<'_0_1, '_1_1, '_2_1>
+    vtable: {impl Debug for std::io::error::Error}::{vtable}
+}
+
+// Full name: alloc::vec::Vec
+#[lang_item("Vec")]
+pub opaque type Vec<T>
+where
+    [@TraitClause0]: Sized<T>,
+    [@TraitClause1]: Sized<type_error("removed allocator parameter")>,
+
+// Full name: std::io::impls::{impl Write for Vec<u8>[{built_in impl Sized for u8}, @TraitClause0]}::write_all
+pub fn {impl Write for Vec<u8>[{built_in impl Sized for u8}, @TraitClause0]}::write_all<'_0, '_1, A>(_1: &'_0 mut Vec<u8>[{built_in impl Sized for u8}, @TraitClause0], _2: &'_1 [u8]) -> Result<(), std::io::error::Error>[{built_in impl Sized for ()}, {built_in impl Sized for std::io::error::Error}]
+where
+    [@TraitClause0]: Sized<A>,
+= <opaque>
+
+// Full name: alloc::alloc::Global
+#[lang_item("global_alloc_ty")]
+pub struct Global {}
+
+// Full name: alloc::vec::Vec::{impl Destruct for Vec<T>[@TraitClause0, @TraitClause1]}::drop_in_place
+unsafe fn {impl Destruct for Vec<T>[@TraitClause0, @TraitClause1]}::drop_in_place<T, A>(_1: *mut Vec<T>[@TraitClause0, @TraitClause1])
+where
+    [@TraitClause0]: Sized<T>,
+    [@TraitClause1]: Sized<A>,
+= <opaque>
+
+// Full name: alloc::vec::Vec::{impl Destruct for Vec<T>[@TraitClause0, @TraitClause1]}
+impl<T, A> Destruct for Vec<T>[@TraitClause0, @TraitClause1]
+where
+    [@TraitClause0]: Sized<T>,
+    [@TraitClause1]: Sized<A>,
+{
+    fn drop_in_place = {impl Destruct for Vec<T>[@TraitClause0, @TraitClause1]}::drop_in_place<T, A>[@TraitClause0, @TraitClause1]
+    non-dyn-compatible
+}
+
+// Full name: alloc::vec::{Vec<T>[@TraitClause0, {built_in impl Sized for Global}]}::new
+#[lang_item("vec_new")]
+pub fn new<T>() -> Vec<T>[@TraitClause0, {built_in impl Sized for Global}]
+where
+    [@TraitClause0]: Sized<T>,
+= <opaque>
+
+fn UNIT_METADATA()
+{
+    let _0: (); // return
+
+    _0 = ()
+    return
+}
+
+const UNIT_METADATA: () = @Fun0()
+
+// Full name: test_crate::main
+fn main()
+{
+    let _0: (); // return
+    let buf_1: Vec<u8>[{built_in impl Sized for u8}, {built_in impl Sized for Global}]; // local
+    let _2: (); // anonymous local
+    let _3: Result<(), std::io::error::Error>[{built_in impl Sized for ()}, {built_in impl Sized for std::io::error::Error}]; // anonymous local
+    let _4: &'1 mut Vec<u8>[{built_in impl Sized for u8}, {built_in impl Sized for Global}]; // anonymous local
+    let _5: &'3 [u8]; // anonymous local
+    let _6: &'5 [u8; 5 : usize]; // anonymous local
+    let _7: &'6 [u8; 5 : usize]; // anonymous local
+    let _8: [u8; 5 : usize]; // anonymous local
+    let _9: &'7 [u8; 5 : usize]; // anonymous local
+
+    _0 = ()
+    storage_live(buf_1)
+    buf_1 = new<u8>[{built_in impl Sized for u8}]()
+    storage_live(_2)
+    storage_live(_3)
+    storage_live(_4)
+    _4 = &two-phase-mut buf_1
+    storage_live(_5)
+    storage_live(_6)
+    storage_live(_7)
+    storage_live(_8)
+    _8 = [const 104 : u8, const 101 : u8, const 108 : u8, const 108 : u8, const 111 : u8]
+    storage_live(_9)
+    _9 = &_8
+    _7 = move _9
+    _6 = &(*_7)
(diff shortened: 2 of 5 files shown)


charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -333,6 +379,20 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
                 .retain(|(name, _)| method_is_used(trait_id, *name));
         }
     }
+
+    pub fn monomorphize_mode(&self) -> bool {
+        self.options.monomorphize_with_hax
+    }
+
```
Nadrieril: A note about this: throughout Charon I was careful to make monomorphization a per-item choice. You should be able to use `ItemTransCtx::monomorphize()` to know whether the current item is to be translated monomorphically or not, instead of looking at the global setting.

charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -147,19 +149,34 @@ impl<'tcx, 'ctx> TranslateCtx<'tcx> {
...
+                if mono {
+                    if matches!(kind, TraitImplSource::Normal) {
+                        bt_ctx.translate_trait_impl_mono(item_meta, &def)?;
+                    } else {
+                        error!("TraitImpl other than Normal is not supported in Mono");
                     }
-                };
-                self.translated.trait_impls.set_slot(id, trait_impl);
+                    // We do not translate trait impl blocks.
+                    self.translated.trait_impls.remove_and_shift_ids(id);
```
Nadrieril: That's not how I want to do this. If we do things right, there should simply be no monomorphic references to trait impls (except maybe for some builtin traits). Combined with the per-item monomrphization state I mentioned, this should be enough to avoid any attempts at translating a trait impl monomorphically. So we should keep them around, if only to get precise names for methods.

charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -631,6 +637,25 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
         self.translate_item(span, item, TransItemSourceKind::TraitDecl)
     }
 
+    // Used for translate_predicate with dynamic trait object in Mono mode.
+    // In Mono mode, we keep trait decl as in Poly mode, with generic and associative parameters.
+    pub(crate) fn translate_trait_decl_ref_poly(
```
Nadrieril: I would strongly prefer to avoid the need for this split. Why did you need a separate method?

Nadrieril: Basically, Charon is carefully architected so that we only branch between mono/poly at translation leaves (specifically, when calling into hax) as much as possible. If you have need to branch somewhere else then I'm hoping we can change something to avoid that, because that would makes everything harder to maintain

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -891,6 +1415,56 @@ impl ItemTransCtx<'_, '_> {
         })
     }
 
+    pub(crate) fn translate_vtable_instance_init_mono(
```
Nadrieril: I would like to avoid this duplication between `method` and `method_mono` as much as possible. Can you not reuse most of the method and change the behavior only where they differ?

charon/src/bin/charon-driver/translate/translate_items.rs
Nadrieril: Also shifting ids should never be done (that's why the method has such a long name): it will make any subsequent impls be in the wrong place.

charon/src/bin/charon-driver/translate/translate_crate.rs
Sam-Ni: > I would strongly prefer to avoid the need for this split. Why did you need a separate method?

Sorry for the unnecessary split. I realized that `translate_trait_decl_ref_poly` can actually be merged back into the original `translate_trait_decl_ref` by replacing the body of the latter with that of the former, without affecting the generated LLBC code in poly mode.

The key idea is that, whenever a trait declaration reference is needed in mono mode, we deliberately construct a poly item for translation. This avoids registering duplicated trait declarations instantiated with different generic arguments in `TranslateCtx::id_map`.

Sam-Ni: > Basically, Charon is carefully architected so that we only branch between mono/poly at translation leaves (specifically, when calling into hax) as much as possible. If you have need to branch somewhere else then I'm hoping we can change something to avoid that, because that would makes everything harder to maintain

Combined with the note above about `ItemTransCtx::monomorphize()`, does this architecture suggest that we should construct only Mono items for translation?

In my current implementation, however, `trait decls` and `struct vtables` are constructed as poly items for translation. The intuition is that there is only one opaque vtable struct per trait. As a result, we need global information to determine whether they should ultimately be translated monomorphically.

charon/src/bin/charon-driver/translate/translate_items.rs
Sam-Ni: I will find out how to avoid these two hacks.

charon/src/bin/charon-driver/translate/translate_trait_objects.rs
Sam-Ni: Yes, most of the methods have large portion of codes in common. I'll combine these methods.

charon/src/bin/charon-driver/translate/translate_crate.rs
Nadrieril: Hm that's tricky. If we use `Poly` then `self.monomorphize()` will return `false` indeed. I propose to introduce a new variant `RustcItem::MonoTrait(DefId)` which would work mostly like `Poly` except that it knows it's being translated monomorphically.

Nadrieril: It's possible we still want `self.monomorphize()` to return `false` for this `RustcItem` since any other items mentioned in the trait should be translated polymorphically. Though I don't think there should be any such items?

Sam-Ni: I think introducing the new variant `RustcItem::MonoTrait(DefId)` makes sense. I’ll go ahead and add it to the code.

As for the trait declaration, I don’t believe there are any items that need to be translated polymorphically. The most complicated case I can think of is a generic trait with default methods. However, even in this case, the methods still need to be instantiated for each trait implementation, rather than being translated polymorphically.

Nadrieril: Well, if we e.g. translated the method signatures as part of the trait declaration then they could refer to other types like `Option` and we'd need to translate that polymorphically. And we indirectly do that because we translate `FunDeclRef`s to method declarations.

For this initial implementation we should remove the list of methods from the translated trait declarations, exactly to avoid this issue.

charon/src/bin/charon-driver/translate/translate_items.rs
Nadrieril: What issue arises if you just remove this line?

charon/src/bin/charon-driver/translate/translate_crate.rs
Sam-Ni: Got it.

charon/src/bin/charon-driver/translate/translate_items.rs
Sam-Ni: No issue in Charon, actually. The only functionality of trait impls is to offer names for methods. Maybe we can keep the trait impls so that Eurydice will pass the translation without the modification disscussed on Zulip?

Nadrieril: Yeah, that might very well work!

Sam-Ni: It seems that Eurydice passes the check for trait impl references! Now the error of Eurydice reduces to:
```
ERROR translating test_mono::{test_mono::Trait<i32>}::{vtable}: Failure("expression_of_operand: Opaque(missing supertrait vtable)")
```
I think this is because when initializing the vtable, the field of super trait is filled with `Opaque(missing supertrait vtable)`? Am I missing some compilation options to generate the supertrait vtables?

Nadrieril: You can grep for this message inside charon, this is a missing feature. For `dyn` of builtin traits it seems we don't yet go fetch the right charon-generated impl

Nadrieril: If you have a reproducer, please open an issue in charon

Sam-Ni: I see. I’ve opened issue #1047 to track this.

Sam-Ni:
Based on the above discussion, I have improved the code. Specifically, modifications in this commit:

1. Removed the global `monomorphize_mode`.

2. Trait implementations are now kept in mono mode. Only their associated types and constants are translated.

3. Introduced a new variant `RustcItem::MonoTrait`. Both `struct vtable` and `trait decl` items are now registered as `MonoTrait`.

4. Removed all `method_mono` functions by merging them into the corresponding `method` functions.

charon/src/ast/gast.rs
```diff
@@ -167,11 +169,15 @@ pub enum ItemSource {
...
     /// This is a vtable value for an impl.
-    VTableInstance { impl_ref: TraitImplRef },
+    VTableInstance {
+        impl_ref: TraitImplRef,
+    },
     /// The method shim wraps a concrete implementation of a method into a function that takes `dyn
     /// Trait` as its `Self` type. This shim casts the receiver to the known concrete type and
     /// calls the real method.
     VTableMethodShim,
+    VTableInstanceMono,
```
Nadrieril, line 179: What's the reason you can't use `VTableInstance`?

charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -41,6 +41,7 @@ pub struct TransItemSource {
 pub enum RustcItem {
     Poly(hax::DefId),
     Mono(hax::ItemRef),
+    MonoTrait(hax::DefId),
```
Nadrieril, line 49: Could you document the meaning of this variant please?

charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -126,6 +136,13 @@ impl TransItemSource {
         Self::new(RustcItem::Mono(item.clone()), kind)
     }
 
+    /// Refers to the monomorphic trait (or vtable)
+    /// that works mostly like the polymorphic version
+    /// except that it knows it's being translated monomorphically.
```
Nadrieril: Could you add a note like "see the docs of `RustcItem::MonoTrait` for details", and on `MonoTrait` I'd like some details like "the trait is translated like in poly mode except that we don't translate any of its associated item lists and erase the signature of methods in its vtable".

charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -628,7 +654,24 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
-        self.translate_item(span, item, TransItemSourceKind::TraitDecl)
+        let item_src = if self.polymorphize() {
+            trace!(
+                "MONO: catch polymorphic item {:?} in translate_trait_decl_ref",
+                self.item_src
+            );
+            TransItemSource::polymorphic(&item.def_id, TransItemSourceKind::TraitDecl)
+        } else {
+            TransItemSource::monomorphic_trait(&item.def_id, TransItemSourceKind::TraitDecl)
+        };
```
Nadrieril: Is there a reason to do this check here instead of inside `register_item_maybe_enqueue`? You can check that the target `def_id` is a trait by looking at `item.def_id.kind`

charon/src/bin/charon-driver/translate/translate_ctx.rs
```diff
@@ -223,6 +228,16 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+    /// Whether to polymorphize items we encounter.
+    pub fn polymorphize(&self) -> bool {
+        matches!(self.item_src.item, RustcItem::Poly(..))
+    }
+
+    /// Whether to pay special attention to items in mono mode
+    pub fn monomorphize_trait(&self) -> bool {
+        matches!(self.item_src.item, RustcItem::MonoTrait(..))
+    }
+
```
Nadrieril: Instead of these three methods, does it work if you change `monomorphize` to return `true` for both `Mono` and `MonoTrait`, and only use that method or its negation?

charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -715,6 +732,13 @@ impl ItemTransCtx<'_, '_> {
                 hax::AssocKind::Const { .. } => TransItemSourceKind::Global,
                 hax::AssocKind::Type { .. } => TransItemSourceKind::Type,
             };
+
+            // skip all methods of trait decl in mono mode
+            // question: what if the method has default implmentation?
```
Nadrieril: Good point: this will break defaulted methods, which currently rely on finding the method inside the trait decl. I'm ok with that, that will get fixed once we compute method signatures properly.

charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -1024,85 +1052,97 @@ impl ItemTransCtx<'_, '_> {
...
-                    let bound_fn_ref = match &impl_item.value {
-                        Provided { .. } => self.translate_binder_for_def(
-                            item_span,
-                            binder_kind,
-                            &item_def,
-                            |ctx| {
-                                let generics = ctx
+                    // In mono mode, we skip all methods in trait decl.
+                    // Correspondingly, here we skip all methods in trait impl.
+                    if self.polymorphize() {
```
Nadrieril: Nitpick but could you make this
```rust
if self.monomorphize() {
  continue;
}
```
to avoid the indentation

charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -1024,85 +1052,97 @@ impl ItemTransCtx<'_, '_> {
...
+                                if !self.monomorphize() {
+                                    generics.trait_refs.push(self_predicate.clone());
+                                }
+                                generics
                             }
-                            generics
                         }
+                    } else {
+                        // In mono impls, we keep the instantiated generics from the const item itself.
+                        self.the_only_binder().params.identity_args()
```
Nadrieril: I believe that's incorrect: in mono mode we have translated the global in mono mode too, hence it takes no generics. I think this code was correct before

charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -1024,85 +1052,97 @@ impl ItemTransCtx<'_, '_> {
...
                     let gref = GlobalDeclRef {
                         id,
                         generics: Box::new(generics),
                     };
                     consts.push((name, gref));
                 }
-                // Monomorphic traits have no associated types.
-                hax::FullDefKind::AssocTy { .. } if self.monomorphize() => continue,
+                // In mono mode, trait decl keeps associative parameters.
+                // Correspondingly, here we translate associative arguments for trait impl (the same way as in poly mode).
```
Nadrieril: If I may:
```suggestion
                // In mono mode, trait decls keep their associated types.
                // Correspondingly, here we translate the associated types like in poly mode.
```

charon/src/bin/charon-driver/translate/translate_items.rs
```diff
@@ -1136,12 +1176,15 @@ impl ItemTransCtx<'_, '_> {
...
         Ok(TraitImpl {
             def_id,
             item_meta,
             impl_trait: implemented_trait,
-            generics: self.into_generics(),
+            generics: if self.polymorphize() {
+                self.into_generics()
+            } else {
+                Default::default()
+            },
```
Nadrieril: This should be already done correctly by the rest of the machinery

charon/src/ast/gast.rs
Sam-Ni: Because the `VTableInstance` variant receives a `TraitImplRef`. In mono mode, I think we should use trait impl refs as little as possible, as the proposal suggests. (Although in the current implementation we keep trait impls to pass Eurydice’s checks)

charon/src/bin/charon-driver/translate/translate_crate.rs
Sam-Ni: Sure

charon/src/bin/charon-driver/translate/translate_crate.rs
Sam-Ni: Sure. (same as above)

charon/src/bin/charon-driver/translate/translate_crate.rs
Sam-Ni: Thanks for pointing this out! I have pushed the check down to `register_item_maybe_enqueue` in my local branch.

charon/src/bin/charon-driver/translate/translate_ctx.rs
Sam-Ni: Yes. I have tested it in my local branch and it works.

charon/src/bin/charon-driver/translate/translate_items.rs
Sam-Ni: Sure

charon/src/bin/charon-driver/translate/translate_items.rs
Sam-Ni: Yes, you're right. I misunderstood the process of translating assoc consts.

charon/src/bin/charon-driver/translate/translate_items.rs
Sam-Ni: Thank you for pointing this out!

charon/src/bin/charon-driver/translate/translate_ctx.rs
Nadrieril: Nice

PR #1054, opened 2026-03-12
Publish pre-built artifacts to GitHub

### LINKED ISSUE #1053: Feature request: Publish pre-builts on GitHub
I'd like to add a Hermes feature (https://github.com/google/zerocopy/issues/3106) which will automatically install dependencies from pre-built binaries so that users don't have to have all of Charon's build dependencies installed. I'm open to hosting these on the zerocopy repo, but I wanted to see if y'all would be willing to publish and host in this repo instead. I'm happy to do it, but if *someone* is going to do it, it makes more sense for them to live here so that *all* Charon users can benefit, not just Hermes users.

See also: https://github.com/AeneasVerif/aeneas/issues/834

joshlf:
Closes #1053

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,97 @@
+name: Release
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build (${{ matrix.os }} - ${{ matrix.target }})
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+            artifact_name: charon-linux-x86_64
+          - os: macos-latest
+            target: x86_64-apple-darwin
+            artifact_name: charon-macos-x86_64
+          - os: macos-latest
+            target: aarch64-apple-darwin
+            artifact_name: charon-macos-aarch64
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      # Extract the pinned toolchain from the source of truth.
+      - name: Read toolchain
+        id: toolchain
+        run: |
+          CHANNEL=$(grep 'channel =' rust-toolchain | cut -d '"' -f 2)
+          echo "channel=$CHANNEL" >> "$GITHUB_OUTPUT"
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ steps.toolchain.outputs.channel }}
+          targets: ${{ matrix.target }}
+          components: rustc-dev
+
+      - name: Install musl-tools
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get update && sudo apt-get install -y musl-tools
+
+      - name: Build
+        run: |
+          cd charon
+          cargo build --release --target ${{ matrix.target }}
+        env:
+          # Ensure static linking for OpenSSL to maximize binary portability.
+          OPENSSL_STATIC: 1
+
+      - name: Package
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly-2026-02-07
```
Nadrieril: Is there a way to use the pinned toolchain directly instead of duplicating the version here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
+name: Release
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
```
Nadrieril, line 8: Honestly let's do it on every commit, I don't want to have to manage tags until we have a proper semver + release setup.

Nadrieril: Oh, does this have to be tied to a GitHub "Release"? Can we have Releases on commits instead of tags?

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+  release:
+    name: Create Release
+    needs: build
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
+    permissions:
+      contents: write
+    steps:
+      - name: Download Artifacts
+        uses: actions/download-artifact@v4
```
Nadrieril: Isn't it silly to download artifacts that we just uploaded? Can this not be merged with the `build` job?

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.artifact_name }}
+          path: ${{ matrix.artifact_name }}.tar.gz
+
+  release:
+    name: Create Release
+    needs: build
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
```
Nadrieril: That's redundant with the on-push filter no?

.github/workflows/release.yml
joshlf: Done

.github/workflows/release.yml
joshlf: How does this solution look?

.github/workflows/release.yml
joshlf: Removed

.github/workflows/release.yml
joshlf: Good catch; fixed

.github/workflows/release.yml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          set -eo pipefail
+          
+          mkdir staging
+          cp charon/target/${{ matrix.target }}/release/charon staging/
+          cp charon/target/${{ matrix.target }}/release/charon-driver staging/
+          cd staging
+          tar -czvf ../${{ matrix.artifact_name }}.tar.gz *
+
+      # Verify the binary is functional and correctly statically linked on Linux.
```
Nadrieril, line 70: Are you aware that `charon-driver` is dynamically linked against the rustc toolchain? May that be a problem for your usecase? It will require rustup to be installed on the target machine.

.github/workflows/release.yml
Nadrieril: Very nice :+1:

.github/workflows/release.yml
joshlf: Yeah, probably should've mentioned that caveat in this comment.

joshlf:
Okay hopefully this fixes it: https://github.com/AeneasVerif/charon/pull/1058

PR #1058, opened 2026-03-13
In pre-builts, use dynamic gnu, not static musl

joshlf:


joshlf:
@Nadrieril can you re-run https://github.com/AeneasVerif/charon/actions/runs/23034198204/job/66898503844 and https://github.com/AeneasVerif/charon/actions/runs/23034198204/job/66898503865? They were canceled because https://github.com/AeneasVerif/charon/actions/runs/23034198204/job/66898503850 failed. I'd like to see what happens when they run to completion.

PR #1059, opened 2026-03-13
Run release Action on native runners

joshlf:
Build Mac targets on Mac runners

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,15 +15,14 @@ jobs:
       contents: write
     strategy:
       matrix:
+        # NOTE: We use native runners instead of cross-compiling to work around
+        # https://github.com/rust-lang/rust/issues/132728#issuecomment-2462937687
         include:
           - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -35,38 +32,46 @@ jobs:
...
         env:
           # Ensure static linking for OpenSSL to maximize binary portability.
           OPENSSL_STATIC: 1
+        run: |
+          set -eo pipefail
+
+          cd charon
+          # We don't specify `--target` because `charon-driver` uses `rustc_private`,
+          # which is only compatible with the host architecture of the runner.
+          # We use specific runners for each target instead.
```
Nadrieril: That's a surprising comment to me, I would not have expected we'd use `--target`. Maybe just add a comment like "we can't cross-compile, see https://github.com/rust-lang/rust/issues/132728#issuecomment-2462937687 " near the list of targets?

joshlf: Done

PR #1066, opened 2026-03-16
Fix an issue in the control-flow reconstruction

### LINKED ISSUE #1051: Bug: incorrect control-flow reconstruction in the presence of loops
**Code snippet to reproduce the bug**:

```rust
fn f(mut y: u32) -> bool {
    if y > 0 {
        loop {
            if y == 0 {
                return true;
            }
            y -= 1;
            if y == 0 {
                break;
            }
        }
    }
    false
}
```

Resulting LLBC:
```rust
fn temp::f(y^1 : u32) -> bool
{
  v@0 : bool;
  y^1 : u32;
  v@2 : bool;
  v@3 : u32;
  v@4 : bool;
  v@5 : u32;
  v@6 : u32;
  v@7 : bool;
  v@8 : u32;

  storage_live v@6;
  storage_live v@2;
  storage_live v@3;
  v@3 := copy y^1;
  v@2 := move v@3 > 0u32;
  if (move v@2) {
    storage_dead v@3;
    loop {
      storage_live v@4;
      storage_live v@5;
      v@5 := copy y^1;
      v@4 := move v@5 == 0u32;
      if (move v@4) {
        // The return got replaced by a break
        break 0
      }
      else {
        storage_dead v@5;
        storage_dead v@4;
        v@6 := copy y^1 panic.- 1u32;
        y^1 := move v@6;
        storage_live v@7;
        storage_live v@8;
        v@8 := copy y^1;
        v@7 := move v@8 == 0u32;
        if (move v@7) {
          storage_dead v@8;
          storage_dead v@7
          // HERE: there was initially a break.
          // Generally speaking, we should never reach the end of a loop body without going through
          // a continue or break. Moreover, whatever the choice we make (insert a continue or a break)
          // the resulting LLBC is erroneous.
        }
        else {
          storage_dead v@8;
          storage_dead v@7;
          continue 0
        }
      }
    };
    storage_dead v@5;
    v@0 := true;
    storage_dead v@4;
    storage_dead v@2;
    return
  }
  else {
    storage_dead v@3
  };
  storage_dead v@2;
  v@0 := false;
  return
}
```

sonmarcho:
Fixes https://github.com/AeneasVerif/charon/issues/1051

This PR is AI generated *but*: I carefully reviewed the code and prompted the AI until I was globally happy with it, before doing the last modifications by hand. I reviewed all the code line by line.

This PR does two things:
1. improve the heuristics in `compute_loop_exits` to find better loop exits in the presence of branchings
2. fix `translate_jump` by detecting the case problematic in #issue1051 and failing

About 2.: I tried fixing it by inlining the target of the jump instead of introducing a no-op. I think a solution along those lines would make sense, but just inlining probably doesn't work: for instance, I'm not completely sure what happens in case this inlined code ends with, e.g., a break or a continue. As the initial attempt made sense I committed it to keep track of it, but then reverted it. The version I submit in this PR simply fails in the problematic case.

### DIFF
--- a/charon/src/transform/control_flow/ullbc_to_llbc.rs
+++ b/charon/src/transform/control_flow/ullbc_to_llbc.rs
@@ -639,40 +639,15 @@ impl ExitInfo {
             //     if b {
             //         loop {
             //             if b { return true; }  // bb5: dominated by loop header
-            //             break;                  // bb9→bb4: exits the loop
+            //             break;                 // bb9→bb4: exits the loop
             //         }
             //     }
-            //     false                           // bb4: NOT dominated by loop header
+            //     false                          // bb4: NOT dominated by loop header
             // }
             // ```
             //
-            // Without the filter, bb5 (`return true`) ranks higher than bb4 because it is
-            // closer to the loop header. Picking bb5 as the exit incorrectly turns `return
-            // true` into `break`, producing the following wrong reconstruction:
-            // ```
-            // fn f(b: bool) -> bool {
-            //     if b {
-            //         loop {
-            //             if b { break; }   // was `return true` — wrongly turned into break
-            //             // the actual `break` path is lost here
-            //         }
-            //     }
-            //     false
-            // }
-            // ```
-            // The filter discards bb5 (dominated) so that bb4 (not dominated) is correctly
-            // chosen, producing the correct reconstruction:
-            // ```
-            // fn f(b: bool) -> bool {
-            //     if b {
-            //         loop {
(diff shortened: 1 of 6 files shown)


Nadrieril:
I made https://github.com/AeneasVerif/charon/pull/1306 which does inline instead of incorrectly falling through. This should always be correct thanks to the invariants of the reconstruction algorithm. I'm still curious to get heuristic improvements; I let Codex try a few things and none were good enough to keep.

Nadrieril:
The original issue got fixed, I think I can close this

PR #1087, opened 2026-04-06
Keep generic types when translating trait alias in monomorphized mode

### LINKED ISSUE #1086: Bug: `mismatched type generics` appears when translating trait alias in monomorphized mode
This error occurs when [translating the Rust standard library](https://github.com/AeneasVerif/charon/issues/863) with Charon running in monomorphized mode. Below we first describe how the error arises when translating `std`, and then provide a minimal example (independent of `std`) to reproduce it.

**Reproducing the error in `std`**

The error can be triggered when translating `std::env::*` with the following commands:

```
cd charon
echo "fn main() {}" > file.rs
cargo build
./target/debug/charon rustc --monomorphize --start-from="std::env::*" --include=std --no-serialize -- --sysroot=$(rustc --print sysroot) file.rs
```

Part of the error output (there are other errors, which are omitted here) is:

```
error: Type error after translation:
       Mismatched type generics:
       target: Thin
       expected: [Self]
            got: []
       Visitor stack:
         charon_lib::ast::types::TraitDeclRef
         charon_lib::ast::gast::TraitImpl
       Binding stack (depth 1):
         0: <[@TraitClause0]: @TraitDecl41<()>, {built_in impl @TraitDecl41 for ()}::Metadata = ()>
 --> /rustc/library/core/src/ptr/metadata.rs:85:1

error: Type error after translation:
       Mismatched type generics:
       target: {impl Thin}::<()>
       expected: []
            got: [()]
       Visitor stack:
         charon_lib::ast::types::TraitImplRef
         charon_lib::ast::types::TraitRefKind
         charon_lib::ast::types::TraitRefContents
         charon_lib::ast::hash_cons::HashConsed<charon_lib::ast::types::TraitRefContents>
         charon_lib::ast::types::TraitRef
         charon_lib::ids::index_map::IndexMap<charon_lib::ast::types::vars::TraitClauseId, charon_lib::ast::types::TraitRef>
         charon_lib::ast::types::GenericArgs
         charon_lib::ast::types::Binder<charon_lib::ast::types::GenericArgs>
         alloc::boxed::Box<charon_lib::ast::types::Binder<charon_lib::ast::types::GenericArgs>>
         charon_lib::ast::names::PathElem
         alloc::vec::Vec<charon_lib::ast::names::PathElem>
         charon_lib::ast::names::Name
         charon_lib::ast::meta::ItemMeta
         charon_lib::ast::gast::FunDecl
       Binding stack (depth 2):
         0:
         1:
 --> /rustc/library/core/src/ptr/mod.rs:862:1
```

After inspecting `/rustc/library/core/src/ptr/metadata.rs:85:1` and `/rustc/library/core/src/ptr/mod.rs:862:1` (using `rustc` version `nightly-2026-02-07`, which matches the version used to compile Charon), @maltuned and I found that the error is caused by the use of trait alias).

**Minimal reproduction**

We constructed the following minimal example to reproduce the error:

```rust
#![feature(trait_alias)]

trait A {}

trait B = A;

impl A for i32 {

}

fn f<T : B>() {

}

fn main() {
    f::<i32>();
}
```

Running:

```
charon cargo --monomorphize --no-serialize
```

produces:

```
error: Type error after translation:
       Mismatched type generics:
       target: B
       expected: [Self]
            got: []
       Visitor stack:
         charon_lib::ast::types::TraitDeclRef
         charon_lib::ast::gast::TraitImpl
       Binding stack (depth 1):
         0: <[@TraitClause0]: @TraitDecl1<i32>, [@TraitClause1]: @TraitDecl0<i32>>
 --> src/main.rs:5:1
  |
5 | trait B = A;
  | ^^^^^^^^^^^^

error: Type error after translation:
       Mismatched type generics:
       target: {impl B}::<i32>
       expected: []
            got: [i32]
       Visitor stack:
         charon_lib::ast::types::TraitImplRef
         charon_lib::ast::types::TraitRefKind
         charon_lib::ast::types::TraitRefContents
         charon_lib::ast::hash_cons::HashConsed<charon_lib::ast::types::TraitRefContents>
         charon_lib::ast::types::TraitRef
         charon_lib::ids::index_map::IndexMap<charon_lib::ast::types::vars::TraitClauseId, charon_lib::ast::types::TraitRef>
         charon_lib::ast::types::GenericArgs
         charon_lib::ast::types::Binder<charon_lib::ast::types::GenericArgs>
         alloc::boxed::Box<charon_lib::ast::types::Binder<charon_lib::ast::types::GenericArgs>>
         charon_lib::ast::names::PathElem
         alloc::vec::Vec<charon_lib::ast::names::PathElem>
         charon_lib::ast::names::Name
         charon_lib::ast::meta::ItemMeta
         charon_lib::ast::gast::FunDecl
       Binding stack (depth 2):
         0:
         1:
  --> src/main.rs:11:1
   |
11 | / fn f<T : B>() {
12 | |
13 | | }
   | |_^

error: Type error after transformations:
       Mismatched type generics:
       target: B
       expected: [Self]
            got: []
       Visitor stack:
         charon_lib::ast::types::TraitDeclRef
         charon_lib::ast::gast::TraitImpl
       Binding stack (depth 1):
         0: <[@TraitClause0]: @TraitDecl1<i32>, [@TraitClause1]: @TraitDecl0<i32>>
 --> src/main.rs:5:1
  |
5 | trait B = A;
  | ^^^^^^^^^^^^

error: Type error after transformations:
       Mismatched type generics:
       target: {impl B}::<i32>
       expected: []
            got: [i32]
       Visitor stack:
         charon_lib::ast::types::TraitImplRef
         charon_lib::ast::types::TraitRefKind
         charon_lib::ast::types::TraitRefContents
         charon_lib::ast::hash_cons::HashConsed<charon_lib::ast::types::TraitRefContents>
         charon_lib::ast::types::TraitRef
         charon_lib::ids::index_map::IndexMap<charon_lib::ast::types::vars::TraitClauseId, charon_lib::ast::types::TraitRef>
         charon_lib::ast::types::GenericArgs
         charon_lib::ast::types::Binder<charon_lib::ast::types::GenericArgs>
         alloc::boxed::Box<charon_lib::ast::types::Binder<charon_lib::ast::types::GenericArgs>>
         charon_lib::ast::names::PathElem
         alloc::vec::Vec<charon_lib::ast::names::PathElem>
         charon_lib::ast::names::Name
         charon_lib::ast::meta::ItemMeta
         charon_lib::ast::gast::FunDecl
       Binding stack (depth 2):
         0:
         1:
  --> src/main.rs:11:1
   |
11 | / fn f<T : B>() {
12 | |
13 | | }
   | |_^
```

**Charon version**: the latest Charon in monomorphized mode (commit 77d520657e76f265f21a0516c2e4d8d49ba27056)

**Explain the bug**: (why is this behavior incorrect, and what do you think Charon should do instead)

The trait aliases are not correctly translated in monomorphized mode because [the PR about supporting dynamic trait object in mono mode](https://github.com/AeneasVerif/charon/pull/1042) only deals with the "normal" (contrary to trait alias) traits by keeping the generic and associative types. This makes the translation of trait aliases inconsistent with the nornal traits, which caused the `mismatched type generics` error.

To solve this problem, we think the generic types of trait aliases should also be preserved in mono mode, which is consistent with the treatment of normal traits.

Sam-Ni:
This PR closes #1086 . We keep the generic types (such as `Self`) of trait aliases in their declarations and impl blocks to make it consistent with the "normal" trait declarations in mono mode.


Nadrieril:
Hm that feels wrong, the way trait aliases are translated is that they're translated twice: once as `TraitDecl` and once as a `TraitImpl`. The trait impl side should be translated as a normal `RustcItem::Mono`, can you justify why the special mono-trait exception is useful here?

Sam-Ni:
> Hm that feels wrong, the way trait aliases are translated is that they're translated twice: once as `TraitDecl` and once as a `TraitImpl`. The trait impl side should be translated as a normal `RustcItem::Mono`, can you justify why the special mono-trait exception is useful here?

I see. I misunderstood how impl blocks work for trait aliases. So we should keep a single trait declaration and provide separate impl blocks for different types, just like the normal traits. Is that correct?

Nadrieril:
Yep, that's what I had in mind so that trait aliases can be as close to normal traits as possible

charon/src/bin/charon-driver/translate/translate_meta.rs
```diff
@@ -440,8 +440,15 @@ impl<'tcx> TranslateCtx<'tcx> {
     pub fn translate_name(&mut self, src: &TransItemSource) -> Result<Name, Error> {
         let mut name = self.name_for_src(src)?;
         // Push the generics used for monomorphization, if any.
+        // Exclude trait impl of trait alias,
+        //    For exmaple., we use the name `{impl B for i32}` rather than `{impl B for i32}::i32`,
+        //    where `B` is a trait alias and `i32` is the generic argument for `Self`.
```
Nadrieril: I'm confused by that remark: I don't see how we can ever get `{impl B for i32}::i32`. We can get `{impl B for i32}` or we can get `{impl<T> B for T}::i32`. To me `{impl B for i32}::i32` feels like something got messed up in the generics handling, no?

Sam-Ni: Yes, you’re right. The remark is misleading. What I meant to say is that, to avoid generating names like `{impl B for i32}::i32` for the trait impl of a trait alias, we need to check `src.kind` in the `if` statement below the remark.

Is it better not to include such incorrect examples in the remark (since they can cause confusion)?

Nadrieril: It is obviously better not to include incorrect examples, but what I'd really want is a correct example and explanation

Sam-Ni: Got it. I'll do that.

Sam-Ni:
The remark is updated to include clearer explanations and examples on how trait implementations for aliases are translated in mono mode. If anything is still unclear or confusing, please let me know.

charon/src/bin/charon-driver/translate/translate_meta.rs
```diff
@@ -441,8 +441,32 @@ impl<'tcx> TranslateCtx<'tcx> {
     pub fn translate_name(&mut self, src: &TransItemSource) -> Result<Name, Error> {
         let mut name = self.name_for_src(src)?;
         // Push the generics used for monomorphization, if any.
+        // As an exception, we do not push generics for impls of trait aliases,
+        // since this information is already encoded in the impl block itself,
+        // just like impls of normal traits.
```
Nadrieril, line 446: I still don't understand why this ever comes up, and you've not explained why there would ever be generics applied to an item which is monomorphic. But I'll accept this fix since it only applies to names

Nadrieril: I just expect there's a deeper bug with trait aliases + mono that will show up somewhere else, so let's keep that in mind

Sam-Ni: > I still don't understand why this ever comes up, and you've not explained why there would ever be generics applied to an item which is monomorphic. But I'll accept this fix since it only applies to names

This comes up from a difference between how normal trait impls and trait alias impls are represented in Hax. In Mono mode, the `generic_args` stored in the Hax `item_ref` for normal trait impls is empty. In contrast, for trait alias impls, the `generic_args` field still contains type arguments (including those for `Self` and other generic parameters).

At the same time, other monomorphic items (e.g., functions instantiated from generics) require an `Instantiated` suffix (e.g., `f::<i32>`) to record the concrete instantiation in their names. However, for impl blocks, the generic arguments are already encoded directly in the impl itself (e.g., `impl A<i32> for u32`). Therefore, I treat trait alias impls specially and avoid pushing generics for them.

Sam-Ni: > I just expect there's a deeper bug with trait aliases + mono that will show up somewhere else, so let's keep that in mind

I did encounter another issue related to trait aliases when translating the Rust standard library in Mono mode. I have not yet checked it in detail. I'll do that next.

Nadrieril: > In Mono mode, the generic_args stored in the Hax item_ref for normal trait impls is empty.

That doesn't seem right, an `ItemRef` always should have exactly the number of arguments that the item's `ParamEnv` expects

Nadrieril: I understand now: your code is right, and in fact should apply to all impls and not just trait alias impls. See the last commit of https://github.com/AeneasVerif/charon/pull/1123.

Sam-Ni: I see. The current code that deals with names is much more strucutured and general. Thank you!

Issue #864, opened 2025-10-22
Feature request: make a big enum for intrinsics


Nadrieril:
Currently rust intrinsics are translated as normal functions with an opaque body. Intrinsics are different beasts however: they describe built-in operations of the Rust language. From a discussion with @N1ark, I'd like for `FunDecl.body` to be a new enum:
```rust
enum FunBody {
  /// A normal function body.
  Body(Body),
  /// A built-in operation in MIR.
  Intrinsic(Intrinsic),
  /// A body that was not translated due to `--include`/`--opaque` choices in the Charon invocation.
  Opaque,
}

#[non_exhaustive]
enum Intrinsic {
  /// The [saturating_add](https://doc.rust-lang.org/std/intrinsics/fn.saturating_add.html) intrinsic.
  SaturatingAdd,
  ...
}
```

The way to know if a function is an intrinsic is `tcx.intrinsic(def_id)` using the rustc def_id, and then it's identified by its name. That intrinsic name should eventually be added in `hax::FullDef`, alongside `lang_item`. There are [quite a bunch of intrinsics](https://doc.rust-lang.org/std/intrinsics/index.html) so we'll probably want a macro or something to make it easy to update the enum.

Come discuss on Zulip if you want to tackle this.

Nadrieril:
Thanks to https://github.com/AeneasVerif/charon/pull/1088 we're pretty much there. The only thing is that we identify intrinsics by name instead of having an enum. Is it really worth having an enum? That feels like maintenance pain for little benefit, now that I think about it.

N1ark:
yeah, i'm not convinced it's useful. on one hand it is trivial to run Charon with `--start-from core::intrinsics` and gather them, on the other I'm not sure this won't just add an extra step (matching intrinsic names to the enum) for no reason, since intrinsics are identified by anyways.
happy to close this, the current state is good enough for my purposes

Nadrieril:
Nice, let's

PR #1088, opened 2026-04-06
Add `Extern` and `Intrinsic` body types

### LINKED ISSUE #982: Feature request: extern function names
Currently, there is no way to tell apart a function that has no body because it's external from a function that has no body because it couldn't be translated. Furthermore, external function names inherit the name of the module they're defined in, e.g. here:

```rust
// file foo.rs
fn bar() {
    unsafe extern "C" {
        fn foo(); // full name: foo::bar::foo()
    }

    unsafe { foo() }
}

fn main() {
    unsafe extern "C" {
        fn foo(); // full name: foo::main::foo()
    }

    unsafe { foo() };
    bar();
}
```

This shouldn't be the case, I think -- extern functions are identified by name and all map to the "same" function from what i understand, i.e. both foo functions should be the same and have the name `foo`. Them being the same exactly is probably not too much of a big deal (at least for me), however the name problem is quite bothersome. 

In particular, `alloc::Global` uses [the extern functions `__rust_alloc`, `__rust_dealloc`, etc.](https://doc.rust-lang.org/src/alloc/alloc.rs.html#11-38) to do dynamic allocations, but if one wants to stub these functions in their Charon client it is quite annoying:
1. their name is e.g. `std::alloc::__rust_alloc`, making using name matching not very reliable (e.g. a user could just re-specify `__rust_alloc` in their own module, where the name pattern wouldn't match) 
2. there is no way to know these functions are `extern`, so if a user were to create a function that happens to be called `__rust_alloc`, it might get accidentally stubbed, despite having nothing to do with the magic `__rust_alloc` function.

Adding e.g. `is_extern: bool` to `FunDecl` would fix this :)

N1ark:
Add two variants to `Body`: `Extern(string)` and `Intrinsic(string)`! On the OCaml side, a body isn't just a `'body option` anymore: we instead properly translate the `Body` enum, to give more information to the client.

For now intrinsics are just strings rather than an enum, as I don't have the time to do that part. There is also the question of [how we make this enum make sense](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Intrinsics.20lowered.20to.20MIR/with/580679106): some intrinsics actually never get lowered into a function call, and a few of them disappear through the compilation pipeline. 

One thing we lose with this change is that we can't get the name of the intrinsic's arguments. This isn't a big deal but in Soteria I [generate intrinsic definitions](https://github.com/soteria-tools/soteria/blob/9f496145ceb6fa898244440e22447de60981036d/soteria-rust/lib/builtins/intrinsics_intf.ml#L735-L743) from the ULLBC output, and having names for the arguments is super useful. Maybe we could make a new `Argument` type that stores a type, optional name (and an ArgAbi?) to keep this information? And then `FunSig.inputs` would be a `Vec<Arg>` instead. Let me know.

Fixes #982 
First step towards #864

ci: use https://github.com/AeneasVerif/aeneas/pull/919
ci: use https://github.com/AeneasVerif/eurydice/pull/395

### DIFF
--- a/charon-ml/src/LlbcAstUtils.ml
+++ b/charon-ml/src/LlbcAstUtils.ml
@@ -15,8 +15,8 @@ let fun_decl_list_from_crate (crate : crate) : fun_decl list =
     returns None *)
 let get_fun_args (fun_decl : fun_decl) : local list option =
   match fun_decl.body with
-  | Some body -> Some (GAstUtils.locals_get_input_vars body.locals)
-  | None -> None
+  | Body body -> Some (GAstUtils.locals_get_input_vars body.locals)
+  | _ -> None
 
 (** Check if a {!type:Charon.LlbcAst.statement} contains loops *)
 let block_has_loops (blk : block) : bool =
@@ -34,8 +34,8 @@ let block_has_loops (blk : block) : bool =
 (** Check if a {!type:Charon.LlbcAst.fun_decl} contains loops *)
 let fun_decl_has_loops (fd : fun_decl) : bool =
   match fd.body with
-  | Some body -> block_has_loops body.body
-  | None -> false
+  | Body body -> block_has_loops body.body
+  | _ -> false
 
 let crate_get_item_meta (m : crate) (id : item_id) : Types.item_meta option =
   match id with
@@ -91,7 +91,27 @@ class ['self] map_crate =
       let is_global_initializer =
         self#visit_option self#visit_global_decl_id env is_global_initializer
       in
-      let body = self#visit_option self#visit_expr_body env body in
+      let body =
+        match body with
+        | Body b -> Body (self#visit_expr_body env b)
+        | TraitMethodWithoutDefault -> TraitMethodWithoutDefault
+        | Extern sym -> Extern (self#visit_string env sym)
+        | Intrinsic { name; arg_names } ->
+            Intrinsic
+              {
+                name = self#visit_string env name;
+                arg_names = List.map (self#visit_string env) arg_names;
+              }
+        | TargetDispatch targets ->
+            TargetDispatch
+              (self#visit_list
+                 (fun env (tgt, fref) ->
+                   (self#visit_string env tgt, self#visit_fun_decl_ref env fref))
+                 env targets)
+        | Opaque -> Opaque
+        | Missing -> Missing
+        | Error err -> Error (self#visit_error env err)
+      in
       {
         def_id;
         item_meta;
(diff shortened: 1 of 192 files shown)


Nadrieril (approved):
Love this, tyvm :)

Nadrieril:
Re argument names, we can't change `FunSig` because it's used in `TyKind::FnPtr`. I'm open to finding a solution though, argument names are useful documentation to preseve.

N1ark:
> Re argument names, we can't change `FunSig` because it's used in `TyKind::FnPtr`. I'm open to finding a solution though, argument names are useful documentation to preseve.

options:
1.  we could make `Arg.name` an `Option<String>`; we don't always have names for locals anyways so it's consistent
2. `FnSig<Meta>` with `inputs: Vec<(Ty, Meta)>` where `Meta` is either `()` or `String`
3. add a field `arg_names: Vec<String>` to `FunDecl`

i think I am happy with either of the three options

Nadrieril:
How about `arg_names: Vec<String>` inside `Body::Extern` and `Body::Intrinsic`? We could even unify the two variants if useful since they're semantically pretty similar (you tell me what's most useful for you).

N1ark:
i think i prefer both as separate variants just because it makes matching on the enum easier, and i handle both cases separately anyways

yes this sounds good, ill do it now :)

Nadrieril (approved):

N1ark:
This is ready :D @Nadrieril

charon-ml/src/LlbcAstUtils.ml
```diff
@@ -91,7 +91,27 @@ class ['self] map_crate =
...
+        | TargetDispatch targets ->
+            TargetDispatch
+              (self#visit_list
+                 (fun env (tgt, fref) ->
+                   (self#visit_string env tgt, self#visit_fun_decl_ref env fref))
+                 env targets)
+        | Opaque -> Opaque
+        | Missing -> Missing
+        | Error err -> Error (self#visit_error env err)
+      in
```
Nadrieril, line 114: you could derive that visitor on `body`, no?

charon-ml/src/Substitute.ml
```diff
@@ -430,13 +430,23 @@ let block_substitute (subst : subst) (blk : block) : block =
...
+      in
+      let body = block_substitute subst body.body in
+      (locals, body)
+  | TraitMethodWithoutDefault
+  | Extern _
+  | Intrinsic _
+  | TargetDispatch _
+  | Opaque
+  | Missing
+  | Error _ -> raise (Failure "Can't substitute a non-structured function body")
```
Nadrieril, line 449: Seems ok to return the body unmodified here

charon/src/ast/gast.rs
```diff
@@ -100,10 +100,17 @@ pub enum Body {
     /// The body of the function item we add for each trait method declaration, if the trait
     /// doesn't provide a default for that method.
     TraitMethodWithoutDefault,
-    /// Function declared in an `extern { ... }` block. The string is the foreign symbol name.
+    /// Function declared in an `extern { ... }` block.
```
Nadrieril, line 103: Was the string not the symbol name? What does the string represent then?

N1ark: oops yeah that should stay

charon-ml/src/LlbcAstUtils.ml
N1ark: hmm probably, but anyways this will change in #1102 where the visitor won't be possible, so i'd rather leave this as is

Nadrieril: sounds good

N1ark:
> Re argument names, we can't change `FunSig` because it's used in `TyKind::FnPtr`. I'm open to finding a solution though, argument names are useful documentation to preseve.

maybe that could change ?
https://github.com/rust-lang/rfcs/pull/3955

PR #1091, opened 2026-04-10
Add `--start-from-if-exists`

N1ark:
Add `--start-from-if-exists`, which acts as a `--start-from` that doesn't error when it fails to match


Nadrieril:
how about `--start-from-if-exists`?

N1ark:
sure :) will do later

PR #1099, opened 2026-04-16
Roll pinned Rust toolchain automatically

joshlf:


Nadrieril:
I don't know, the current process works pretty well. This would also add unnecessary pressure for us to fix these builds. Moreover jumping many nightlies at once probably reduces work because complex changes typically land over several PRs and updating for every interpediate state is wasteful.

joshlf:
Would it be better if, on failure, this just silently aborted, and only uploaded PRs which will definitely work? Ie, remove only the most manual and mindless part of the workflow?

Nadrieril:
Said differently, the resource we're currently optimizing for in this project is my personal capacity. Anything that unduly asks more of me is a loss. Keeping up-to-date with nightlies is not at all a priority overall

Nadrieril:
You could turn this into a script that I run myself and updates everything to a toolchain date I give it, that would help. I'm skeptical of the benefits of github action automation

Nadrieril:
I don't see a strong need for this, closing. Let me know if your constraints make this more important than I know of

PR #1033, opened 2026-02-24
Bump rustc to remove `ShallowInitBox`

N1ark:
Continuation of #1030
Fixes #973 
Related to https://github.com/rust-lang/rust/pull/148190

Added a transformation pass to make `vec!` safe. For the following original MIR:
```
((((*_4)).1).0).0 = [move _5]
storage_dead(_5)
_3 = move _4
conditional_drop[{impl Destruct for alloc::boxed::Box<T>[@TraitClause0, @TraitClause1]}<MaybeUninit<[u8; 1 : usize]>[{built_in impl Sized for [u8; 1 : usize]}], Global>[{built_in impl MetaSized for MaybeUninit<[u8; 1 : usize]>[{built_in impl Sized for [u8; 1 : usize]}]}, {built_in impl Sized for Global}]] _4
storage_dead(_4)
vec_2 = box_assume_init_into_vec_unsafe<u8, 1 : usize>[{built_in impl Sized for u8}](move _3)
```
we want something akin to
```
_T = [move _5]
_3 = Box::write(move _4, move _T)
_B = unsize<Box<[T; N]>, Box<[T]>>(move _3)
vec_2 = std::slice::{impl}::into_vec(move _B)
```

ci: use https://github.com/AeneasVerif/aeneas/pull/1033

PR #1100, opened 2026-04-16
Roll pinned Rust toolchain

joshlf:


joshlf:
Looks like this update breaks expectations from Aeneas and Eurydice. How do you normally handle that? Is there a recommended graceful upgrade path?

charon/hax-frontend/src/types/def_id.rs
```diff
@@ -20,8 +20,6 @@ sinto_reexport!(rustc_span::symbol::Symbol);
 sinto_reexport!(rustc_span::symbol::ByteSymbol);
 
 /// Reflects [`rustc_hir::def::DefKind`]
-#[derive(AdtInto)]
-#[args(<S>, from: rustc_hir::def::DefKind, state: S as tcx)]
```
joshlf, line 24: Calling this out – not sure of the significance of this change or why the agent did it.

charon/hax-frontend/src/types/def_id.rs
```diff
@@ -52,7 +50,6 @@ pub enum DefKind {
     ForeignMod,
     AnonConst,
     InlineConst,
-    #[disable_mapping]
```
joshlf, line 55: Calling this out – not sure of the significance of this change or why the agent did it.

charon/hax-frontend/src/types/ty.rs
```diff
@@ -655,57 +655,50 @@ impl Alias {
...
-                // impl<T, U> Trait for Result<T, U>
-                // where
-                //     for<'a> &'a Result<T, U>: IntoIterator,
-                //     for<'a> <&'a Result<T, U> as IntoIterator>::Item: Copy,
-                // {}
-                // ```
-                // the `&'a Result<T, U> as IntoIterator` trait ref has escaping bound variables
-                // yet we dont have a binder around (could even be several). Binding this correctly
-                // is therefore difficult. Since our trait resolution ignores lifetimes anyway, we
-                // just erase them. See also https://github.com/hacspec/hax/issues/747.
```
joshlf, line 687: Calling this out – should I keep this comment, or was the agent right to remove it?

Nadrieril:
there is already https://github.com/AeneasVerif/charon/pull/1033 waiting that bumps it a bit forward, let's wait for that

charon/hax-frontend/src/types/ty.rs
Nadrieril: definitely should be kept

PR #1102, opened 2026-04-17
Clean json generation

N1ark:
This PR sets up the codebase to enable the addition of a new serialisation format, using serde's [postcard](https://docs.rs/postcard/latest/postcard/) format.

This cleans up a lot of ad-hoc behaviours in serialisation: the crate's type is now entirely derived, and it's deserialiser entirely generated too, with only a tiny wrapper to get a more nicely shaped crate out of it.

More precisely:
- Split `generate-ml/main.rs` (previously 1400 lines) into several modules: `to_ocaml_ty` for the generation of OCaml types, `of_json` for the generation of the JSON deserializer, and `util`.
- Add `Generated_FullAst.ml` (exposed in `GAst.ml`), which contains the "root" types of the AST that weren't generated before: `translated_crate` and `body` (and I also moved some related types in there)
- Merged all `_OfJson` files, to simplify translation
- Merged all `Print_` files, to also simplify things 

I tried splitting the commits to be as reviewable as possible; note that OCaml does not compile from commit 4 (Add `FullAst`) to commit 10 (Merge all `OfJson`), since making everything compile in one go would have been too big of a commit.

ci: use https://github.com/AeneasVerif/aeneas/pull/959
ci: use https://github.com/AeneasVerif/eurydice/pull/397


N1ark:
~~hmm the OCaml tests pass for me locally, not sure how this fails given files should be deserialized first? could it be Nix somehow reusing old `.llbc` files or not re-building something...?~~ i forgot to push a change :p 

if you're happy with this I'll update Aeneas and Eurydice

charon/src/ast/gast.rs
```diff
@@ -110,6 +110,7 @@ pub enum Body {
     /// We encountered an error while translating this body.
     #[drive(skip)]
     #[serde_state(stateless)]
+    #[charon::rename("ErrorBody")]
```
Nadrieril: I'd go with `charon::variants_suffix("Body")`

charon-ml/src/GAstUtils.ml
```diff
@@ -243,6 +243,9 @@ class ['self] filter_decl_id =
     method visit_trait_decl_id _ (id : TraitDeclId.id) = Some id
     method visit_trait_impl_id _ (id : TraitImplId.id) = Some id
 
+    method visit_trait_type_constraint_id _ (id : TraitTypeConstraintId.id) =
+      Some id
```
Nadrieril: Why is this needed now? This is not an item id

charon/src/ast/krate.rs
```diff
@@ -192,6 +192,12 @@ pub struct TranslatedCrate {
     #[serde(with = "SeqHashMapToArray::<TargetTriple, TargetInfo>")]
     pub target_information: SeqHashMap<TargetTriple, TargetInfo>,
 
+    /// The translated files. This field must come before any field containing spans,
+    /// as the files must be translated before we can translate any span.
```
Nadrieril:
```suggestion
    /// The translated files. This field must come before any field containing spans,
    /// as the OCaml deserialization of spans requires the files to be deserialized already.
```

charon/src/bin/generate-ml/util.rs
```diff
@@ -203,4 +214,22 @@ impl<'a> GenerateCtx<'a> {
     pub fn names_to_type_id_set(&self, data: &[&str]) -> HashSet<TypeDeclId> {
         data.iter().map(|name| self.id_from_name(name)).collect()
     }
+
+    pub fn with_item<F, T>(&mut self, new_item: &TypeDecl, f: F) -> T
+    where
+        F: FnOnce(&mut Self) -> T,
+    {
+        let mut item = Some(new_item.clone());
```
Nadrieril: Cloning the whole `TypeDecl` feels a bit excessive. You only need the output of `type_to_ocaml_ident_raw` right?

charon/src/bin/generate-ml/util.rs
```diff
@@ -47,6 +47,10 @@ pub fn make_ocaml_ident(name: &str) -> String {
     name
 }
 
+pub fn name_str(name: &Name) -> &String {
+    name.name.last().unwrap().as_ident().unwrap().0
+}
```
Nadrieril: May as well move that to be a method on `Name`, it's often useful

Nadrieril (commented):
That's so much better :') Looks great to me

charon/src/bin/generate-ml/util.rs
N1ark: yes indeed, changed it to only get the identifier :)

charon-ml/src/GAstUtils.ml
N1ark: true, i wanted to see if generating it like everything else simplified things but it broke that :p reverted

charon/src/bin/generate-ml/util.rs
```diff
@@ -217,17 +210,14 @@ impl<'a> GenerateCtx<'a> {
...
+        self.current_item = old_name;
         res
     }
 
     pub fn use_opt_index_map(&self) -> bool {
-        match &self.current_item {
-            None => false,
-            Some(ty) => self.type_to_ocaml_ident_raw(ty).0 == "translated_crate",
-        }
+        self.current_item == Some("translated_crate".to_string())
```
Nadrieril, line 221: No need to fix this but fyi `self.current_item.as_deref() == Some("translated_crate")` exists :)

Nadrieril:
Let's goooo

PR #1103, opened 2026-04-19
docs: clarify opacity semantics in what_charon_translates.md

oliver-butterley:
Several clarifications to the item selection documentation:

- Distinguish *opacity* (how much of an item is translated, determined solely by its own name/annotations) from *reachability* (whether the item is queued at all, which depends on the exploration path). Items only reachable through an opaque body are never queued.

- Explain the scope difference between `#[charon::opaque]` and `--opaque`: the source annotation applies only to the annotated item; items nested within it (functions in a module, methods in an impl block) are separate items that keep their own opacity. The CLI flag, via prefix matching, makes the item and all subitems opaque.

- Fix an inaccuracy: the default base pattern is `_ → Foreign`, not `_ → Opaque`. There is no `--foreign` flag, so the default cannot be expressed directly via CLI flags, and using `--opaque` in its place would silently change behaviour for foreign types.

- Fix `--extract-opaque-bodies` description: it changes the base `_ → Foreign` entry to `_ → Transparent`, not merely `--include *`.

- Fix terminology: "default method" → "provided method" (trait method with a default body).

### DIFF
--- a/docs/what_charon_translates.md
+++ b/docs/what_charon_translates.md
@@ -8,7 +8,7 @@ Each item has an **opacity** level[^1]:
 
 1. **Transparent** — the item is fully translated.
 2. **Foreign** — the default for items outside the current crate. Translation depends on normal Rust visibility: for types, translate fully if it's a struct with all-public fields or an enum; for other items this is equivalent to Opaque.
-3. **Opaque** — only the name and signature are translated, not the contents. For functions and globals, the body is not translated. For types (structs, enums, unions), the fields/variants are not translated. For traits and trait impls, provided methods (those with a default body in the trait) are only translated if used somewhere else. For modules, the contents are not explored (but items referenced from elsewhere are still translated).
+3. **Opaque** — only the name and signature are translated, not the contents. For functions and globals, the body is not translated. For types (structs, enums, unions), the fields/variants are not translated. For traits and trait impls, methods with a default body are only translated if mentioned (used or overridden) anywhere in the final crate. For modules, the contents are not explored (but items referenced from elsewhere are still translated).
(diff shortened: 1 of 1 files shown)


docs/what_charon_translates.md
```diff
@@ -8,7 +8,7 @@ Each item has an **opacity** level[^1]:
 
 1. **Transparent** — the item is fully translated.
 2. **Foreign** — the default for items outside the current crate. Translation depends on normal Rust visibility: for types, translate fully if it's a struct with all-public fields or an enum; for other items this is equivalent to Opaque.
-3. **Opaque** — only the name and signature are translated, not the contents. For functions and globals, the body is not translated. For types (structs, enums, unions), the fields/variants are not translated. For traits and trait impls, the default method is  only translated if it is used somewhere else. For modules, the contents are not explored (but items referenced from elsewhere are still translated).
+3. **Opaque** — only the name and signature are translated, not the contents. For functions and globals, the body is not translated. For types (structs, enums, unions), the fields/variants are not translated. For traits and trait impls, provided methods (those with a default body in the trait) are only translated if used somewhere else. For modules, the contents are not explored (but items referenced from elsewhere are still translated).
```
Nadrieril: I'm trying to move away from the "provided method" terminology because it's a bit ambiguous. Moreover there's a bit of a "used globally thing going on": I would say "For traits and trait impls, methods with a default body are only translated if mentioned (used or overridden) anywhere in the final crate"

Nadrieril (commented):
V nice, ty for this. Just one comment

docs/what_charon_translates.md
oliver-butterley: yep! phrasing changed as suggested

PR #1104, opened 2026-04-20
feat: refactor Charon.toml to use Cargo.toml and add support for start_from, hide_marker_traits

oliver-butterley:
The Charon.toml config file previously supported only four fields (extract_opaque_bodies, include, opaque, exclude) and rustc flags.

This PR refactors the functionality to avoid the Charon.toml file and instead uses a [metadata] section in Cargo.toml. Moreover we extend functionality to cover other commonly-needed options:

- start_from / start_from_if_exists / start_from_attribute / start_from_pub — control translation entry points
- hide_marker_traits — suppress marker traits from output

The --help text for the `cargo` subcommand now mentions that the [package.metadata.charon] section in Cargo.toml is read when present. 

Note: The charon option `preset` is not exposed in the config. The rationale being that this depends on which tool will consume Charon's output rather than on the crate itself, so it doesn't belong in the crate's Cargo.toml.


oliver-butterley:
I noticed there is an existing integration test for the config file (`tests/cargo/toml/`). It could be extended to cover the new fields, in particular `[cargo] flags` and `start_from`. 

Concrete possibilities:

- A project with a feature-gated function and `[cargo] flags = ["--features", "my_feature"]` in `Charon.toml`: confirms the function appears in the output (analogous to the existing `[rustc] flags` test).
- A project with two modules and `start_from = ["crate::included"]`: confirms the excluded module is absent from the output.
- A bad preset name in `Charon.toml` should produce a clear error message: I could also do a test for this although there is likely a slight maintenance burden here since the panic output includes an internal file:line location that would make change with later updates.

Currently such tests are drafted in another branch (config-file-tests), let me know if they should be added to this PR.

Nadrieril:
I kinda want to rip out Charon.toml and replace it with a `[metadata]` section in `Cargo.toml` as is more standard

Nadrieril:
would you wanna do taht instead?

oliver-butterley:
> I kinda want to rip out Charon.toml and replace it with a `[metadata]` section in `Cargo.toml` as is more standard

We are very flexible with our needs. Just needs to have something in order to have a convenient way to write the flags. E.g., our repos currently need to pass many flags like seen in [this](https://github.com/Beneficial-AI-Foundation/SparsePostQuantumRatchet-verify/blob/main/aeneas-config.yml).

oliver-butterley:
> would you wanna do taht instead?

OK! Will do.

charon/src/bin/charon/toml_config.rs
```diff
@@ -32,27 +40,53 @@ pub struct RustcTomlConfig {
...
-    /// options take precedence.
-    pub(crate) fn apply(self, mut config: CliOpts) -> CliOpts {
-        config.extract_opaque_bodies |= self.charon.extract_opaque_bodies;
-        config.include.extend(self.charon.include);
-        config.opaque.extend(self.charon.opaque);
-        config.exclude.extend(self.charon.exclude);
-        config.rustc_args.extend(self.rustc.flags);
+    /// Applies the options specified in the toml file to the cli options and cargo args.
+    /// Scalar options (preset, start_from_attribute): CLI value wins if set; TOML value used
+    /// otherwise. Boolean options: OR of CLI and TOML. List options: always merged (both combined).
```
Nadrieril:
```suggestion
    /// Applies the options specified in the toml file to the cli options and cargo args.
    /// CLI options take precedence when relevant.
```

charon/src/bin/charon/toml_config.rs
```diff
@@ -32,27 +40,53 @@ pub struct RustcTomlConfig {
...
+    pub(crate) fn apply(
+        self,
+        mut config: CliOpts,
+        cargo_args: &mut Vec<String>,
+    ) -> anyhow::Result<CliOpts> {
+        if let (None, Some(s)) = (&config.preset, self.preset) {
+            config.preset = Some(
+                Preset::from_str(&s, true /* ignore_case */).map_err(|_| {
+                    anyhow::anyhow!("Unknown preset {s:?} in Cargo.toml. Valid values: aeneas, eurydice, soteria, old-defaults, raw-mir, tests")
```
Nadrieril: Having the list written by hand here will absolutely go out of date. Is there a way to reuse `Clap`'s knowledge to emit a nicer error?

Nadrieril: Also I'd prefer not to ignore case

charon/src/bin/charon/toml_config.rs
```diff
@@ -32,27 +40,53 @@ pub struct RustcTomlConfig {
     pub flags: Vec<String>,
 }
 
+/// Extra arguments forwarded to `cargo build` (equivalent to flags after `--`).
+#[derive(Debug, Default, Deserialize)]
+pub struct CargoTomlConfig {
+    #[serde(default)]
+    pub flags: Vec<String>,
+}
```
Nadrieril: If `Cargo.toml` had a way to supply that we would use that; having our own feels like going against how cargo was designed, let's remove it.

charon/tests/cargo/toml-start-from/src/lib.rs
Nadrieril, line 1: Please merge these new tests into one, many cargo tests are not fun to have around.

docs/usage.md
```diff
@@ -9,14 +9,60 @@ provides various options and flags to tweak its behaviour: you can display a
 detailed documentation with `--help`.
 In particular, you can pretty-print the translated crate with both `--print-ullbc` and `--print-llbc`, depending on the Charon intermediate representation you wish to use.
 
-If there is a `Charon.toml` file at the root of your project, `charon` will also take options from it.
-The file supports the same options at the cli interface, except for the options that relate to
-input/output like `--print-llbc`. Example `Charon.toml`:
+Charon supports per-project configuration via the `[package.metadata.charon]` section in `Cargo.toml`.
```
Nadrieril:
```suggestion
Charon supports per-crate configuration via the `[package.metadata.charon]` section in `Cargo.toml`. 
```

docs/usage.md
```diff
@@ -9,14 +9,60 @@ provides various options and flags to tweak its behaviour: you can display a
 detailed documentation with `--help`.
 In particular, you can pretty-print the translated crate with both `--print-ullbc` and `--print-llbc`, depending on the Charon intermediate representation you wish to use.
 
-If there is a `Charon.toml` file at the root of your project, `charon` will also take options from it.
-The file supports the same options at the cli interface, except for the options that relate to
-input/output like `--print-llbc`. Example `Charon.toml`:
+Charon supports per-project configuration via the `[package.metadata.charon]` section in `Cargo.toml`. 
+Scalar options (e.g. `preset`) use the CLI value if given, otherwise the config file
+value. Boolean options are OR-ed. List options (e.g., `start_from`, `exclude`) are always merged.
```
Nadrieril: This is a rather excessive detail that fails to include the actually important info. In the future please review the English text in your PRs not just with "does this make sense" but with "If I were writing this, what would I write". I dislike the feeling that I'm the one reviewing your LLM's output.

```suggestion
A couple of Charon options can be passed as toml values there. The exact list
can be found by looking at `src/bin/charon/toml_config.rs`. If an option is set both as a CLI flag and as a toml value and we can't merge them, the CLI flag wins.
```

docs/usage.md
```diff
@@ -9,14 +9,60 @@ provides various options and flags to tweak its behaviour: you can display a
...
+
 ```toml
-[charon]
-extract_opaque_bodies = true
-[rustc]
-flags = ["--cfg", "abc"]
+[package.metadata.charon]
+# Preset to apply. Uses the same names as the CLI (e.g. "aeneas", "eurydice", "soteria",
+# "old-defaults", "raw-mir"). CLI --preset takes precedence.
```
Nadrieril: Don't include the list here

docs/usage.md
```diff
@@ -9,14 +9,60 @@ provides various options and flags to tweak its behaviour: you can display a
...
+preset = "aeneas"
+
+# Entry points for translation. Charon translates these items and everything they reference,
+# following the opacity rules. When absent (and no --start-from is given on the CLI), the whole
+# crate is translated. Uses name-matcher patterns.
```
Nadrieril: Don't document things here, it'll go out of date. Point users to the cli help.

docs/usage.md
```diff
@@ -9,14 +9,60 @@ provides various options and flags to tweak its behaviour: you can display a
...
+
+# Translate bodies of all items, including foreign ones (off by default).
+# extract_opaque_bodies = false
+
+# Hide marker traits (e.g. Send, Sync) from the output.
+# hide_marker_traits = false
+
+[package.metadata.charon.cargo]
+# Extra arguments forwarded to `cargo build` (equivalent to flags after `--` on the CLI).
+flags = ["--features", "my_feature"]
```
Nadrieril: We'll remove that

charon/src/bin/charon/toml_config.rs
```diff
@@ -1,29 +1,37 @@
...
-    pub charon: CharonTomlConfig,
-    #[serde(default)]
-    pub rustc: RustcTomlConfig,
-}
-
-/// The struct used to define the options available in `Charon.toml` files. These all mirror the
-/// corresponding cli option.
-#[derive(Debug, Default, Deserialize)]
-pub struct CharonTomlConfig {
+    pub preset: Option<String>,
```
Nadrieril: I'm wondering if we should include the preset here. The flags that make sense in a Cargo.toml are those that are relevant to the crate qua crate. A preset depends on the tool you're gonna use, which to me should be a separate thing.

docs/usage.md
oliver-butterley: >  In the future please review the English text in your PRs not just with "does this make sense" but with "If I were writing this, what would I write". I dislike the feeling that I'm the one reviewing your LLM's output.

I'm so sorry! This is a very valid complaint and I don't want to waste your time like this. I rushed and didn't carefully look at everything, I should have kept the PR as draft until it was properly ready again. I'll do better!

Nadrieril: To be clear it's not that it's wasting my time, it's that reviewing LLM code and human code are rather different exercises. When it's LLM I have no qualms ripping out a whole change and rewriting it myself, or being very terse and brutal in my reviews; wiith a human I want to propose my changes to their judgment and iterate together (and be friendly).

charon/src/bin/charon/toml_config.rs
oliver-butterley: Yes, that logic is convincing: the preset is not a property of the crate and so doesn't belong in cargo.toml. This functionality removed.

docs/usage.md
oliver-butterley: Removed

docs/usage.md
oliver-butterley: Very minimal text in usage,md now.

docs/usage.md
oliver-butterley: Also, it fair to be brutal to me if I fail to check that the llm has added foolish stuff to a file! 

I took a bit more time which I finally had and hopefully everything is cleaner now... I should explain that my expertise and experience is with Lean. I have a working knowledge of Rust but rather limited experience. I hope I'm helping and not making things less efficient for you.

charon/src/bin/charon/toml_config.rs
oliver-butterley: I took the option to remove preset from the possible config so these comments were never implemented.

charon/src/bin/charon/toml_config.rs
oliver-butterley: Of course! Removed.

oliver-butterley:
Thanks for your patience!

Summary of the latest updates following all the comments:

- Dropped [package.metadata.charon.cargo] flags entirely
- Dropped preset entirely. (it's tool-dependent rather than crate-dependent)
- The additional tests added in the previous version of the PR have been removed. What remains is the single test from before, updated to the latest functionality.
- docs/usage.md is now very concise info

Note: Updated read_toml reads ./Cargo.toml directly like was done before for Charon.toml. It's possible that for Cargo.toml there is a specific procedure which should be used but searching I didn't find a solution which was clearly better than the current version.

PR #1111, opened 2026-04-24
Add missing types in `find_primitive_impls`

N1ark:
It was missing f16 and f128 :)


Nadrieril:
Hah, that code dates from before they existed it seems :) Ty!

PR #1118, opened 2026-04-27
Add support for postcard format

N1ark:
Add a new output format to Charon, based on [postcard](https://github.com/jamesmunns/postcard/).

This format is:
- 2 to 10 times smaller on disk than JSON, 5x on average
- 3 times faster to deserialise than JSON
- much faster than JSON to serialise; i don't have a number here, but compiling zerocopy with tests went from 19s to 11s (including compilation)

The format can be selected with `--format <postcard|json>`, defaulting to JSON.

The vast majority of code is generated, and the postcard generation is very close to OCaml's. Furthermore, the OCaml decode uses `in_channel` to decode bytes, meaning the full postcard file is never entirely held into memory.

I modified `tests/ui` to also serialize to postcard on the side, and the OCaml tests now also deserialize these files and print some small comparison information:
```
Average deserialization time: JSON: 0.005708 seconds, Postcard: 0.001687 seconds (3.382520 times faster)
Size ratio (JSON/Postcard): min: 2.380217, max: 10.772649, avg: 5.969264, med: 5.691480
```

ci: use https://github.com/AeneasVerif/aeneas/pull/1004
ci: use https://github.com/AeneasVerif/eurydice/pull/403


Nadrieril:
Let's go all the way honestly: rip out the ocaml json deserializer and use postcard by default. We can keep `--format=json` on the Rust side for debugging and to make it easier for other projects to use.

EDIT: to avoid horrible crashes with an earlier version of charon trying to read a postcard file, actually let's keep json the default for a while.

Nadrieril:
One thing that feels important is errors: I took a lot of care that trying to read an llbc file from the wrong charon version gives you a nice error. With this PR, accidentally mixing json and postcard will blow up in your face.

Could you make sure we can nicely error when confronted with the wrong format? (And even detect the format if that's possible).

Also can you make sure that reading the wrong version errors early before trying to read the rest? That's more important actually, especially if we deprecate the json format.

N1ark:
will do ; version detection should work the same way as in JSON, but yes i can also try to add detection of the version one way or another
https://github.com/soteria-tools/charon/blob/78ecf307ccd17b70586693cf5c7ffcbff393bbe9/charon-ml/src/OfPostcard.ml#L5-L14

N1ark:
done in 66f0e7f; both deserializers now try reading the first 64 bytes and just figure out the format: ignoring whitespace, if a `{` is found it's JSON, anything else is postcard

I guess I should update aeneas/eurydice to use the new function rather than deserializing the file to JSON themselves

charon-ml/tests/Test_NameMatcher.ml
```diff
@@ -254,6 +253,39 @@ let annotated_rust_tests test_file =
 
   if all_pass then log#linfo (lazy "Name matcher tests: success") else exit 1
 
+let annotated_rust_tests_postcard test_file =
```
Nadrieril: This is supposed to test name matcher logic not serialization logic, I don't think it's useful to duplicate this test

charon/tests/ui.rs
```diff
@@ -287,6 +307,16 @@ fn perform_test(test_case: &Case) -> anyhow::Result<()> {
...
 
+    if output.status.success()
+        && !matches!(test_case.magic_comments.test_kind, TestKind::Skip)
+        && serialize.is_some()
+    {
+        build_command(test_case, Some(SerializationFormat::Postcard), &deps)?
+            .assert()
+            .try_success()
+            .map_err(|e| anyhow!(e.to_string()))?;
+    }
```
Nadrieril: That instantly doubles the test runtime, that's an issue for development. Can you add `--format=all` that serializes to all the formats instead?

charon/Cargo.toml
```diff
@@ -112,6 +112,7 @@ tracing.workspace = true
 ustr = { version = "1.1.0", features = ["serde"] }
 wait-timeout = { version = "0.2.0", optional = true }
 which = "7.0"
+postcard = { version = "1.1.3", features = ["use-std"] }
```
Nadrieril: nit: mind keeping the dependencies sorted by name?

Nadrieril (commented):
Looks great, we'd been wanting to chang eformats for a long while now. My only issue is running tests twice.

charon/tests/ui.rs
Nadrieril: Alternatively, serialize as postcard in tests and add a `charon change-format` command that converts between formats

N1ark:
@Nadrieril done :) added `--format all`. note that the interaction with `--dest` is a bit weird but its the best i could come up with: it uses whatever file path is given, but adds the right extension for the format on top, so e.g. `--dest foo/bar --format all` will generate `foo/bar.llbc` and `foo/bar.llbc.postcard`; i think it's a good compromise

Nadrieril (approved):
Let's gooo

PR #1121, opened 2026-05-05
feat: add a node for char values

clementblaudeau:
This PR adds a wrapper around char values. It is used by https://github.com/AeneasVerif/aeneas/pull/917. In Aeneas, it allows handling char values without exposing the internal representation as a `UChar.t`

### DIFF
--- a/charon/src/bin/generate-ml/main.rs
+++ b/charon/src/bin/generate-ml/main.rs
@@ -264,7 +264,7 @@ fn generate_ml(
                     ancestors: &["big_int"],
                     name: "literal",
                     reduce: true,
-                    extra_types: &[],
+                    extra_types: &["char_value"],
                 })), &[
                     "Literal",
                     "IntegerTy",
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.191"
+let supported_charon_version = "0.1.192"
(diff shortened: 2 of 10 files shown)


Nadrieril:
On the Rust side this change does not make sense. Could you make this change only on the ocaml side without affecting the Rust side? You should just special case chars in the ocaml generator

clementblaudeau:
Ok, I'll revert the changes on the rust ast side and only affect the ml generator then

clementblaudeau:
I've updated the patch : now the rust ast is untouched, but during ml-generation an extra node for chars is added. The fixes are partially Claude generated, partially hand-written.

charon/src/bin/generate-ml/main.rs
```diff
@@ -157,7 +158,8 @@ fn main() -> Result<()> {
         }
     }
 
-    let crate_data: TranslatedCrate = charon_lib::deserialize_llbc(&charon_llbc)?;
+    let mut crate_data: TranslatedCrate = charon_lib::deserialize_llbc(&charon_llbc)?;
+    patch_in_char_value(&mut crate_data);
```
Nadrieril: That's an ugly hack I'm sorry. Please patch in the ocaml generation, not the LLBC file we got from charon. Unless you tried that and it didn't work?

Nadrieril (approved):
Much better :) Looks good, you'll need to rebase once https://github.com/AeneasVerif/charon/pull/1122 gets merged and increment the version number in `Cargo.toml` to satisfy CI

Nadrieril:
This is waiting on a version number bump

clementblaudeau:
I've rebased on main and bumped the version number. I've also squashed the commits into a single one. How does it look ? :)

Nadrieril:
Looks great :)

PR #1122, opened 2026-05-05
Fix: handle dangling references left by `--remove-adt-clauses`

### LINKED ISSUE #1120: Bug: error on dereferencing a `Copy` type inside a closure
**Code snippet to reproduce the bug**:

```rust
fn f<T: Copy>(x: &T) {
    let _c = || *x;
}
```

**Charon version**: ed22146b1cd4d0b578006a58b3299d41ecbe0fd4

**Charon command**: charon cargo --preset=aeneas 

**Charon output**:

```
error: Type error after transformations:
       Found incorrect clause var: Bound(1, 0)
       Visitor stack:
         charon_lib::ast::types::TraitRef
         charon_lib::ids::index_vec::IndexVec<charon_lib::ast::types::vars::TraitClauseId, charon_lib::ast::types::TraitRef>
         charon_lib::ast::types::GenericArgs
         alloc::boxed::Box<charon_lib::ast::types::GenericArgs>
         charon_lib::ast::types::TraitImplRef
         charon_lib::ast::types::RegionBinder<charon_lib::ast::types::TraitImplRef>
         charon_lib::ast::types::ClosureInfo
         charon_lib::ast::gast::ItemSource
         charon_lib::ast::types::TypeDecl
       Binding stack (depth 2):
         0: 
         1: <'_0, T>
 --> src/main.rs:4:14
  |
4 |     let _c = || *x;
  |              ^^^^^
```

maximebuyse:
Fixes #1120 

This is a claude implementation for the design from https://github.com/AeneasVerif/charon/issues/1120#issuecomment-4378918055

### DIFF
--- a/charon/src/ast/types.rs
+++ b/charon/src/ast/types.rs
@@ -110,6 +110,11 @@ pub enum TraitRefKind {
     /// A trait implementation that is computed by the compiler, such as for built-in trait
     /// `Sized`. This morally points to an invisible `impl` block; as such it contains
     /// the information we may need from one.
+    ///
+    /// Also used as a placeholder for trait clauses that were stripped by the
+    /// `--remove-adt-clauses` pass: the original `Clause` reference is replaced with a
+    /// `BuiltinOrAuto { builtin_data: RemovedAdtClause, .. }`. See
+    /// [`BuiltinImplData::RemovedAdtClause`].
     BuiltinOrAuto {
         #[drive(skip)]
         builtin_data: BuiltinImplData,
@@ -155,6 +160,10 @@ pub enum BuiltinImplData {
     FnOnce,
     Copy,
     Clone,
+    /// Placeholder used by the `--remove-adt-clauses` pass when it strips a trait clause from a
+    /// type declaration. References to the removed clause are rewritten as
+    /// `BuiltinOrAuto { builtin_data: RemovedAdtClause, .. }`.
+    RemovedAdtClause,
 }
 
 /// A reference to a trait.
--- a/charon/src/transform/simplify_output/remove_adt_clauses.rs
+++ b/charon/src/transform/simplify_output/remove_adt_clauses.rs
@@ -1,19 +1,158 @@
+//! `--remove-adt-clauses` strips trait clauses from type declarations when it's possible to do so.
+//! Because it's not possible to recover associated type information when we remove clauses,
+//! we don't remove clauses if any of them have associated types. For that reason,
+//! this flag is best used with `--lift-associated-types`.
+//!
+//! Every reference to a clause removed in this way is replaced with a `TraitRefKind::BuiltinOrAuto
+//! { builtin_data: RemovedAdtClause, .. }`.
+
+use std::collections::HashSet;
+
+use derive_generic_visitor::*;
+
 use crate::ast::*;
+use crate::common::CycleDetector;
+use crate::ids::{IndexMap, IndexVec};
 use crate::transform::{TransformCtx, ctx::TransformPass};
 
-#[derive(Visitor)]
-struct RemoveAdtClausesVisitor;
+/// Compute whether a trait has associated types, even through supertraits.
+fn has_assoc_types(
+    translated: &TranslatedCrate,
+    cache: &mut IndexMap<TraitDeclId, CycleDetector<bool>>,
+    id: TraitDeclId,
+) -> bool {
+    if cache[id].start_processing() {
+        let result = match translated.trait_decls.get(id) {
+            Some(tdecl) => {
+                !tdecl.types.is_empty()
+                    || tdecl
+                        .implied_clauses
+                        .iter()
+                        .any(|p| has_assoc_types(translated, cache, p.trait_.skip_binder.id))
+            }
+            None => false,
+        };
+        cache[id].done_processing(result);
+    }
+    match &cache[id] {
+        CycleDetector::Processed(b) => *b,
+        CycleDetector::Cyclic | CycleDetector::Processing => false,
+        CycleDetector::Unprocessed => unreachable!(),
+    }
+}
 
-impl VisitAstMut for RemoveAdtClausesVisitor {
+/// ADTs that have at least one trait clause pointing at a trait with associated types
+/// (transitively). We leave these ADTs entirely untouched.
+fn untouchable_adts(translated: &TranslatedCrate) -> HashSet<TypeDeclId> {
+    let mut cache: IndexMap<TraitDeclId, CycleDetector<bool>> = translated
+        .trait_decls
+        .map_ref_opt(|_| Some(CycleDetector::Unprocessed));
+    translated
+        .type_decls
+        .iter()
+        .filter(|d| {
+            d.generics
+                .trait_clauses
+                .iter()
+                .any(|c| has_assoc_types(translated, &mut cache, c.trait_.skip_binder.id))
+        })
+        .map(|d| d.def_id)
+        .collect()
+}
+
+#[derive(Visitor)]
+struct RemoveAdtClausesVisitor<'a> {
+    translated: &'a TranslatedCrate,
+    untouchable_adts: &'a HashSet<TypeDeclId>,
+    binder_stack: BindingStack<GenericParams>,
+}
+
+impl VisitorWithBinderStack for RemoveAdtClausesVisitor<'_> {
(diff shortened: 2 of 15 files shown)


charon/src/ast/types.rs
```diff
@@ -155,6 +160,11 @@ pub enum BuiltinImplData {
     FnOnce,
     Copy,
     Clone,
+    /// Placeholder used by the `--remove-adt-clauses` pass when it strips a trait clause from a
+    /// type declaration. References to the removed clause are rewritten as
+    /// `BuiltinOrAuto { builtin_data: RemovedAdtClause, .. }` with empty `parent_trait_refs` and
+    /// `types`. The trait declaration is preserved in the surrounding `trait_decl_ref`.
```
Nadrieril: Well actually that's a problem too: `parent_trait_refs` and `types` should match the trait definition. For example `Copy` has a `Clone` supertrait so a `BuiltinOrAuto` `TraitRef` for `Copy` must have an entry for the `Clone` supertrait in its `parent_trait_refs` list.

Please change two things:
- skip removing an ADT clause if the trait has any associated types (just leave the whole ADT untouched in that case, we can add more precision later if ever needed);
- for the removed clauses, fill in the `parent_trait_refs` with more `BuiltinOrAuto { builtin_data: RemovedAdtClause, .. }` trait refs, and recursively.

charon/src/transform/typecheck_and_unify.rs
```diff
@@ -412,8 +412,13 @@ impl VisitAstMut for TypeCheckVisitor<'_> {
...
                 parent_trait_refs,
                 types,
-                ..
+                builtin_data,
             } => {
+                if matches!(builtin_data, BuiltinImplData::RemovedAdtClause) {
+                    // Placeholder for a clause stripped by `--remove-adt-clauses`. Its
+                    // parent_trait_refs and types are intentionally empty.
+                    return;
+                }
```
Nadrieril: Nope, this check is here for a reason (otherwise we may as well just accept the dangling clause reference you know, but that's sad)

charon/src/ast/types.rs
maximebuyse: Makes sense, I pusehd a fix, again AI-generated so not sure it is using the best architecture to know if traits have an associated type, and to get the parent traits.

charon/src/transform/simplify_output/remove_adt_clauses.rs
```diff
@@ -1,28 +1,203 @@
...
+//! For ADTs we do strip, this invalidates any reference to those clauses still present in the AST
+//! -- in particular the `RegionBinder<TraitImplRef>` values stored in a closure ADT's
+//! `ItemSource::Closure { info }`, whose generic args contain `Clause(...)` refs that previously
+//! pointed at the closure ADT's parent-inherited clauses.
+//!
+//! After clearing the clauses, we walk the AST with a binder stack and replace any dangling
+//! `Clause(...)` reference with a `BuiltinOrAuto { builtin_data: RemovedAdtClause, .. }`
+//! placeholder whose `parent_trait_refs` are filled recursively from the trait's implied clauses,
+//! so the synthesized ref matches the shape declared by the trait (e.g. a `Copy` placeholder
+//! must carry a `Clone` parent ref since `Copy: Clone`).
```
Nadrieril:
```suggestion
//! `--remove-adt-clauses` strips trait clauses from type declarations when it's possible to do so.
//! Because it's not possible to recover associated type information when we remove clauses,
//! we don't remove clauses if any of them have associated types. For that reason,
//! this flag is best used with `--lift-associated-types`.
//!
//! Every reference to a clause removed in this way is replaced with a `TraitRefKind::BuiltinOrAuto
//! { builtin_data: RemovedAdtClause, .. }`. 
```

charon/src/transform/simplify_output/remove_adt_clauses.rs
```diff
@@ -1,28 +1,203 @@
...
+                blocked.insert(d.def_id);
+                changed = true;
+            }
+        }
+        if !changed {
+            break;
+        }
+    }
+    blocked
+}
```
Nadrieril: That's not acceptable complexity: a couple traits could cause this to iterate of all the traits in the crate several times. Please make a memoized recursive implementation instead. You may use the `CycleDetector` helper that's used in other passes.

charon/src/transform/simplify_output/remove_adt_clauses.rs
```diff
@@ -1,28 +1,203 @@
...
         }
-        let _ = ctx.translated.drive_mut(&mut RemoveAdtClausesVisitor);
+        let blocked = traits_with_assoc_types(&ctx.translated);
+        let untouchable = untouchable_adts(&ctx.translated, &blocked);
+        let mut first = RemoveAdtClausesVisitor {
+            untouchable_adts: &untouchable,
+            removed_any: false,
+        };
+        let _ = ctx.translated.drive_mut(&mut first);
+        if !first.removed_any {
```
Nadrieril: Please remove this shortcutting and make sure to traverse the crate only once, as this is by far the things that takes most time

charon/src/transform/simplify_output/remove_adt_clauses.rs
maximebuyse: Ok, I pushed a new version. I hope this is better. 
I thought this would be a small fix but this is more complex than I imagined. I don't want to take too much of your time for reviews (and I don't know charon well enough to be efficient in this work). If this looks too far from something we can merge, let me know and I'll close the PR.

charon/src/transform/simplify_output/remove_adt_clauses.rs
```diff
@@ -1,20 +1,164 @@
...
+                        .iter()
+                        .any(|p| has_assoc_types(translated, cache, p.trait_.skip_binder.id))
+            }
+            None => false,
+        };
+        cache[id].done_processing(result);
+    }
+    match &cache[id] {
+        CycleDetector::Processed(b) => *b,
+        CycleDetector::Cyclic | CycleDetector::Processing => true,
```
Nadrieril:
```suggestion
        CycleDetector::Cyclic | CycleDetector::Processing => false,
```

charon/src/transform/simplify_output/remove_adt_clauses.rs
```diff
@@ -1,20 +1,164 @@
...
+
 use crate::ast::*;
+use crate::common::CycleDetector;
+use crate::ids::{IndexMap, IndexVec};
 use crate::transform::{TransformCtx, ctx::TransformPass};
 
+/// Memoized recursion: a trait id has associated types if it directly declares any, or if any
+/// transitive supertrait does. Cyclic chains (which trait inheritance shouldn't produce, but the
+/// helper handles defensively) are treated conservatively as "has assoc types" so we err on the
+/// side of preserving clauses.
```
Nadrieril: You'll notice that AI-generated doc typically describes the how instead of the what, which is rarely what we want.
```suggestion
/// Compute whether a trait has associated types, even through supertraits.
```
Also in this case there is no sense in treating cycles conservatively: we can correctly compute whether a cycle involves assoc types with the current architecture.

charon/src/transform/simplify_output/remove_adt_clauses.rs
```diff
@@ -1,20 +1,164 @@
...
     fn enter_type_decl(&mut self, decl: &mut TypeDecl) {
+        if self.untouchable_adts.contains(&decl.def_id) {
+            return;
+        }
         decl.generics.trait_clauses.clear();
         decl.generics.trait_type_constraints.clear();
+        // The wrapper has already pushed the (uncleared) generic params onto the binder stack.
+        // Replace the top with the cleared version so dangling-clause detection works as we
+        // descend into the body.
+        if !self.binder_stack.is_empty() {
```
Nadrieril: nit: I'd rather not add this `empty` check because there really should be a binder level here

Nadrieril (changes requested):
Looks good, just two nits!

Nadrieril:
Looks good! https://github.com/AeneasVerif/charon/pull/1127 just got merged; please rebase (and may as well squash) your PR, and because you've modified the AST you'll also need to increment the version number in `Cargo.toml` then run `make test` in the root to propagate that number.

Nadrieril (approved):

PR #1160, opened 2026-05-15
Fix union layout variants and field offsets

N1ark:
Union layouts are currently translated as having no variants, which is wrong (given they always have the trivial variant 0). Changed it so they have one variant, and all fields are encoded at offset 0; this is hardcoded for now because it's what Rust does but note [this isn't a guarantee](https://doc.rust-lang.org/reference/items/unions.html#r-items.union.fields.offset) (though it seems like [it probably will be](https://github.com/rust-lang/unsafe-code-guidelines/issues/595))

Another doubt I have is that right now for enums with only one uninhabited variant we set all other variant layouts to an empty value with no fields; I'm not _sure_ this is correct. Can one not write values (and thus enums) field-by-field, regardless of uninhabitedness? i.e.
```rust
enum Foo {
	Bar,
	Baz { x: u32, y: !, z: u32 }
}

let foo: Foo;
set_discriminant!(foo, Baz);
foo.x = 0;
foo.z = 0;
```

Or is it instant UB to just set the discriminant to an uninhabited variant?

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_types.rs
+++ b/charon/src/bin/charon-driver/translate/translate_types.rs
@@ -471,7 +471,8 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
                 r_abi::FieldsShape::Arbitrary { offsets, .. } => {
                     offsets.iter().map(|o| o.bytes()).collect()
                 }
-                r_abi::FieldsShape::Primitive | r_abi::FieldsShape::Union(_) => IndexVec::default(),
+                r_abi::FieldsShape::Union(n) => vec![0; n.get()].into(),
+                r_abi::FieldsShape::Primitive => IndexVec::default(),
                 r_abi::FieldsShape::Array { .. } => panic!("Unexpected layout shape"),
             };
             VariantLayout {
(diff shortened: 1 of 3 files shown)


charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -619,25 +620,31 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                                field_offsets: IndexVec::default(),
+                                uninhabited: true,
+                                tagger: vec![],
+                            })
+                            .collect();
+                        variant_layouts[variant_id] = translate_variant_layout(&layout, vec![]);
+                        variant_layouts
+                    }
+                    r_abi::FieldsShape::Union(_) => {
+                        vec![translate_variant_layout(&layout, vec![])].into()
```
Nadrieril, line 643: nit: I believe that works:
```suggestion
                        [translate_variant_layout(&layout, vec![])].into()
```

Nadrieril:
> Another doubt I have is that right now for enums with only one uninhabited variant we set all other variant layouts to an empty value with no fields; I'm not _sure_ this is correct. Can one not write values (and thus enums) field-by-field, regardless of uninhabitedness?

Oh yep you're totaly right, they should have field layouts. It's still unsettled what exactly is or isn't UB here, but regardless the `TypeDecl` says there are fields so the corresponding variant layouts should have the same number of fields.

EDIT: I opened https://github.com/AeneasVerif/charon/issues/1166

> Or is it instant UB to just set the discriminant to an uninhabited variant?

It is! But that's unrelated

PR #1226, opened 2026-06-01
Bump rustc to nightly-2026-05-31

N1ark:
Bump rustc again! Our current toolchain is causing me problems in Obol so i'd like to move on from it quickly 🫠 

Relevant PRs:
- `Unnormalized<_>`: i tried adding the right calls here and there, but feel free to comment on places where i shouldn't skip normalization etc. Hopefully this improves as rustc's adoption of `Unnormalized` improves.
   https://github.com/rust-lang/rust/pull/156224
- `Rvalue::Reborrow`: for now we don't translate it
  https://github.com/rust-lang/rust/pull/151753
- `Rvalue::Use(_, WithRetag)`: added it, and added it in the transformations. I tried to be liberal and use `WithRetag::No` by default, to avoid introducing UB
  https://github.com/rust-lang/rust/pull/154341

ci: use https://github.com/AeneasVerif/aeneas/pull/1094

ci: use https://github.com/AeneasVerif/eurydice/pull/418


charon/src/bin/charon-driver/translate/translate_bodies.rs
```diff
@@ -1037,9 +1043,12 @@ impl<'tcx> BlockTransCtx<'tcx, '_, '_, '_> {
...
                         Operand::Copy(p) | Operand::Move(p) => {
-                            return Ok(Rvalue::Use(Operand::Copy(
-                                p.project(ProjectionElem::PtrMetadata, tgt_ty.clone()),
-                            )));
+                            return Ok(Rvalue::Use(
+                                Operand::Copy(
+                                    p.project(ProjectionElem::PtrMetadata, tgt_ty.clone()),
+                                ),
+                                WithRetag::No,
+                            ));
```
Nadrieril, line 1051: Why not a retag here? The only non-retagging copy is `CopyForDeref`, no?

N1ark: it's a metadata copy so i just assume a retag here is worthless, so i avoid doing one (since i don't believe we do retags for vtable pointers)

Issue #1168, opened 2026-05-17
Bug: constant unevaluation loses padding bytes


Nadrieril:
I don't know yet if this is observable, but there are cases where Charon goes from a MIR-computed constant, which has info about bytes inside padding, to a type-level constant, which doesn't. We should make extra sure to do something sound there.

Issue #1167, opened 2026-05-17
Clarify the evaluation of named constants


Nadrieril:
The first approximation we do is https://github.com/AeneasVerif/charon/issues/1168. Representationally this may not change anything since we already have `ConstantExprKind::Memory`.

Then there's the question of constant evaluation for named constants: today they always have an initializer function; especially in mono mode it makes sense to want the evaluated constant instead. I'm thinking we could represent both options by doing:
```rust
struct GlobalDecl {
  ...
  value: ConstantExpr,
}
enum ConstantExprKind {
  ...
  /// A call to a `const fn`
  Call(FnPtr, Vec<ConstantExpr>),
}
```

Nadrieril:
The eval situation is actually quite a bit complicated: when I see a ty::Const, I can always get a valtree and I can sometimes recover a "top-level" item this comes from. When I see a mir::Const, I can always get bytes and I can sometimes recover the "top-level" item this comes from, and I can also attempt to uneval into a valtree. And then given a top-level item I can get the initializer function or evaluate to valtree/bytes.

Moreover, constants can be found in two locations: inside a body, or as generic arguments (with the in-between case of pointer metadata constants, I think they're more like generic args?). A ty::Const can be found in either but a mir::Const can only be found in a body.

I _think_ a given top-level item is either of body nature or of type-level nature but not both: the body ones are only the promoted constants I think. Does an inline const block preserve padding bytes? I sure hope not.

Conversion between any of these representations requires an interpreter so I really want to give user control.

So I see a number of reasonable behaviors:
1. Evaluate as little as possible: initializer when possible, otherwise ty->valtree, mir->bytes;
2. Same but also uneval bytes to valtrees (loses padding info so unsound, but aeneas would probably want that);
3. Evaluate constants to their nature: ty->valtree, mir->bytes;
4. Evaluate constants to their use: type-level->valtree, body->bytes;
5. Force-evaluate everything to bytes (well, having ConstantExpr::Bytes as a const-eval argument is a dangerous (type-level equality must ignore padding bytes), so maybe not that?);
6. Force-(un)eval everything to valtrees (unsound);
7. :sob:

If I try and take a step back, the rough usecases we know are:
- I'm writing a compiler, I want raw bytes in bodies/allocations and valtrees in types. Don't care about top-level items constants, want accurate allocations. That's option 4.
- I'm writing an interpreter, I want valtrees whenever possible but also soundness. That's a sound version of 6.
[...]

PR #1268, opened 2026-06-05
Add `ConstantExprKind::Call`

N1ark:
Add `ConstantExprKind::Call`

<details><summary>outdated</summary>

Let me know if you have any opinions!

Draft, I still need to:
- clean up a couple transformations around const inlining/simplification
- add const evaluation of VTables (bc why not) 

(I ran out of tokens and I got more important things to do manually)

One thing is that this now has some odd interaction with `--raw-consts`, and it's not clear to me what it is x) 
- possibly a clearer alternative is `--consts={raw,unevaluated,evaluated}`...?
- maybe `--consts={initializer,evaluated}` and `--consts-inline={true,false}` are the two axis? and then for #1168 we can add `--consts=bytes` which doesn't even unevaluate?

</details>

ci: use https://github.com/AeneasVerif/eurydice/pull/423

ci: use https://github.com/AeneasVerif/aeneas/pull/1104

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.211"
+let supported_charon_version = "0.1.212"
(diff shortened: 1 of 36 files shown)


Nadrieril:
I'm a fan of ConstantExprKind::Call, could you split that off to another PR?

I'm realizing the situation is rather complicated, see https://github.com/AeneasVerif/charon/issues/1167#issuecomment-4632702627

Nadrieril (approved):

Issue #1225, opened 2026-06-01
Feature request: represent allocations accurately


Nadrieril:
Today if you write `const { &&1 }`, Charon will happily give you a single const with an initializer like:
```rust
let x = 1;
let y = &x;
return &y
```
which basically require these locals to actually be static variables.

We should add an option where every static allocation that the program might refer to gets its own `static`. This would include string literals.

PR #1269, opened 2026-06-05
Add `--consts=<values|initializers>`

N1ark:
Continuation of #1268 

Add a `--consts` arg, with either `values` for evaluating constants into a value, or `initializers` for preserving the constant's initialiser (default).

One thing is that this now has some odd interaction with `--raw-consts`, and it's not clear to me what it is x) 
- possibly a clearer alternative is `--consts={raw,unevaluated,evaluated}`...?
- maybe `--consts={initializer,evaluated}` and `--consts-inline={true,false}` are the two axis? and then for #1168 we can add `--consts=bytes` which doesn't even unevaluate?

Related to #1167

### DIFF
--- a/charon/src/options.rs
+++ b/charon/src/options.rs
@@ -46,7 +46,7 @@ pub struct CliOpts {
     #[clap(long)]
     #[serde(default)]
     pub skip_borrowck: bool,
-    /// The MIR stage to extract. This is only relevant for the current crate; for dpendencies only
+    /// The MIR stage to extract. This is only relevant for the current crate; for dependencies only
     /// MIR optimized is available.
     #[arg(long)]
     pub mir: Option<MirLevel>,
@@ -225,6 +225,13 @@ pub struct CliOpts {
     #[clap(long)]
     #[serde(default)]
     pub raw_consts: bool,
+    /// How to handle constants and statics: whether they should be represented as a call to their
+    /// initializer function, or whether we should attempt to evaluate them into a value. When
+    /// evaluation isn't possible (e.g. the constant is generic, or for recursive statics), we fall
+    /// back to the initializer call.
+    #[clap(long)]
+    #[serde(default)]
+    pub consts: Option<ConstHandling>,
     /// Replace string literal constants with a constant u8 array that gets unsized,
     /// expliciting the fact a string constant has a hidden reference.
     #[clap(long)]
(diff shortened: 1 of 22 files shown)


charon/src/options.rs
```diff
@@ -225,6 +225,13 @@ pub struct CliOpts {
     #[clap(long)]
     #[serde(default)]
     pub raw_consts: bool,
+    /// How to handle constants and statics: whether they should be represented as a call to their
+    /// initializer function, or whether we should attempt to evaluate them into a value. When
+    /// evaluation isn't possible (e.g. the constant is generic, or for recursive statics), we fall
+    /// back to the initializer call.
+    #[clap(long)]
+    #[serde(default)]
+    pub consts: Option<ConstHandling>,
```
Nadrieril, line 234: Does it work if you just let the default be the default?
```suggestion
    pub consts: ConstHandling,
```

charon/tests/ui/consts/evaluate-consts-inline.rs
```diff
@@ -0,0 +1,21 @@
+//@ charon-args=--consts=values
+
+// A promoted constant: `&42` promotes `42` to an anonymous constant and borrows it.
```
Nadrieril: Promoted constants are those that are automatically promoted by the compiler; that's a different code path than the inline const block you wrote. Can you add some tests with actual promoted constants?

Nadrieril (approved):
Looks quite good to me! I do think `--raw-consts` should be turned into something about inlining consts, but this might also just be about promoted consts rather than other ones so idk

charon/src/options.rs
N1ark: nope :( not sure why

Nadrieril: fair enough, I use an option for `MirLevel` too so maybe it's not meant to work

charon/tests/ui/consts/evaluate-consts-inline.rs
N1ark: done :) though we dont translate those as globals because of #1225

Nadrieril:
There's a diff due to a recent pretty-printing change on `main`, you'll have to rebase

N1ark:
> There's a diff due to a recent pretty-printing change on `main`, you'll have to rebase

oohhhh pretty :O

Issue #865, opened 2025-10-22
Feature request: Translate the bodies of all std functions.


Nadrieril:
Offshoot of https://github.com/AeneasVerif/charon/issues/545. As noted there, we have the following issue: charon translates information from other crates by using the `.rlib` file that cargo provides us for each crates. For the crates we build ourselves we pass `-Zalways-encode-mir` to make sure that the rlib file includes the MIR of all functions. The stdlib rlib file is however pre-built so we can't pass this flag, so some non-generic non-inlineable functions may have no MIR.

Miri is faced with the exact same limitation, and they've solved it for us: running `cargo miri setup -v --print-sysroot` will build all the built-in crates with full MIR information. We can then pass it to charon by adding `--sysroot=the/path/miri/gave/us` to `rustc_args`.

I would like charon to do this automatically. The difficulty here is to do it in a way that's cross-platform and works both inside and outside of Nix. This issue requires some knowledge of Nix. There are three build setups we support:
- Fully without nix, using rustup; that one is easiest;
- Fully within nix, when building the default package and emitting an all-included `charon` binary; for that my preference would be for the sysroot to be already built as part of the nix build so that charon can just reference it (that's the tricky part);
- Nix-based development within a nix shell; we can usually do the same as above, just have to make sure to add the right env vars to the nix shell.
For the non-nix case we should also gracefully fallback to the default sysroot if for some reason `miri` failed to build its own or if miri isn't installed.

To test if the sysroot is correctly set up, you can run
```bash
echo "fn main() {}" > file.rs
cargo build && ./target/debug/charon rustc --start-from=std::process::abort --include=std --print-llbc -- file.rs
```
If the sysroot is good, the translated `abort` function will have a body; otherwise not (that's a good test to add to the ui test suite actually).

N1ark:
yep, though im not very fond of the way they just hard-code `cfg(miri)` into the stdlib, rather than using the pre-existing UB checks nullop/intrinsic which is meant exactly for that i believe :'/

PR #1278, opened 2026-06-09
Add `--sysroot=miri` option

Nadrieril:
This adds the option for running Charon with a full-MIR sysroot. Didn't enable it by default because this made all the `--extract-opaque-bodies` test die x)

cc https://github.com/AeneasVerif/charon/issues/865

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.213"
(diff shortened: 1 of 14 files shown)


N1ark:
> this made all the `--extract-opaque-bodies` test die x)

oh ? what happened :O

Nadrieril:
> > this made all the `--extract-opaque-bodies` test die x)
> 
> oh ? what happened :O

"die" as in "take forever". Together they added a couple extra 100k lines of pretty-printed output x)

N1ark:
> "die" as in "take forever". Together they added a couple extra 100k lines of pretty-printed output x)

i wonder if replacing the `--exclude core::fmt::Formatter` from the test preset with `--exclude core::fmt` helps? there is a lot of formatting code </3

PR #1279, opened 2026-06-09
Use a full-MIR sysroot by default

### LINKED ISSUE #865: Feature request: Translate the bodies of all std functions.

Nadrieril:
Fixes https://github.com/AeneasVerif/charon/issues/865.

@N1ark you were right, a lot of the extra noise was just `core::fmt` x)

ci: use https://github.com/AeneasVerif/eurydice/pull/424

### DIFF
--- a/charon/src/options.rs
+++ b/charon/src/options.rs
@@ -471,7 +471,7 @@ impl CliOpts {
                         .push("-Zcrate-attr=feature(register_tool)".to_owned());
                     self.rustc_args
                         .push("-Zcrate-attr=register_tool(charon)".to_owned());
-                    self.exclude.push("core::fmt::Formatter".to_owned());
+                    self.exclude.push("core::fmt".to_owned());
                 }
             }
         }
--- a/charon/src/options.rs
+++ b/charon/src/options.rs
@@ -61,8 +61,10 @@ pub struct CliOpts {
     #[clap(long, value_delimiter = ',')]
     #[serde(default)]
     pub targets: Vec<String>,
-    /// Sysroot to use for rustc invocations. Use `miri` to ask Charon to prepare a per-target
-    /// sysroot with full MIR for standard library items.
+    /// Sysroot to use for rustc invocations. By default Charon builds a sysroot that has full MIR
+    /// for the standard library. You can pass a custom sysroot to use instead, or pass "default"
+    /// to use the normal distributed sysroot, which lacks MIR bodies for many standard library
+    /// functions.
     #[clap(long)]
     #[serde(default)]
     pub sysroot: Option<String>,
--- a/charon/tests/ui/regressions/issue-1233-foreign-static.out
+++ b/charon/tests/ui/regressions/issue-1233-foreign-static.out
@@ -1,24 +1,4 @@
 # Final ULLBC before serialization:
 
-// Full name: core::fmt::num::DECIMAL_PAIRS
-fn DECIMAL_PAIRS() -> &'static [u8; 200usize]
-{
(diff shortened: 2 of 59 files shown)


charon/tests/ui/regressions/issue-1233-foreign-static.out
```diff
@@ -1,24 +1,4 @@
...
-        storage_live(_2);
-        _2 = &_1;
-        _0 = move _2;
-        return;
-    }
-}
-
-// Full name: core::fmt::num::DECIMAL_PAIRS
-static DECIMAL_PAIRS: &'static [u8; 200usize] = DECIMAL_PAIRS()
-
```
N1ark, line 22: what

charon/src/options.rs
```diff
@@ -61,8 +61,9 @@ pub struct CliOpts {
     #[clap(long, value_delimiter = ',')]
     #[serde(default)]
     pub targets: Vec<String>,
-    /// Sysroot to use for rustc invocations. Use `miri` to ask Charon to prepare a per-target
-    /// sysroot with full MIR for standard library items.
+    /// Sysroot to use for rustc invocations. By default Charon builds a sysroot that has full MIR
+    /// for the standard library. You can pass a custom sysroot to use instead, or pass "default"
+    /// to use the normal distributed sysroot.
```
N1ark: maybe mention this?
```suggestion
    /// Sysroot to use for rustc invocations. By default Charon builds a sysroot that has full MIR
    /// for the standard library. You can pass a custom sysroot to use instead, or pass "default"
    /// to use the normal distributed sysroot, which may miss some items in the standard library.
```

charon/tests/ui/regressions/issue-1233-foreign-static.out
Nadrieril: I changed the default for tests to exclude all of `core::fmt`

N1ark: ohhh thats in core fmt yes

Issue #1235, opened 2026-06-01
Feature request: start from `#[test]` functions


N1ark:
The rust compiler actually does quite a lot of heavy lifting around test functions (`#[test]`), meaning that by the time Charon is able to reach the functions the test attribute is gone and the tested functions aren't easily retrievable. 

It would be nice if Charon had a flag (or automatically detected?) that Cargo is compiling tests, and keeps all the test entry points identifiable (since Cargo will introduce a big `main` function that does all the heavy lifting).

[I did something of the sort in Obol](https://github.com/soteria-tools/obol/blob/1ce4c36928b0ed946c13214c4f416d18abb4380a/src/bin/obol-driver/translate/translate_crate.rs#L428-L448), but there's maybe nicer ways of doing it? An option is also opening a PR in Rust to make test functions preserve a test attribute post-transformation, something similar is done for `cfg` with `cfg_trace`

Nadrieril:
Indeed, I'd rather have rustc preserve the information there.

PR #1282, opened 2026-06-11
Translate some builtin attributes

### LINKED ISSUE #1231: Feature request: translate attributes
Charon used to basically translate all attributes into declarations, however after The Great Attribute Rework™ all attributes are `Parsed`, meaning Charon doesn't translate them anymore. This makes it impossible to e.g. know if an item's `rustc_diagnostic_item`, among other things. Would be nice if we started translating them again!

It would be cute if we had a whole enum for all possible attributes, like what's in rustc, but I don't feel too strongly about it; if you'd prefer converting them to strings and doing it the old way it's cool too

Nadrieril:
Fixes #1231 by adding code to auto-translate some rustc asts to Charon.

@N1ark need anything else than what I included?

ci: use https://github.com/AeneasVerif/eurydice/pull/434


charon/src/bin/generate-asts/generate_rust.rs
N1ark, line 1: this is so stupid i love it

charon/src/ast/attributes.rs
```diff
@@ -0,0 +1,651 @@
...
+///
+/// For more details on attribute parsing, see the [`rustc_attr_parsing`] crate.
+///
+/// [`rustc_parse`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/index.html
+/// [`rustc_codegen_ssa`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/index.html
+/// [`rustc_attr_parsing`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_attr_parsing/index.html
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+#[cfg_attr(feature = "charon_on_charon", charon::rename("BuiltinAttr"))]
+#[cfg_attr(feature = "charon_on_charon", charon::variants_prefix("BuiltinAttr"))]
+pub enum BuiltinAttr {
```
N1ark: hmm  i think it would be nice to also have `Deprecated`, and `ShouldPanic` (which I didn't know was kept!), otherwise i think most things i'd care about are here! ignoring `#[test]` which i mentioned in #1235 which needs a rustc PR </3

charon/src/ast/meta.rs
N1ark: can we replace this with the `LangItem` type?

Nadrieril: yep!

charon/src/ast/attributes.rs
N1ark: i have become aware of [this Miri test](https://github.com/rust-lang/miri/blob/master/tests/pass/static_align.rs), which exercices static alignment... it's an unstable feature so im ok not supporting it, but it would be cute and easy to add, so `rustc_align_static` please (https://github.com/rust-lang/rust/issues/146177) (and `align` while we're at it, so we can handle function alignment too! (https://github.com/rust-lang/rust/issues/82232))

Nadrieril: I think `rustc_align_static` isn't implemented yet ^^

PR #391, opened 2024-10-01
ci: Rebuild kyber lockfile ourselves

Nadrieril:

PR #1284, opened 2026-06-12
Add `reconstruct_ptr_null_checks` pass

giltho:
In Soteria, reasoning about pointer-to-integer cast is expensive. It requires instantiating a fresh variable and adding a bunch of constraints on it that need to be checked.
After a bit of investigation, I realised that a *lot* of these casts are just there to check if a pointer is null. It's done by transmuting to an integer and checking if it's 0. Checking for null pointers in Soteria directly is cheap though, so we'd rather do that.

This PR addresses this need by reconstructing the following pattern into a builtin:
```rs
_x := transmute<*const _, usize>(_y);
switch _x [0usize -> .., _ -> ]
```
or 
```rs
_x := transmute<*const _, usize>(_y);
_z := _x == 0 // or _x != 0
```

The first pattern occurs in optimsied MIR and the second in unoptimised MIR. We unfortunately have to deal with both since a lot of these patterns happen in the stdlib that we obtain as already optimised.

We replace this with:
```rs
_tmp := copy _y != (no-provenance 0);
_x := cast<bool, usize>(move _tmp);
switch _x ....
```

### DIFF
--- a/charon-ml/src/generated/Generated_FullAst.ml
+++ b/charon-ml/src/generated/Generated_FullAst.ml
@@ -177,6 +177,10 @@ and cli_options = {
           information. *)
   reconstruct_asserts : bool;
       (** Replace [if x { panic() }] with [assert(x)]. *)
+  reconstruct_null_checks : bool;
+      (** Replace the [transmute::<*T, usize>(p)] that rustc emits when lowering
+          pointer null-checks (e.g. [<*const T>::is_null]) with a comparison
+          against the null pointer. *)
   unbind_item_vars : bool;
       (** Use [DeBruijnVar::Free] for the variables bound in item signatures,
           instead of [DeBruijnVar::Bound] everywhere. This simplifies the
--- a/charon/src/transform/resugar/reconstruct_ptr_null_checks.rs
+++ b/charon/src/transform/resugar/reconstruct_ptr_null_checks.rs
@@ -1,0 +1,189 @@
+//! # Micro-pass: reconstruct pointer null-checks.
+//!
+//! When rustc lowers a pointer null-check (e.g. `<*const T>::is_null`), it reinterprets the
+//! pointer's bits as an integer and tests that integer against `0`. Depending on the MIR level, the
+//! test takes one of two shapes:
+//! ```text
+//! // optimized MIR: the comparison is a `switch` terminator
+//! _23 := transmute<*mut u8, usize>(copy raw_ptr_6);
+//! switch copy _23 -> [0: bb20, otherwise: bb21]
+//!
+//! // built MIR: the comparison is a `== 0` / `!= 0` statement
+//! _3 := transmute<*const u8, usize>(copy _2);
+//! _0 := move _3 == const 0usize;
+//! ```
+//!
+//! Some tools would like to avoid reasoning about the pointer-to-integer transmutation. Instead we
+//! compare the pointer against the null pointer (a no-provenance pointer with address `0`), so
+//! that the comparison stays in the pointer world:
+//! ```text
+//! _tmp := copy raw_ptr_6 != const (no-provenance 0);
+//! _23 := cast<bool, usize>(move _tmp);
+//! switch copy _23 -> [0: bb20, otherwise: bb21]
+//! ```
+//! The new value assigned to `_23` is `0` if and only if the pointer is null, exactly like the
+//! transmuted address it replaces, so the null-check itself is left untouched. The only
+//! information lost is the address of a non-null pointer (the value becomes `1`), which is sound
+//! because that value feeds nothing but the null-check.
+//!
+//! The transformation is only fired when:
+//! - we transmute a *sized* (thin) pointer to `usize`/`isize`;
+//! - the result feeds either a switch with a single `0` case and an otherwise branch, or a
+//!   `== 0` / `!= 0` comparison against the null address;
+//! - the transmuted value is used nowhere else (necessary for soundness)
+
+use derive_generic_visitor::Visitor;
+use std::ops::ControlFlow::{self, Continue};
+
+use crate::ids::IndexVec;
+use crate::transform::TransformCtx;
+use crate::transform::ctx::UllbcPass;
+use crate::ullbc_ast::*;
+
+/// Count how many times each local appears in the body, ignoring `StorageLive`/`StorageDead`
+/// statements (which mention a local but don't actually use its value).
+#[derive(Visitor)]
+struct LocalUsageCounter {
+    counts: IndexVec<LocalId, usize>,
+}
+
+impl VisitBody for LocalUsageCounter {
+    fn enter_local_id(&mut self, lid: &LocalId) {
+        self.counts[*lid] += 1;
+    }
+    fn visit_ullbc_statement(&mut self, st: &Statement) -> ControlFlow<Self::Break> {
+        match &st.kind {
+            StatementKind::StorageLive(_) | StatementKind::StorageDead(_) => Continue(()),
+            _ => self.visit_inner(st),
+        }
+    }
+}
+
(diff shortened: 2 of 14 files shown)


giltho:
Also ping @N1ark 

Btw, I'm not 100% familiar with what is idiomatic to do in Charon, so maybe I missed some important utilities that would've made my life simpler.

And usual disclosure: this was written with Fable, but I did read the entirety, and rewrite what I was unhappy with

N1ark:
see my comment in https://github.com/soteria-tools/soteria/pull/386#discussion_r3402580009

giltho:
@N1ark Remove your thumbs down now 😠

giltho:
Making it a draft until I can test it still does what I need it to do

charon/tests/ui/ptr-null-checks.out
```diff
@@ -0,0 +1,109 @@
...
+    storage_dead(_3)
+    storage_live(y_4)
+    switch copy x_2 {
+        0usize => {
+            y_4 = const 1usize
+        },
+        _ => {
+            y_4 = const 2usize
+        },
+    }
```
N1ark, line 95: shouldn't this be gone?

charon/tests/ui/ptr-null-checks.rs
```diff
@@ -0,0 +1,34 @@
...
+// Should NOT be rewritten: the transmuted value `x` is used after the switch, so replacing it with
+// the address-erasing null-check would change the observable result.
+pub fn other_use(p: *mut u8) -> usize {
+    let x: usize = unsafe { core::mem::transmute(p) };
+    let y = match x {
+        0 => 1usize,
+        _ => 2,
+    };
+    x.wrapping_add(y)
+}
```
N1ark, line 34: also here; can't you write some piece of rust code that implicitly does the null-ptr check? supposedly that's what we're interested in, rather than in user-code that matches the pattern
i don't know when they occur so you might have to find occurences of it in some existing function, but i suspect something like this does the trick (?)
```rust
let ptr = &11 as *const u32;
let _x = unsafe { *ptr };
```

N1ark: the tooltip of the flag you added mentions `<*const T>::is_null` so... try calling that? you'll have to specify `//@ charon-args=--extract-opaque-bodies` in the test

N1ark: Also from reading the doc of `is_null`:
> Note that unsized types have many possible null pointers, as only the raw data pointer is considered, not their length, vtable, etc. Therefore, two pointers that are null may still not compare equal to each other.

So i believe what you're doing doesn't work when the pointer has metadata. You might need to add a right before to make sure the pointer has its metadata ignored
```
_tmp0 := cast<typeof<_ptr>, *const ()>(_ptr)
```

giltho: https://github.com/AeneasVerif/charon/pull/1284/changes#diff-d5c933971e9c998e2452ccfceccd6af3108d51b37f384b530d448434d77b39faR30

The transformation only applies to sized pointers

N1ark: ah right! nice

charon/src/transform/resugar/reconstruct_ptr_null_checks.rs
```diff
@@ -0,0 +1,189 @@
...
+        for block in b.body.iter_mut() {
+            let mut idx = 0;
+            while idx < block.statements.len() {
+                let Some((dest, src_ty, tgt_lit_ty, operand)) =
+                    match_transmuted_null_check(ctx, &usages, block, &block.statements[idx])
+                        .map(|(place, ty, lit, op)| (place.clone(), ty.clone(), lit, op.clone()))
+                else {
+                    idx += 1;
+                    continue;
+                };
```
Nadrieril, line 160: Please use `FunDecl::transform_ullbc_statements`, it provides a bunch of utilities for editing a body in-place instead of you doing the splicing by hand

Nadrieril:
First, I have two questions:
- Why do you still get optimized MIR now that charon uses a Miri-built (unoptimized) toolchain? Could be that just updating would remove some of these cases;
- A lot of these casts are actually `if ptr == null { unreachable_unchecked() }` introduced by rustc's `CheckNull` pass. I disabled it for all crates recently (except std because std is built with whatever flags Miri gives it). If that's the casts you encounter, I could figure out how to disable that pass when building the stdlib and maybe that pass wouldn't be needed?

Basically, I would expect that if you update and if we fix the stdlib flags, you'd never encounter such a null check except in cases where the user literally wrote `x as usize == 0`. So this pass wouldn't be needed.

Not that I mind this pass, it's cheap. Happy to merge it in the meantime.

charon-ml/src/generated/Generated_FullAst.ml
```diff
@@ -177,6 +177,10 @@ and cli_options = {
           information. *)
   reconstruct_asserts : bool;
       (** Replace [if x { panic() }] with [assert(x)]. *)
+  reconstruct_null_checks : bool;
+      (** Replace the [transmute::<*T, usize>(p)] that rustc emits when lowering
+          pointer null-checks (e.g. [<*const T>::is_null]) with a comparison
+          against the null pointer. *)
```
Nadrieril, line 183: I believe that's incorrect today: we never get optimized MIR anymore, so these null pointer checks should only be the ones inserted by the `CheckNull` rustc pass whenever a raw pointer gets dereferenced.

giltho:
> Why do you still get optimized MIR now that charon uses a Miri-built

Because the default for Soteria when not in polymorphic mode is still Obol (though I believe we're excited to use Charon only and it may happen soon-ish?)

The pipeline is MIR -OBOL> ULLBC -Charon Passes> ULLBC 

However, I just realised: I originally made this PR in Charon because I was modifying the AST. Now that the PR requires no change in AST, if this pattern doesn't exist in Charon anymore, we can possibly move it to Obol? @N1ark

charon/src/transform/resugar/reconstruct_ptr_null_checks.rs
giltho: AH, this is exactly what I was looking for 🙃 Will rewrite this if we decide to merge it in here

Nadrieril: It's extremely not documented anywhere, sowwy :upside_down_face:

N1ark:
> > Why do you still get optimized MIR now that charon uses a Miri-built
> 
> Because the default for Soteria when not in polymorphic mode is still Obol

Yes, we don't use Charon for most of our code. But most importantly until #391 we simply didn't use a sysroot by default :P We wouldn't be able to safely use Charon's sysroot anyways because we first compile some libraries we link into the rustc call with `-L`, and those libs need to be compiled with the same sysroot, so we can't let Charon pick it.


> Now that the PR requires no change in AST, if this pattern doesn't exist in Charon anymore, we can possibly move it to Obol? @N1ark

yes :)

Nadrieril:
> We wouldn't be able to safely use Charon's sysroot anyways because we first compile some libraries we link into the rustc call with `-L`, and those libs need to be compiled with the same sysroot, so we can't let Charon pick it.

I could give you a `charon print-sysroot` so you can use the same one

Nadrieril:
Actually you can use charon itself to build the rlibs, I do it in my test suite: `CHARON_EMIT_ARTIFACTS=1 charon rustc --no-serialize -- --crate-type=rlib` will compile the crate with the right sysroot and flags, and emit an rlib.

giltho:
Ok, so the conclusion is:
- This PR isn't modifying the AST
- It is useless in Charon's pipeline since the pattern doesn't exist

I'm closing this PR in favour of https://github.com/soteria-tools/obol/pull/3

PR #1290, opened 2026-06-18
Add variadics flag to `FunSig`

N1ark:
Add a `is_variadic` flag to `FunSig`, for extern C functions 

ci: use https://github.com/AeneasVerif/aeneas/pull/1154

### DIFF
--- a/charon/tests/ui/variadics.rs
+++ b/charon/tests/ui/variadics.rs
@@ -1,0 +1,18 @@
+#![feature(c_variadic)]
+
+pub unsafe extern "C" fn sum(mut ap: ...) -> i32 {
+    let a: i32 = ap.next_arg();
+    let b: i32 = ap.next_arg();
+    a + b
+}
+
+extern "C" {
+    fn foo(x: u32, y: u64, rest: ...);
+}
+
+pub fn call() {
+    let _ = unsafe { sum(1i32, 2i32) };
+    let _ = unsafe { foo(1, 2, 3, 4, 5) };
+    let fn_ptr: unsafe extern "C" fn(u32, u64, ...) = foo;
+    let _ = unsafe { fn_ptr(1, 2, 3, 4, 5) };
+}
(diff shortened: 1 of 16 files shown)


charon-ml/tests/Test_Deserialize.ml
```diff
@@ -105,11 +105,12 @@ let assert_pretty_matches_rust (file : string) (printed : string) : unit =
...
   in
   let ocaml = normalize_trailing_newlines printed in
-  if expected <> ocaml then
+  if expected <> ocaml then (
     let diff = unified_diff ~expected ~ocaml in
+    Format.print_string diff;
     Format.kasprintf failwith
       "OCaml pretty-printer output differs from %s pretty-print for %s:\n%s"
-      charon_bin file diff
+      charon_bin file diff)
```
Nadrieril: Doesn't that print the diff twice now?

charon/tests/ui/variadics.rs
```diff
@@ -0,0 +1,20 @@
+#![feature(c_variadic)]
+
+// A C-variadic function: its signature should be flagged as variadic in the ULLBC
+// (rendered as a trailing `, ...` in the printed signature).
```
Nadrieril: Please avoid letting the llms indulge in their tendencies to add useless comments
```suggestion
```

charon-ml/tests/Test_Deserialize.ml
N1ark: oops that sneaked past
yes, the issue is that the `failwith` printer doesn't actually print newlines but instead escapes them, so when the output differs it's very hard to read :') that was just a temporary fix for me to figure out what changed

charon/tests/ui/variadics.rs
N1ark: oh oops yes

charon-ml/tests/Test_Deserialize.ml
Nadrieril: the diff is meant to be legible, so let's print only the useful version x)

PR #1291, opened 2026-06-18
Fix `reconstruct_fallible_operations`

N1ark:
The reconstruct_fallible_operations pass didn't look outside its block for uses, which led to code that wouldn't type check (either with `.0` access to an integer variable, or with `int -> (int, bool)` assignments.

The fix i did was to expand the transformation to work across blocks, but another solution would be to simply not do the pass if we know the local is used outside the block, since it's questionable whether it makes sense to do this pass when the user explicitly calls `add_with_overflow`

Either solutions are fine with me I just want well-typed code :D

### DIFF
--- a/charon/tests/ui/regressions/reconstruct-fallible.rs
+++ b/charon/tests/ui/regressions/reconstruct-fallible.rs
@@ -1,0 +1,41 @@
+//@ charon-args=--sysroot default --monomorphize --mir elaborated
+
+#![feature(core_intrinsics)]
+
+fn other() {}
+fn sink_tuple(_x: (usize, bool)) {}
+fn sink_int(_x: usize) {}
+fn sink_bool(_x: bool) {}
+
+fn must_preserve_tuple() {
+    let z = std::intrinsics::add_with_overflow(0usize, 0usize);
+    other();
+    sink_tuple(z);
+}
+
+fn can_erase_check() {
+    let z = std::intrinsics::add_with_overflow(0usize, 0usize);
+    other();
+    sink_int(z.0);
+}
+
+fn must_preserve_overflow_flag() {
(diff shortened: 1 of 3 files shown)


Nadrieril:
> another solution would be to simply not do the pass if we know the local is used outside the block, since it's questionable whether it makes sense to do this pass when the user explicitly calls add_with_overflow

I agree, I'd like that please! I only want to catch code generated by MIR lowering.

charon/tests/ui/regressions/reconstruct-fallible.rs
```diff
@@ -0,0 +1,61 @@
+//@ charon-args=--sysroot default --monomorphize --mir elaborated
```
Nadrieril, line 1: I mean this is ok but overpowered just to catch a bug in a pass. please at least remove the `inline(never)` (we don't use mir optimizations in this test) and remove noise from the comments

N1ark:
> > another solution would be to simply not do the pass if we know the local is used outside the block, since it's questionable whether it makes sense to do this pass when the user explicitly calls add_with_overflow
> 
> I agree, I'd like that please! I only want to catch code generated by MIR lowering.

done :) it's sooo much cleaner i much prefer that too actually ^^ note i didnt add the `!used_outside_block` to all other pattern matches in that pass, because all of the others rely on the presence of a `StatementKind::Assert`, so I think we can be much more confident that they are legit MIR lowering code, rather than user code being lowered funny

charon/tests/ui/regressions/reconstruct-fallible.rs
```diff
@@ -0,0 +1,41 @@
...
+
```
Nadrieril:
```suggestion
// Check that the reconstruct_fallible_operations pass correctly handles check_add usage
// that spans multiple basic blocks.
```

PR #1292, opened 2026-06-22
Fix/miri sysroot portability

maximebuyse:
This PR fixes two different issues affecting aeneas nix builds:

## Missing `/build` directory on github runners

The `full-mir-sysroot.nix` build script introduced by https://github.com/AeneasVerif/charon/pull/1278 uses `/build` which exists on self-hosted runners but not on github runners which results in failing [builds](https://github.com/AeneasVerif/aeneas/actions/runs/28148842408/job/83361757659) for aeneas.
This PR changes this nix script to use a temporary directory instead.

## `charon-unwrapped` missing `miriSysroots`

Also following https://github.com/AeneasVerif/charon/pull/1278, charon needs the `CHARON_MIRI_SYSROOTS` which is correctly passed to the `charon` nix package, but not to `charon-unwrapped`. This one affects [linux-x86_64 aeneas builds](https://github.com/AeneasVerif/aeneas/actions/runs/28148842408/job/83361757660) which avoid the previous one because they run on self-hosted runners.
This PR fixes that, and also adds a CI job that build `charon-unwrapped` to avoid future regressions for this package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
       #- uses: cachix/install-nix-action@v22
       - uses: actions/checkout@v4
       - run: nix build -L .#charon
+      - run: nix build -L .#charon-portable
       - run: nix build -L .#charon-ml
       - run: nix flake check -L
 
--- a/flake.nix
+++ b/flake.nix
@@ -21,19 +21,40 @@
           inherit system;
           overlays = [ (import rust-overlay) ];
         };
+        inherit (pkgs) lib stdenv makeWrapper bintools;
 
         rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain;
         craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
         ocamlformat = pkgs.ocamlPackages.ocamlformat_0_27_0;
 
         fullMirSysroots = pkgs.callPackage ./nix/full-mir-sysroots.nix { inherit rustToolchain; };
-        charon = pkgs.callPackage ./nix/charon.nix {
-          inherit craneLib rustToolchain;
-          miriSysroots = fullMirSysroots;
-        };
+        charon = pkgs.runCommand "charon"
+          {
+            nativeBuildInputs = [ makeWrapper ]
+              # For `install_name_tool`.
+              ++ lib.optionals stdenv.isDarwin [ bintools ];
+            passthru = charon-unwrapped.passthru;
+          }
+          (''
+            cp -r ${charon-unwrapped} $out
+            chmod -R u+w $out
+
+            # Make sure the toolchain is in $PATH so that `cargo` can work
+            # properly. On mac we also have to tell `charon-driver` where to
+            # find the rustc_driver dynamic library; this is done automatically
+            # on linux.
+            wrapProgram $out/bin/charon \
+              --set CHARON_TOOLCHAIN_IS_IN_PATH 1 \
+              --set CHARON_MIRI_SYSROOTS "${fullMirSysroots}" \
+              --prefix LD_LIBRARY_PATH : "${lib.makeLibraryPath [ rustToolchain ]}" \
+              --prefix PATH : "${lib.makeBinPath [ rustToolchain ]}"
+          ''
+          + (lib.optionalString stdenv.isDarwin ''
+            # Ensures `charon-driver` finds the dylibs correctly.
+            install_name_tool -add_rpath "${rustToolchain}/lib" "$out/bin/charon-driver"
+          ''));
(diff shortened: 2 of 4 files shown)


nix/full-mir-sysroots.nix
```diff
@@ -16,15 +16,27 @@ runCommand "charon-full-mir-sysroots"
...
+    # Build under a fixed path rather than the nix build dir. `cargo miri setup`
+    # compiles std's dependencies and bakes their $CARGO_HOME source paths into
+    # the sysroot's rlibs/rmeta, so the output is only reproducible if it is
+    # built at a constant path. The nix build dir is *not* constant: it is
+    # /build under a sandbox but a temp dir otherwise, which would make this
+    # fixed-output derivation's hash depend on whether the build was sandboxed
+    # (and the old hardcoded /build simply does not exist on a sandbox-less
+    # builder, so `mkdir -p /build/...` failed there with EACCES). /tmp is
+    # writable both inside the nix sandbox and on a sandbox-less builder, so a
+    # fixed /tmp path builds in both and yields an identical hash.
```
Nadrieril: Please write this comment by hand so that it contains only the information that will be useful to someone coming to understand this code later

flake.nix
```diff
@@ -31,7 +31,15 @@
...
           miriSysroots = fullMirSysroots;
         };
-        charon-unwrapped = pkgs.callPackage ./nix/charon.nix { inherit craneLib rustToolchain; enableWrapping = false; };
+        charon-unwrapped = pkgs.callPackage ./nix/charon.nix {
+          inherit craneLib rustToolchain;
+          enableWrapping = false;
+          # Pass the sysroot like `charon` above. Without it the test suite (run
+          # as the check phase) makes charon fall back to `cargo miri setup` at
+          # build time, which fails in a network-isolated sandbox with EACCES
+          # ("Permission denied (os error 13)") and breaks `charon-portable`.
```
Nadrieril: Same here

Nadrieril (changes requested):
Hi, could you rewrite the PR descirption in your own words? I have little patience for LLM verbosity

maximebuyse:
> Hi, could you rewrite the PR description in your own words? I have little patience for LLM verbosity

@Nadrieril Sorry about the LLM noise. I rewrote the PR description and removed the verbose comments from all commits.

nix/full-mir-sysroots.nix
```diff
@@ -16,15 +16,17 @@ runCommand "charon-full-mir-sysroots"
...
   # Rust metadata records rust-src paths from rustToolchain; charon supplies that toolchain
   # separately.
   unsafeDiscardReferences.out = true;
 }
   ''
-    export HOME="/build/home"
-    export CARGO_HOME="/build/cargo"
-    export TMPDIR="/build/tmp"
+    builddir=/tmp/charon-full-mir-sysroots-build
+    rm -rf "$builddir"
```
Nadrieril:
```suggestion
```

shouldn't be needed right, we're in a nix sandbox. and if it was needed, that would mean concurrent builds would be botched that way.

.github/workflows/ci.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       #- uses: cachix/install-nix-action@v22
       - uses: actions/checkout@v4
       - run: nix build -L .#charon
+      - run: nix build -L .#charon-portable
```
Nadrieril, line 29: I'm not a fan of doubling my CI time. Instead, could you move the wrapping stuff out of charon.nix, then build `charon` by doing just the wrapping on top of `charon-unwrapped`?

maximebuyse: Makes sense, I pushed a new version.

flake.nix
```diff
@@ -21,19 +21,40 @@
...
+          + (lib.optionalString stdenv.isDarwin ''
+            # Ensures `charon-driver` finds the dylibs correctly.
+            install_name_tool -add_rpath "${rustToolchain}/lib" "$out/bin/charon-driver"
+          ''));
         charon-unwrapped = pkgs.callPackage ./nix/charon.nix {
-          inherit craneLib rustToolchain;
-          enableWrapping = false;
+          inherit craneLib;
           miriSysroots = fullMirSysroots;
         };
```
Nadrieril: please move the charon-unwrapped definition above charon so we can follow where it comes from

flake.nix
Nadrieril: Could you do this comparison in nix instead of bash, using stdenv.isLinux?

Nadrieril: Just like the isDarwin check above, essentially

Nadrieril (approved):
LGTM, thank you for taking care of this!

Nadrieril:
I think the CI is down due to the heat x')

Nadrieril:
Could you rebase onto main to get the CI fixes?

PR #1293, opened 2026-06-23
Nicer char pretty-printing

N1ark:
We used to print chars as literally their value, which is hard to parse for non-ascii characters and quite unclear (e.g. `const X` could refer to the character `'X'`, or to the global `X`)

changed it so characters are printed like they would be written in Rust


Nadrieril:
Oh yeah that's much better!

PR #1310, opened 2026-07-01
Include cacert in the native inputs to fix build on darwin

### LINKED ISSUE #1301: Bug: build fails on macOS with SIP enabled
Running the flake on macOS with System Integrity Protection (SIP) enabled results in:
```
> nix run github:aeneasverif/charon
error: builder for '/nix/store/l4sfhwdbbh2p5996x01mpxyib60rsmnr-charon-full-mir-sysroots.drv' failed with exit code 1;
       last 4 log lines:
       > structuredAttrs is enabled
       > mkdir: cannot create directory '/build': Read-only file system
       > mkdir: cannot create directory '/build': Read-only file system
       > mkdir: cannot create directory '/build': Read-only file system
       For full logs, run:
         nix log /nix/store/l4sfhwdbbh2p5996x01mpxyib60rsmnr-charon-full-mir-sysroots.drv
error: 1 dependencies of derivation '/nix/store/bd400lsbc308gif4a1i0yl0vxd3pssqs-charon-0.1.217.drv' failed to build
```

I actually encountered this when running the Aeneas flake, as indicated in Aeneas' README. I tried this both on modern macOS ARM64 and an ancient version on x86. Apparently disabling SIP would make it possible to create `/build`.

A solution would be to use `$NIX_BUILD_TOP` instead of `/build` in `nix/full-mir-sysroot.nix`, but it seems that this affects the hash of the final binaries on macOS.

manmatteo:
Fixes https://github.com/AeneasVerif/charon/issues/1301

### DIFF
--- a/nix/full-mir-sysroots.nix
+++ b/nix/full-mir-sysroots.nix
@@ -17,11 +17,11 @@ runCommand "charon-full-mir-sysroots"
   nativeBuildInputs = [ rustToolchain cacert ];
   outputHashMode = "recursive";
   outputHashAlgo = "sha256";
-  outputHash =
-    if stdenv.isDarwin then
-      "sha256-UyX+j+F5peb9NUqgkyvb6JItp6xHoAJbpFMoQ1K7MIM="
-    else
-      "sha256-sbmBXFEOx2v3gIg+yRYpijd2MBouoeD9oFyfEDFGjcs=";
+  outputHash = {
+    x86_64-linux = "sha256-sbmBXFEOx2v3gIg+yRYpijd2MBouoeD9oFyfEDFGjcs=";
+    aarch64-darwin = "sha256-UyX+j+F5peb9NUqgkyvb6JItp6xHoAJbpFMoQ1K7MIM=";
+    x86_64-darwin = "sha256-IXD96odt+Ss6Fj+oPYeUvwWsbIxS1JUHYLgqBr2N2LU=";
+  }.${stdenv.hostPlatform.system};
   # Rust metadata records rust-src paths from rustToolchain; charon supplies that toolchain
   # separately.
   unsafeDiscardReferences.out = true;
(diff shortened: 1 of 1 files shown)


manmatteo:
Apparently the hash only changes on darwin, so two different hashes would need to be specified?

Nadrieril:
Yeah, I expected something like that. You can branch on `stdenv.isDarwin` to give the two different hashes

.direnv/bin/nix-direnv-reload
Nadrieril, line 1: Could you remove this file from the commit?

manmatteo: My bad, sorry

nix/full-mir-sysroots.nix
```diff
@@ -13,10 +14,14 @@ in
...
-  nativeBuildInputs = [ rustToolchain ];
+  nativeBuildInputs = [ rustToolchain cacert ];
   outputHashMode = "recursive";
   outputHashAlgo = "sha256";
-  outputHash = "sha256-sbmBXFEOx2v3gIg+yRYpijd2MBouoeD9oFyfEDFGjcs=";
+  outputHash =
+    if stdenv.isDarwin then
+      "sha256-UyX+j+F5peb9NUqgkyvb6JItp6xHoAJbpFMoQ1K7MIM="
+    else
+      "sha256-UyX+j+F5peb9NUqgkyvb6JItp6xHoAJbpFMoQ1K7MIM=";
```
Nadrieril: Unsurprisingly the linux hash also changed by adding a dependency, let's hope this works:
```suggestion
      "sha256-sbmBXFEOx2v3gIg+yRYpijd2MBouoeD9oFyfEDFGjcs=";
```

Nadrieril: oh wait no the hash is the same, it's just that you copy-pasted the macos one twice ^^

PR #1316, opened 2026-07-02
Insert `StorageDead`s for all locals

Nadrieril:
May as well be entirely consistent: every non-return local deserves a storage_dead, and every non-argument local deserves a storage_live.

ci: use https://github.com/AeneasVerif/eurydice/pull/432

ci: use https://github.com/AeneasVerif/aeneas/pull/1201


N1ark:
i wonder if this breaks "in place" argument passing ([see MiniRust's ArgumentExpr](https://github.com/minirust/minirust/blob/master/spec/lang/syntax.md#statements-terminators))... the thing is I actually don't know what it is/how it happens/if it even exists in MIR (I don't believe it does?) so its probably ok 
the const eval has [`FnArg::InPlace`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/interpret/enum.FnArg.html#variant.InPlace) but it seems to be an optimisation and unrelated (?)

Nadrieril:
Doesn't a function drop all its arguments anyway? That doesn't include StorageDead?

Nadrieril:
Interesting tho, I could be convinced. And similarly, in-place initialization requires that we don't StorageLive the return place ourselves

N1ark:
> Doesn't a function drop all its arguments anyway? That doesn't include StorageDead?

i think so :P that's why im not sure, supposedly you wouldnt drop arguments you receive in-place...
for now i think this is nice for consistency and i dont think i have any strong objections

Nadrieril:
Too late, you've convinced me that input/output place allocation is handled by the caller :3

Nadrieril:
I keep changing my mind x) For at least the purpose of borrow-checking, arguments are considered deallocated by the end of the function. I preserving caller argument places is at least a valid optimization, but for the MiniRust stuff we would need to make it the actual behavior...

Maybe I should write the MiniRust translator, that would let me know what feels more natural.

Nadrieril:
(waiting on https://github.com/AeneasVerif/eurydice/pull/426 to avoid similar large diffs in the future)

N1ark:
this pr caused a massive perf regression... ill open a PR to make this addition optional

```rust
let mut n = 0;
let x = &mut n;
*x += 1;
// ↑ 10000x times this 
```

for code like this the behaviour is now quadratic, which completely explodes (20GB RSS), this is bc every Nth increment now adds N storage dead markers, leading to 50M new statements for the 10000 increments.

PR #1324, opened 2026-07-14
ci: build release for older glibc versions

abentkamp:
This PR uses polyfill-glibc to make the charon binaries compatible with older glibc versions.

Disclaimer: Mostly coded by Claude. I am not a Nix expert.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,13 +93,6 @@ jobs:
           rm -f "../${{ matrix.artifact_name }}.tar.gz"
           tar -czf "../${{ matrix.artifact_name }}.tar.gz" -- *
 
-      # On Linux, check that the release binaries are compatible with older glibc versions
-      - name: Check glibc version
-        if: ${{ startsWith(matrix.artifact_name, 'charon-linux') }}
-        run: |
-          set -eo pipefail
-          nix build -L .#charon-release-glibc-check
-
       # Verify the binary is functional and still links to `rustc_driver.so` on Linux.
       - name: Smoke Test
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -93,6 +93,13 @@ jobs:
           rm -f "../${{ matrix.artifact_name }}.tar.gz"
           tar -czf "../${{ matrix.artifact_name }}.tar.gz" -- *
 
+      # On Linux, check that the release binaries are compatible with older glibc versions
```
Nadrieril: Would prefer to include this as part of the smoke test, and have it use what we just extracted to avoid the risk of the nix test using a different binary than the one in the tarball

Nadrieril: Alternatively, run this check as part of the charon-release build; it's standard in nix to have tests just after builds

flake.nix
```diff
@@ -74,12 +79,40 @@
...
-        '';
+        ''
+        # Lower the glibc version the binaries require, so the release runs on
+        # any host with glibc >= ${releaseGlibcVersion} regardless of the
+        # (newer) glibc it was built against.
+        + lib.optionalString stdenv.isLinux ''
+          chmod +w charon charon-driver
+          for f in charon charon-driver; do
+            ${polyfill-glibc}/bin/polyfill-glibc \
+              --clear-symbol-version=pidfd_getpid,pidfd_spawnp \
```
Nadrieril, line 105: How did you come to this line? Is this dependent on the Rust code in some way? Or on the glibc versions involved? Why isn't `--target-glibc` enough? (Would like to have this answered in comments)

Nadrieril (approved):

.github/workflows/release.yml
abentkamp: I've made it part of the charon-release build. For the other approach, I wasn't sure where to put the target glibc version.

flake.nix
abentkamp: I've added an explanation. (I had Claude explain it to me, but I wrote the explanation myself)

PR #1325, opened 2026-07-15
Layout guarantee information draft

firefighterduck:
This PR is a first draft of how Charon could support (partially) symbolic layout guarantee information. To make the draft more self-contained and easier to review, all relevant parts are in the new file `charon/src/ast/layout_guarantees.rs`. Additional changes required for this PR can be found in other files, some of which should and probably will be extracted into their own PRs.

## About layout guarantees
Layout guarantees are (partially symbolic) expressions that describe what Rust guarantees about the layout of a type. I've tried to always refer to the reference in the code to keep everything in sync. In general, these guarantees can be computed for polymorphic types and are target-agnostic, but can be concretized/instantiated for specific type arguments and targets if needed (currently only implemented in Rust, not in OCaml).

The draft is insofar complete as it can compute layout guarantees for most types (an open question are DSTs). Further open points for future discussion are the following:
* Can we compute symbolic layout guarantees about more type properties, e.g. whether a type is uninhabited (based on some symbolic predicates)? Optimally, we'd be able to represent all fields of `SizedTypeProperties`, but symbolically.
* ~~[One ui test](tests/ui/associated_types/lift-assoc-ty-in-alias.rs) currently fails, but I was not able to figure out what exactly goes wrong and how to fix it.~~ The problem now gets circumvented.
* Does this already satisfy all requirements by all stakeholders (compare with Zulip thread where we collected them)?


charon/src/ast/types.rs
```diff
@@ -525,7 +525,20 @@ pub type ByteCount = u64;
...
 /// Maps fields to their offset within the layout.
-#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, Drive, DriveMut)]
+#[derive(
+    Debug,
+    Default,
+    Clone,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize,
```
Nadrieril: Please remove these two whenever there's also `De/SerializeState`

charon/src/ast/layout_guarantees.rs
```diff
@@ -0,0 +1,618 @@
...
+use crate::ast::{
+    BuiltinTy, ByteCount, ConstantExpr, ConstantExprKind, FloatTy, HashConsSerializerState, IntTy,
+    Literal, LiteralTy, ScalarValue, SeqHashMap, TargetTriple, TranslatedCrate, Ty, TyKind,
+    TyVisitable, TypeDeclRef, TypeId, UIntTy,
+};
+
+/// The basic building blocks of symbolic layout information.
+#[derive(Debug, Clone, PartialEq, Eq, SerializeState, DeserializeState, Drive, DriveMut)]
+#[serde_state(state_implements = HashConsSerializerState)]
+pub enum LayoutGuaranteeAtom {
```
Nadrieril: Does this always represent a symbolic number of bits?

charon/src/ast/krate.rs
```diff
@@ -194,13 +192,33 @@ pub type DeclarationsGroups = Vec<DeclarationGroup>;
 /// A target triple, e.g. `x86_64-unknown-linux-gnu`.
 pub type TargetTriple = String;
 
+/// Stores the exact alignments for most primitive types.
+#[derive(Clone, Drive, DriveMut, Serialize, Deserialize)]
+pub struct TargetAlignments {
+    pub i1_align: types::ByteCount,
```
Nadrieril: What's the `i1` for?

charon/src/ast/layout_guarantees.rs
firefighterduck: It represents a number of bytes, but yes, its always a symbolic of concrete number of bytes. For some types, knowing the target suffices to know the number of bytes, but since the type could also be a DST, it might still be an indeterminable number.

charon/src/ast/krate.rs
firefighterduck: That's a good question. Rustc has it, LLVM has it, and so I simply kept it. I currently use it for the alignment of booleans, bu I don't know whether that's correct.

charon/src/ast/layout_guarantees.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Default)]
+/// A structure that computes and stores originally symbolic layouts, which have been
+/// concretized as much as possible. Will not be used during translation.
+pub struct Concretizer {
+    computed_layouts: SeqHashMap<Ty, LayoutGuarantees>,
+    curr_requested: HashSet<Ty>,
+}
```
Nadrieril, line 747: What's "concretized"? Is this just "normalized"? The API I'd imagine for that would be:
```rust
pub struct LayoutComputer<'a> {
  krate: &'a TranslatedCrate,
  target: TargetTriple,
  cache: HashMap<Ty, LayoutGuarantee>,
  /// Stack to bail on cycles in the computation.
  stack: Vec<Ty>,
}
impl LayoutComputer<'a> {
    pub fn new(krate: &'a TranslatedCrate, target: Option<TargetTriple>) -> Self { ... }
    /// Computes the most precise layout guarantees we can deduce for this type.
    pub fn compute_layout(&self, ty: Ty) -> LayoutGuarantees { ... }
}
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -495,6 +494,147 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+        } else {
+            // If there is no non-1-ZST field, the type is equivalent to unit.
+            Some(LayoutGuarantees::ONE_ZST)
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    /// Constructs the layout guarantees, even if we failed to obtain a concrete layout for the type.
+    /// Also memoizes all guarantees of types that were required to build the guarantees for the given def.
+    pub fn construct_layout_guarantees(
```
Nadrieril: I'd prefer to move this as much as possible to the layout_guarantees.rs since this doesn't call into rustc internals

charon/src/ast/layout_guarantees.rs
firefighterduck: I'm open to bikeshedding the name. For me, it concretizes the layout guarantees, since it inserts concrete values where possible. Normalizing for me could also mean just rearranging the purely symbolic expressions.

Your API looks nice and I'll use it as the basis for further iterations.

charon/src/bin/charon-driver/translate/translate_types.rs
firefighterduck: I agree. An earlier version was based on the hax representation and for that it made more sense to be there, but now it should live somewhere else.

charon/src/ast/krate.rs
Nadrieril: That isn't correct, booleans are aligned like `u8`. Looking through rustc, it's indeed used to align 1-bit integers, I just have no idea how one may come about

charon/src/ast/layout_guarantees.rs
```diff
@@ -0,0 +1,618 @@
...
+            LiteralTy::Bool => target_info.primitive_alignments.i8_align,
+            LiteralTy::Char => target_info.primitive_alignments.i32_align, // FIXME: the target information in rustc doesn't mention chars?
+        };
+        Some(Self::Concrete(align))
+    }
+}
+
+/// Composite symbolic layout expressions.
+///
+/// `non_exhaustive` since we might need many more composite layouts.
```
Nadrieril, line 90: You didn't actually make it `non_exaustive`

charon/src/ast/krate.rs
Nadrieril: Let's remove it until we have a use for it

charon/src/ast/krate.rs
```diff
@@ -194,13 +192,33 @@ pub type DeclarationsGroups = Vec<DeclarationGroup>;
...
+    pub i16_align: types::ConcreteByteCount,
+    pub i32_align: types::ConcreteByteCount,
+    pub i64_align: types::ConcreteByteCount,
+    pub i128_align: types::ConcreteByteCount,
+    pub f16_align: types::ConcreteByteCount,
+    pub f32_align: types::ConcreteByteCount,
+    pub f64_align: types::ConcreteByteCount,
+    pub f128_align: types::ConcreteByteCount,
+    pub ptr_align: types::ConcreteByteCount,
+}
```
N1ark, line 209: why not just make this a map from `LiteralType` to `ConcreteByteCount`? seems more exhaustive

charon/src/ast/types.rs
firefighterduck: Funny thing, I believe at least `ItemId` requires both. It needs to be stateful to work in `TranslatedCrate::item_names`, but needs to be stateless for use in `DeclarationGroup`.

charon/src/ast/krate.rs
firefighterduck: That would certainly be an option, if we can make it a cheap map. The current version is really just a direct translation of what rustc has.

Nadrieril: I agree it seems more usable as a map (just use a `SeqHashMap`, we're not optimizing for speed in our representation)

charon/src/ast/layout_guarantees.rs
Nadrieril: > Normalizing for me could also mean just rearranging the purely symbolic expressions.

Well yeah, that's what it has to do in cases where fully concrete cannot be obtained, e.g. for any `repr(Rust)` layout

charon/src/ast/types.rs
Nadrieril: Ah, this is hitting a limitation of `serde_state` I should fix :/ It's just that it doesn't support mixing `stateless` and using a custom serialization function.

firefighterduck:
To make things a bit cleaner, I've added a new draft on a [new branch](https://github.com/AeneasVerif/charon/compare/main...firefighterduck:charon:new_layout_guarantees). There are still a few small things missing, but when it's done, I'll see whether I can switch branches or open a new PR.

firefighterduck:
Superseded by #1343

PR #1328, opened 2026-07-16
Promote shifts across blocks in reconstruct_fallible_operations

### LINKED ISSUE #1041: Bug: `reconstruct_fallible_operations` fails to promote shifts assigned through dereferenced pointers
It appears that `reconstruct_fallible_operations` fails to promote a shift from `OWrap` to `OPanic` when the shift result is assigned through a dereferenced mutable pointer (e.g. `(*ptr) = x >> n`, as produced by `IndexMut`). The bounds-check assert is correctly emitted, but the shift itself stays `wrap.>>` instead of being promoted to `panic.>>`.

Problem discovered using Aeneas to extract to Lean and the problem arrived to the Lean code.

## MWE

```rust
use std::ops::IndexMut;

pub struct Wrap(pub [u64; 2]);

impl IndexMut<usize> for Wrap {
    fn index_mut(&mut self, i: usize) -> &mut u64 { &mut self.0[i] }
}
impl std::ops::Index<usize> for Wrap {
    type Output = u64;
    fn index(&self, i: usize) -> &u64 { &self.0[i] }
}

pub fn shr_into_index(x: u64, out: &mut Wrap) {
    out[0] = x >> 20;
}
```

In the LLBC, the bounds check is emitted but the shift stays `wrap.>>`:

```
_4 = cast<i32, u32>(const 20 : i32)
_5 = move _4 < const 64 : u32
assert(move _5 == true) (overflow) else panic
...
(*_6) = move _3 wrap.>> const 20 : i32
```

## Context

This was found via [curve25519-dalek's `from_bytes_wide`](https://github.com/dalek-cryptography/curve25519-dalek/blob/20101184a301471bdf0b6104a1894b3b219c645a/curve25519-dalek/src/backend/serial/u64/scalar.rs#L110), which has many shifts of the form `((words[i] >> N) | (words[j] << M)) & mask`. All of these are correctly promoted to `panic.>>` / `panic.<<` because the shift result is assigned to a local variable. However, the final line:

```rust
hi[4] = words[7] >> 20;
```

assigns the shift result directly through `IndexMut`, so the shift target is a dereferenced pointer `(*_6)` rather than a local. The reconstruction pass pattern match (in `reconstruct_fallible_operations.rs`) does not recognise this case.

## Fix?

Could the pattern match in `reconstruct_fallible_operations` that looks for shift operations in the remaining statements after a bounds check should also match shifts whose result is assigned through a dereferenced pointer, not only to local variables?

**Charon version**: `a9fb1c8f0bd8f1dc6bff295926ec0f708b2296d6`.

MavenRain:
Fixes #1041.

Rustc evaluates a shift's operands, and emits its overflow check, before computing the destination place.  When that computation needs a call, as with the overloaded `index_mut` in the issue, the call ends the block and the shift lands in the block the call returns to.  `reconstruct_fallible_operations` only matches within one block, so it never pairs the check with the shift and the shift stays wrapping.

We now record these checks during the per-block pass and resolve them afterwards over the whole body.  For each one we find the wrapping shift by its amount, drop the check, and compute the shift early into a fresh local where the check used to be, promoted to panic on overflow.  A move then carries the result to the destination back at the original site.

Doing the computation there is sound.  By the time control reaches the check, rustc has already evaluated both operands, and the check itself already reads them and panics on overflow, so the early assignment introduces no new effect and the panic stays in its original position before the call.

The new UI test covers the issue's example and a variable shift amount, which now reconstruct to `panic.>>` and `panic.<<`.  The rest of the suite is unchanged and there is no AST change, so charon-ml is unaffected.

### DIFF
--- a/charon/src/transform/resugar/reconstruct_fallible_operations.rs
+++ b/charon/src/transform/resugar/reconstruct_fallible_operations.rs
@@ -8,6 +8,7 @@ use std::collections::HashSet;
 
 use derive_generic_visitor::*;
 
+use crate::ast::ullbc_ast_utils::StmtLoc;
 use crate::ast::*;
 use crate::ids::IndexVec;
 use crate::transform::TransformCtx;
@@ -143,6 +144,24 @@ fn equiv_op(op_l: &Operand, op_r: &Operand) -> bool {
     }
 }
 
+/// A shift overflow check whose shift operation is not in the same block: rustc evaluates the
+/// shift operands (which is when it emits the overflow check) before computing the destination
+/// place, so when that computation involves a function call (e.g. `out[i] = x >> y` through an
+/// overloaded `IndexMut`), the check ends up separated from the shift by the call terminator.
+/// See issue #1041. These asserts are only ever emitted by the compiler, so instead of following
+/// the control flow we record the check and find the matching shift over the whole body in
+/// [resolve_pending_shift_check].
+struct PendingShiftCheck {
+    /// The block containing the check.
+    block: BlockId,
+    /// The checked shift amount; used to find the shift.
+    amount: Operand,
+    /// The comparison result asserted by the check. Identifies the check's statements to remove.
+    cond_local: LocalId,
+    /// The `y as uN` cast temporary, when the amount is cast before the comparison; also removed.
+    cast_local: Option<LocalId>,
+}
+
 /// Rustc inserts dynamic checks during MIR lowering. They all end in an `Assert` statement (and
 /// this is the only use of this statement).
 fn remove_dynamic_checks(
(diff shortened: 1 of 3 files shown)


Nadrieril:
Hi! Thank you for your contribution. Would you mind re-writing your PR description yourself? It is currently decent but still full of unnecessary details that LLMs love to add.

Then, instead of leaving the check, I would like to remove the check and do the arithmetic computation early instead (and store it in a fresh variable). That's always correct since the only side-effects are the panic and the memory accesses, all of which are already done by the overflow check.

Also your solution is doing too much work to be correct compared to what we need: we don't need to follow the graph structure since these asserts are only ever emitted by the compiler: let's simply look for all the uses the given local in the body.

Also I'd like to handle the shift checks in a single pass: iterate over the whole body looking for shifts that use the operand in question, and replace them all. To know where to insert the original computation, record the `StmtLoc` for the original assert. I expect this will generalize better, since it's likely we'll need something like this for other fallible operations in the future.

Nadrieril (approved):
Looks good, thank you! Would you mind merging/rebasing on master, this should fix the spurious CI failure

PR #1340, opened 2026-07-31
Add attributes for pre/postconditions

Nadrieril:
Used as follows:
```rust
fn function() {
    #[charon::precondition]
    fn precondition() -> bool {
        true
    }
}
```
This considers `precondition` as a precondition for `function`, and reflects that in both these item's attributes, as well as the precondition's `ItemSource`.


N1ark:
woa just saw this ! any reason this wasn't done with requires/ensures attributes like what they're working on for contracts ?

Nadrieril:
This is meant to be what requires/ensures attribute macros expand to! So that different frontends can encode what they want; Charon does not interpret the function in any way, just does a bit of bookkeeping. E.g. https://github.com/anodized-rs/anodized could expand to that

PR #1343, opened 2026-08-03
Add layout guarantees

firefighterduck:
Based on the feedback from the first [draft](https://github.com/AeneasVerif/charon/pull/1325) and the discussions on Zulip, this is now the new draft for layout guarantee information in Charon. 

~~### Caveat
Since the guarantees live alongside the concrete layout information, this PR significantly increases the size of `charon/test/layout.json`.~~

### Open TODO's
* add inhabitedness predicates to express guarantees about this property (this will probably be a follow up PR)


charon/src/ast/layout_guarantees.rs
N1ark, line 1: maybe for the sake of consistency and readability we could split this into `layout.rs` with type definitions, `layout_utils.rs` and in `charon-drive` adding a `translate_layout[_guarantees]`?

charon/tests/layout_guarantees.json
N1ark, line 1: bold take but maybe you could define a pretty printer for layout, to make this file simpler to read because having a 13KLoC test output is completely useless, no one will read it and catch errors

charon/src/ast/values_utils.rs
```diff
@@ -241,6 +254,211 @@ impl ScalarValue {
...
+        } else {
+            let ty = if self.ty() == rhs.ty() {
+                *self.as_unsigned().unwrap().0
+            } else {
+                UIntTy::Usize
+            };
+            ScalarValue::from_unchecked_uint(ty, raw_res.try_into().ok()?)
+        };
+        Some(res)
+    }
```
N1ark, line 307: dont you only use all of these for layout? id make it much simpler: 
- enforce `self.ty == rhs.ty` (defaulting to `Usize` if they mismatch could hide actual typing errors)
- just have two branches for either signedness (possibly failing on `rhs.as_uint().unwrap().try_into().ok()?` seems extremely dubious and risky)
- probably have a helper function, like `do_arith<T>(self, rhs: Self, fn_u: fn(u128, u128) -> T, fn_s: fn(i128, i128) -> T)) -> T`, so you dont have this behaviour duplicated (you could also reuse this in `Ord`

charon/src/ast/layout_guarantees.rs
N1ark: id possibly even separate some of it further to make it very **very** clear where the code that handles each guarantee is, and what its source is. i can see you have comments with links but there are 1300 lines of code in that file so it's practically impossible to actually understand where the core logic is and what it does

charon/src/ast/types.rs
```diff
@@ -544,26 +545,55 @@ pub enum PredicateOrigin {
...
+    pub guarantees: SizeExprBound, // TODO: do we need to make this optional?
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, SerializeState, DeserializeState, Drive, DriveMut)]
+pub struct OffsetInformation {
+    #[serde_state(stateless)]
+    #[drive(skip)]
+    pub raw: Option<RawByteCount>,
+    pub guarantees: Option<OffsetGuarantee>,
+}
```
N1ark, line 736: why are guarantees for `ByteCount` not optional but they are in `OffsetInformation`?

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -501,43 +504,103 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                            .collect()
+                    } else {
+                        offsets
+                            .iter()
+                            .map(|o| OffsetInformation {
+                                raw: Some(o.bytes()),
+                                guarantees: None,
+                            })
+                            .collect()
+                    }
```
N1ark: since this pattern comes up a lot, maybe we could have a helper for it, that basically does
```rust
fn zip_opt(lhs: impl IntoIter<L>, rhs: Option<impl IntoIter<R>>) -> impl Iter<(L, Option<R>)>
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -687,15 +793,51 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+                                })
+                                .collect()
+                        } else {
+                            let mut variant_layouts: IndexVec<VariantId, Option<VariantLayout>> =
+                                (0..n_variants).map(|_| None).collect();
+                            variant_layouts[variant_id] =
+                                Some(translate_layout_data(&layout, vec![], None));
+                            variant_layouts
+                        }
                     }
```
N1ark, line 806: same here -- youre adding a lot of code and duplication in several places while pattern-matching on `layouts_guarantee`, which makes it super hard to understand the difference between each branch. having a tiny abstraction to merge both branches would make this a lot easier to parse, imo

charon/src/ast/values_utils.rs
firefighterduck: I can agree with point 3, but the other ones I pretty much need downstream in my tool and they at least make sense for normalizing layout guarantees (since there we have essentially no guarantee that the types or signedness for const expressions in arrays etc.). Maybe we can extract this behavior into functions with more descriptive names and not use the standard traits for it.

charon/tests/layout.json
N1ark, line 1: huh i guess this file could also use a pretty printer:3 we could even imagine in the .llbc output having a flag to print the layout next to each ADT (maybe styling it as a comment?), it would be SO cute

charon/src/ast/layout_guarantees.rs
firefighterduck: I agree that things should be split. It's a bad habit of mine to pack everything into one file. I'll see where it makes sense. For now, I simply hoped that it could make reviewing easier since one doesn't need to jump between files.

charon/src/ast/values_utils.rs
N1ark: rust is well-typed at the MIR level, why do you need to do operations between mismatched integer types? that seems like it will blow up in your face

N1ark: alternatively you could consider just defining this more permissive version of these functions within your codebase, since they seem quite opinionated and not necessarily intuitive, imo

speaking of experience the stricter i am about typing stuff the better, it catches so many latent bugs that would be hidden by lazy coercions

charon/src/ast/types.rs
firefighterduck: `ByteCount` is only really used for the size and alignment of a type, which can always be represented symbolically. For offsets, information about the number of fields and variants might not be available due to opaqueness issues, thus, we might get the actual offsets but no guarantees in general.

charon/src/ast/values_utils.rs
firefighterduck: The problem with strict typing in these cases (especially in my tool where I reuse many charon-lib datatypes for representation in logic) is that information from different sources should be combined where these sources do not agree on the types. In the case of layout guarantees, allowing only `usize` might be okay, I'll try it out.

N1ark: hmm that sounds like we have ill-typed values in Charon, then

charon/src/ast/types.rs
N1ark: ah.. then imo it makes sense to make `ByteCount.guarantees` an option, and remove `OffsetInformation` entirely

firefighterduck: I'd argue that it makes sense to keep `OffsetInformation`. For the offset of a field we always know exactly whether its zero, computed via `repr(C)`, or only that the offset is aligned to something. Thus wrapping it into a `SizeExprBound` seems to only introduce more boilerplate code.

charon/tests/layout_guarantees.json
firefighterduck: I've added "a" pretty printer. Please feel free to suggest improvements.

charon/src/ast/values_utils.rs
firefighterduck: I've made the functions type-safe as you suggested and it works so far.

charon/src/bin/charon-driver/translate/translate_types.rs
firefighterduck: The signature is a bit more involved than what you suggested, but the boiler plate code has been reduced.

charon/src/bin/charon-driver/translate/translate_types.rs
firefighterduck: That's unfortunately the one case where the two branches do significantly different things and cannot easily be combined.

charon/tests/layout.json
firefighterduck: I've added "a" pretty printer (currently only used in the test output, not in the .llbc). Please feel free to suggest improvements.

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -4,17 +4,12 @@ use derive_generic_visitor::*;
 use macros::{EnumAsGetters, EnumIsA, VariantName};
 use serde_state::{DeserializeState, SerializeState};
 
-/// Guaranteed facts about a layout size.
-#[derive(Debug, Clone, SerializeState, DeserializeState, Drive, DriveMut, DriveTwo)]
-pub enum SizeGuarantee {
-    Equals(ExactSizeExpr),
-    AtLeast(ExactSizeExpr),
-}
```
Nadrieril, line 12: hm, we'll have to rename ExactSizeExpr to SizeExpr or SizeGuarantee in a later PR

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -121,6 +116,10 @@ pub enum ExactSizeExprKind {
         then_size: ExactSizeExpr,
         else_size: ExactSizeExpr,
     },
+    /// The symbolic offset of the field within the variant (if any).
```
Nadrieril:
```suggestion
    /// The offset of the given field inside the current type.
```

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -121,6 +116,10 @@ pub enum ExactSizeExprKind {
...
+    FieldOffset(Option<VariantId>, FieldId),
+    /// Knowledge that the inner size expression is actually a lower bound.
```
Nadrieril:
```suggestion
    /// Any value larger than that one.
```

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -183,9 +212,10 @@ impl ExactSizeExpr {
                                     if let Some(ty_ref) = ty.as_adt()
                                         && let Some(decl) = self.krate.type_decls.get(ty_ref.id)
                                         && let Some(layout) = decl.layout.get(self.target)
-                                        && let Some(value) = exact_guarantee(&layout.size)
+                                        && let Some(value) = (layout.size.guarantee.as_ref())
+                                            .and_then(ExactSizeExpr::is_exact)
```
Nadrieril:
```suggestion
```
The idea of `AtLeast` is that we don't need that condition! We can have an AtLeast in the middle of an expression.

charon/src/ast/items/layout.rs
```diff
@@ -43,7 +43,7 @@ pub struct VariantLayout {
     pub field_offsets: IndexVec<FieldId, OffsetExpr>,
     /// Whether the variant is uninhabited, i.e. has any valid possible value.
     /// Note that uninhabited types can have arbitrary layouts.
-    pub uninhabited: bool,
+    pub uninhabited: Option<bool>,
```
Nadrieril, line 47: Could you document what `None`/`Some` mean here?

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -121,6 +116,10 @@ pub enum ExactSizeExprKind {
         then_size: ExactSizeExpr,
         else_size: ExactSizeExpr,
     },
+    /// The symbolic offset of the field within the variant (if any).
+    FieldOffset(Option<VariantId>, FieldId),
```
Nadrieril: Well now this one is a problem for substitution. As it is, we must error on substitution, as of course we can't move this from a type to another. I think we have no choice but to make it this:
```suggestion
    FieldOffset(TypeDeclRef, Option<VariantId>, FieldId),
```

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -512,44 +513,91 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+        {
+            if let Some(rhs) = rhs {
+                Box::new(
+                    lhs.zip(rhs.into_iter())
+                        .map(move |(lhs, rhs)| map(lhs, Some(rhs))),
+                )
+            } else {
+                Box::new(lhs.map(move |lhs| map(lhs, None)))
+            }
+        }
```
Nadrieril: You should be able to use something like this, no:
```suggestion
        fn zip_opt<'a, L, R: 'a, T>(
            lhs: impl Iterator<Item = L>,
            rhs: Option<impl IntoIterator<Item = R, IntoIter: 'a>>,
        ) -> impl Iterator<Item = (L, Option<R>)> {
            let rhs: Box<dyn Iterator<Item=Option<R>> + '_> = if let Some(rhs) = rhs {
                Box::new(rhs.into_iter().map(Some))
            } else {
                Box::new((0..).map(|_| None))
            };
            lhs.zip(rhs)
        }
```

Nadrieril: Or even just the first bit that does `Option<Iterator> -> Iterator<Option>`

charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -582,10 +630,34 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
+            .t_ctx
+            .translated
+            .target_information
+            .get(&the_target)
+            .unwrap()
+            .target_pointer_size;
 
         // If layout computation returns an error, we return `None`.
-        let layout = tcx.layout_of(pseudo_input).ok()?.layout;
+        // If layout computation returns an error, we return `None`.
```
Nadrieril:
```suggestion
```

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Or even, have the fields of `LayoutGuarantees` be the right shape already

charon/tests/layout_guarantees.txt
```diff
@@ -0,0 +1,625 @@
...
+    offset of 
+        0.0 is aligned to align_of::<u8>(),
+        0.1 is aligned to align_of::<u32>(),
+        0.2 is aligned to align_of::<u8>(),
+}
+normalized Guarantees: {
+    SIZE is >= aligned_to(max(offset(0) + 1usize, offset(1) + 4usize, offset(2) + 1usize), 4usize),
+    ALIGN is >= 4usize,
+    offset of 
+        0.0 is aligned to 1usize,
```
Nadrieril: This need to print the variant name, not its id

charon/src/ast/items/layout_guarantees.rs
firefighterduck: Renamed.

charon/src/ast/items/layout_guarantees.rs
firefighterduck: I've removed all `is_exact` checks for now, even though I'm still not sure whether there might be some cases where they make sense.

charon/tests/layout_guarantees.txt
firefighterduck: Changed to always use names. It required some hack, so please check whether that is acceptable.

charon/src/bin/charon-driver/translate/translate_types.rs
firefighterduck: I prefer the `Option<Iterator> -> Iterator<Option>` version. In my opinion, it doesn't make sense to have partial layout guarantees.

charon/tests/layout.rs
```diff
@@ -228,74 +232,184 @@ fn type_layout() -> anyhow::Result<()> {
...
+                                    IntegerValue::from_bits(int_ty, candidate)
+                                }))
+                        });
+                        assert_eq!(
+                            Ok(var_id),
+                            result,
+                            "For type {name} variant {var_id}, tagger = {tagger:?}",
+                        );
+                    }
+                    _ => (),
```
N1ark:
```suggestion
                    None => (),
```

charon/tests/layout.rs
```diff
@@ -228,74 +232,184 @@ fn type_layout() -> anyhow::Result<()> {
...
+        };
+        let name = tdecl.item_meta.name.debug_repr(&crate_data);
+        writeln!(&mut buffer, "{name}")?;
+        let fake_ty = Ty::new(TyKind::Adt(TypeDeclRef {
+            id: tdecl.def_id,
+            generics: Box::new(tdecl.generics.identity_args()),
+            builtin: None,
+        }));
+
+        let opt_concretized = layout_computer.compute_layout_guarantees(fake_ty.clone());
```
N1ark: i dont think i understand; why do you need a `LayoutComputer` here? dont you have the type decl with the layout the driver gave you?

charon/tests/layout.rs
```diff
@@ -228,74 +232,184 @@ fn type_layout() -> anyhow::Result<()> {
...
+        // Check whether concretized layout guarantees always match known layouts.
```
N1ark, line 405: also i think this file would read more clearly if you had separate tests that all received as input the `TranslatedCrate`, rather than one 400 line function that's hard to track

charon/tests/layout.txt
```diff
@@ -0,0 +1,655 @@
...
+    discriminator: omitted because tree-shaped,
+    uninhabited: false,
+    variant_layouts: 
+        @Variant0: VariantLayout { 
+            offset of 0: 0 (guaranteed aligned to align_of::<u32>()),
+            offset of 1: 4 (guaranteed aligned to align_of::<u32>()),
+            offset of 2: 8 (guaranteed aligned to align_of::<u32>()),
+            uninhabited: false,
+            tagger: [],
+        }
```
N1ark: minor, but why do you have `offset(SimpleStruct.y)` at the top, but `offset of 0` at the bottom? could you not use the same printer?

charon/src/ast/items/layout_guarantee_utils.rs
```diff
@@ -0,0 +1,1047 @@
...
+                    align: mk_address_align().into_expr(),
+                    offsets: OffsetGuarantees::None,
+                };
+            }
+            ScalarTy::Integer(IntegerTy::Signed(int_ty)) => int_ty.target_size(0),
+            ScalarTy::Integer(IntegerTy::Unsigned(uint_ty)) => uint_ty.target_size(0),
+            ScalarTy::Float(float_ty) => float_ty.target_size(),
+            ScalarTy::Bool => 1,
+            ScalarTy::Char => 4,
+        };
```
N1ark: this is `ScalarTy.target_size`

charon/src/ast/items/layout_guarantee_utils.rs
```diff
@@ -0,0 +1,1047 @@
...
+pub struct LayoutGuarantees {
+    pub size: SizeGuarantee,
+    pub align: SizeGuarantee,
+    pub offsets: OffsetGuarantees,
+}
+
+struct LayoutGuaranteeComputer<'a, 'b> {
+    krate: &'a TranslatedCrate,
+    target: Option<&'b TargetTriple>,
+}
```
N1ark, line 129: now that i have a fresh look at this; it seems weird to do calculations in charon? thinking out loud but i wonder if having this in charon-driver wouldn't make more sense, and maybe make things simpler(??)

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Given that `LayoutGuarantees` is never exposed to the user, I don't feel that it matters much

charon/src/ast/items/layout_guarantee_utils.rs
Nadrieril: This doesn't do any rustc queries so may as well have it here?

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: Given that `LayoutGuarantees` is never exposed to the user, I don't feel that it matters much

charon/src/ast/items/layout_guarantee_utils.rs
N1ark: yes yes :3 i guess its fine

charon/tests/layout.rs
firefighterduck: This computes the fully normalized and "as concrete as possible" guarantees. It does so by looking up stuff from the type decl, but especially the symbolic offset expressions cannot be substituted without this computer.

charon/tests/layout.txt
firefighterduck: Strange, it should use the same printer, maybe something went missing due to too many rebasings.

charon/src/ast/items/layout_guarantee_utils.rs
firefighterduck: ~~Not exactly. This must not assume any specific target pointer size and thus at least handle `usize` and `isize` differently. Nonetheless, the rest can be simplified by using `ScalarTy.target_size` and I will change it accordingly.~~ Actually, since we decided to have guarantees only for a given target, this can just be `ScalarTy.target_size`.

charon/tests/layout.rs
Nadrieril: Hm, we should separate normalization from computing layouts imo

charon/src/ast/items/layout_guarantee_utils.rs
firefighterduck: I certainly will need both this and especially `LayoutComputer` to be part of `charon-lib` (or some extension, idk).

charon/tests/layout.rs
firefighterduck: They are separate, the name of the struct is open to bikeshedding.

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -100,42 +95,61 @@ pub struct ExactSizeExpr(pub HashConsed<ExactSizeExprKind>);
...
     /// A size expression that depens on whether the given type is inhabited.
     IfInhabited {
         ty: Ty,
-        then_size: ExactSizeExpr,
-        else_size: ExactSizeExpr,
+        then_size: SizeGuarantee,
+        else_size: SizeGuarantee,
     },
+    /// The offset of the given field inside the referenced type.
+    FieldOffset(TypeDeclRef, Option<VariantId>, FieldId),
```
Nadrieril: Can you make this `ConstantExprKind::OffsetOf` instead? Since that exists as a constant (with `offset_of!`)we'll need it anyway, and so `FieldOffset` is redundant.

firefighterduck: Currently, there is only `NullOp::OffsetOf`. Should we remove this and only have it as just a const expression instead?

Nadrieril: I'm on it

Nadrieril: https://github.com/AeneasVerif/charon/pull/1433

Nadrieril:
Superceded by https://github.com/AeneasVerif/charon/pull/1454 :)

PR #1345, opened 2026-08-04
Add `--deallocate-all-locals`

N1ark:
Add a flag to disable the change done in #1316. It had quadratic behaviour on some benchmarks:
```rs
let mut n = 0;
let x = &mut n;
*x += 1;
// 10K times this
```

every `+=` would create a local to store the increment, and each possible overflow assertion failure would add `storage_dead`s for all the previous locals, leading to 50M new statements added.

I think it's sensible to add a flag to disable this, while keeping the `StorageLive` insertion: `StorageLive` is `O(N)` in `N` always-live locals, so the compilation cost is minor and can never blow up, while relieving users of the annoying task of looking for always live locals. In comparison, not marking a local as dead is technically "ok" (the stack is dropped on exit anyways) and is a minor chore to do for the tool, with a potentially much worse perf impact.


Nadrieril:
Oh, fun x) Hmm, would it be correct to add the `StorageDead`s just after the last use of the local, on the assumption that the storageless locals never have a reference/pointer taken to them?

N1ark:
> Oh, fun x) Hmm, would it be correct to add the `StorageDead`s just after the last use of the local, on the assumption that the storageless locals never have a reference/pointer taken to them?

if no references/pointers are taken, yes. this would probably reduce the problem and fix this particular case... i just dont like modifying the MIR too much ><

Nadrieril:
Ok, then I'd prefer to flip the flag please: by default we don't add them, and we update the docs of `StorageDead` to indicate that locals may still be live at function end. The flag `--deallocate-all-locals` would restore today's behavior, and would be enabled by default in the aeneas preset.

N1ark:
done :D

charon/src/ast/llbc_ast.rs
```diff
@@ -45,9 +45,11 @@ pub enum StatementKind {
...
     /// Indicates that this local should be deallocated; if it is already deallocated, this is
     /// a no-op. A local may not have a `StorageDead` in the function's body, in which case it
-    /// is implicitly deallocated at the end of the function. The return local does not receive a
-    /// `StorageDead`. We ensure in the micro-pass `insert_storage_statements` that all other locals
-    /// have a `StorageDead` before function exits.
+    /// is implicitly deallocated at the end of the function; this notably includes the return
+    /// local. In other words, locals may still be live when the function exits. If
+    /// `--deallocate-all-locals` is set, the micro-pass `insert_storage_statements` adds the
+    /// missing `StorageDead`s so that all locals except the return local are deallocated before
+    /// function exits.
```
Nadrieril:
```suggestion
    /// Deallocates the given local; if it is already deallocated, this is
    /// a no-op. Not all local deallocations are explicit: if a non-return local is still live at function end (return or unwind),
    /// it is implicitly deallocated.
    /// If `--deallocate-all-locals` is set, all local deallocations are made explicit.
```

charon/src/options.rs
```diff
@@ -242,6 +242,12 @@ pub struct CliOpts {
     #[clap(long)]
     #[serde(default)]
     pub reconstruct_asserts: bool,
+    /// Ensure every local is deallocated before the function exits, by adding the extra
+    /// `StorageDead`s that MIR omits for some locals. Note this can add a lot of statements
+    /// to functions that have many locals and many exit points.
```
Nadrieril:
```suggestion
    /// Ensure all local deallocations are made explicit with `StorageDead` statements. If this flag is not passed,
    /// every non-return local is implicitly deallocated on function return.
    /// Note this can add a lot of statements (quadratically-many, because of unwind paths).
```

charon/src/options.rs
```diff
@@ -471,6 +477,7 @@ impl CliOpts {
                     self.remove_unused_self_clauses = true;
                     self.remove_adt_clauses = true;
                     self.unbind_item_vars = true;
+                    self.deallocate_all_locals = true;
```
Nadrieril, line 480: Could you also set it for the `test` preset actually please? So that it gets tested thoroughly

PR #1353, opened 2026-08-18
Bump toolchain to 18-08-2026

N1ark:
Bump the toolchain to 18-08-2026! 

Selection of relevant PRs to explain the changes done:
- `ty::FnDef` now has a binder: https://github.com/rust-lang/rust/pull/158632
- Rename `Predicate` to `Clause` in a few places: https://github.com/rust-lang/rust/pull/160343
- Replace `unstable_opts.threads` with `opts.jobs`: https://github.com/rust-lang/rust/pull/159675
- Removed `AliasTy::def_id()` </3: https://github.com/rust-lang/rust/pull/158013
- Merge `InlineConst` with `AnonConst`: https://github.com/rust-lang/rust/pull/158767
- New `TryAsDyn` auto trait: https://github.com/rust-lang/rust/pull/150161
- Removed `alias_ty_kind_from_def_id`: https://github.com/rust-lang/rust/pull/155323
- Renamed `UnevaluatedConst` to `AliasConst`: https://github.com/rust-lang/rust/pull/158115

ci: use https://github.com/AeneasVerif/aeneas/pull/1277
ci: use https://github.com/AeneasVerif/eurydice/pull/439

PR #1355, opened 2026-08-19
Translate test attribute

N1ark:
Based on #1353 

Now that https://github.com/rust-lang/rust/pull/158854 is merged, we can detect test functions easily! Also added printing of attributes to the pretty output before adding it, to show that it indeed works.


charon/src/pretty/fmt_with_ctx.rs
```diff
@@ -1143,6 +1143,16 @@ impl ItemMeta {
...
 
+        for attr in &self.attr_info.attributes {
+            // Doc-comments are long and don't affect the semantics; skip them.
+            if attr.is_doc_comment() {
+                continue;
+            }
+            // An attribute may span several lines; indent each of them.
+            for line in attr.with_ctx(ctx).to_string().lines() {
+                writeln!(f, "{tab}{line}")?;
+            }
```
Nadrieril: The formatting context tracks an indent level, better to respect it in the `Attribute` printer

charon/tests/cargo/issue-396-lib-bin.out
```diff
@@ -59,6 +63,7 @@ where
 }
 
 // Full name: core::fmt::Display
+#[rustc_diagnostic_item = "Display"]
```
Nadrieril: let's maybe stop translating that attribute since we're tracking the value in ItemMeta already

Nadrieril (approved):

PR #1402, opened 2026-09-01
Translate VTables for Fn traits 

### LINKED ISSUE #1388: Bug: closure `Fn*` vtables are never emitted
Follow-up to #1264.

A closure's `Fn*` impl declares a vtable as a global, but that global is never emitted in the final output.

```rs
fn main() { let _x: &dyn Fn() = &(|| {}); }
```

```sh
$ charon rustc --ullbc --print-ullbc --include=_ -- test.rs
```

Output (shortened). No static `impl_Fn_unit_for_closure::{vtable}` is emitted, and the cast keeps `&vtable_of(...)`:
```
opaque type core::ops::function::Fn::{vtable}

impl "impl_Fn_unit_for_closure" Fn<()> for closure {
    ...
    vtable: impl_Fn_unit_for_closure::{vtable}      // doesn't exist
}

fn main() {
    ...
    _x = unsize_cast<&'6 closure, &'10 (dyn Fn<...>), &vtable_of(impl_Fn_unit_for_closure)>(move _2);
                                                      ^ not normalised to a global reference
}
```

N1ark:
Closes #1388

Generate VTables and their shims for the `Fn*` traits, for closures, function items and function pointers! Also fixed the behaviour of `dyn FnOnce` and other dyn-compatible functions that take `Self`, by making the shim receive `Self` by pointer (this is also what rustc does)

I don't love the code, but I'm also not sure how to do this better; `translate_trait_objects` is already pretty intense:') 

LLMS were used for this PR, though I went through and cleaned up and read the code.

### DIFF
--- a/charon/tests/ui/dyn/builtin-vtables.out
+++ b/charon/tests/ui/dyn/builtin-vtables.out
@@ -44,7 +44,7 @@ struct core::ops::function::FnOnce::{vtable}<Args, Ty0> {
   size: usize,
   align: usize,
   drop: unsafe fn<'_0_1>(&'_0_1 mut (dyn FnOnce<Args, Output = Ty0>)),
-  method_call_once: extern "rust-call" fn((dyn FnOnce<Args, Output = Ty0>), Args) -> Ty0,
+  method_call_once: extern "rust-call" fn(*mut (dyn FnOnce<Args, Output = Ty0>), Args) -> Ty0,
   super_trait_0: &'static core::marker::MetaSized::{vtable},
 }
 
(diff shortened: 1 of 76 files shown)


charon/tests/ui/dyn/builtin-vtables.out
```diff
@@ -147,6 +152,72 @@ fn {impl Fn<(u32,)> for some_fn}::call<'_0>(state: &'_0 some_fn, args: (u32,)) -
...
+    _0 = {impl FnOnce<(u32,)> for some_fn}::call_once(move _3, move _2)
+    ↳⚡ storage_dead(_3)
+        storage_dead(_2)
+        storage_dead(_1)
+        unwind_continue
+    storage_dead(_3)
+    storage_dead(_2)
+    storage_dead(_1)
+    return
+}
```
N1ark, line 175: i don't love this, because it requires passing a `dyn FnOnce` as an argument and then reading it and concretizing it.... from what i understand, the way rustc does these shims is that functions that expect an `dyn FnOnce` actually tread it as behind a pointer. im gna check with Obol to understand if there is a better way of doing this ^^'

N1ark: ok indeed; the way rustc handles unsized arguments is that:
- on the caller side, unsized arguments are actually passed *by thin pointer*, despite the signature saying `self: Self`
- on the callee side, it actually has `self: *mut Self` (conceptually this is `&own Self`), with `Self` monomorphic, meaning it can be loaded into the right type

see: https://github.com/rust-lang/rust/blob/70222712809cd5cc1718ed8995914a1cbacb6b92/compiler/rustc_middle/src/ty/instance.rs#L101-L106

i would muchhhh rather do this, but it does imply some dodgy arg-passing shenanigans. Soteria Rust already handles it fine, by just checking for all arguments of a function if the argument (on the callee) is unsized, in which case the argument must be `Move place | Copy place`, and we just pass `place` as a pointer.

see: https://github.com/soteria-tools/soteria/blob/main/soteria-rust/lib/interp.ml#L1125-L1147

N1ark: fixed in a separate PR; i much prefer this :)

PR #1404, opened 2026-09-01
Fix bugs

### LINKED ISSUE #1391: Bug: duplicate types in monomorphic code because of `dyn Trait`
In some cases, monomorphisation emits the same type as separate declarations, which causes issues for consumers since the program looks mistyped.

```rust
trait Sub {}
struct D;
impl Sub for D {}
fn main() {
    let d = D;
    let p = &d as *const dyn Sub;
    let _ = unsafe { p.as_ref() };
}
```

```
$ charon rustc --monomorphize --print-ullbc -- --crate-type=lib test.rs
```

`Option<&dyn Sub>` is declared twice, identically:

```rust
pub enum Option::<&'_ (dyn Sub)> {
  None,
  Some { _0: &'_ (dyn Sub) },
}

pub enum Option::<&'_ (dyn Sub)> {
  None,
  Some { _0: &'_ (dyn Sub) },
}
```

## Expected

One declaration per monomorphic type.

### LINKED ISSUE #1394: Panic: index out of bounds on a `self: Box<Self>` method
A trait method with a by-box receiver, called through `dyn`, panics.

```rust
trait Trait { fn method(self: Box<Self>); }
struct S;
impl Trait for S { fn method(self: Box<Self>) {} }
fn main() {
    let b: Box<dyn Trait> = Box::new(S);
    b.method();
}
```

```
$ charon rustc --monomorphize --no-typecheck -- test.rs

thread 'main' panicked at index_vec-0.1.4/src/indexing.rs:128:10:
index out of bounds: the len is 0 but the index is 0
```

Without `--no-typecheck` the file dies earlier, in the type checker:

```
$ charon rustc --monomorphize -- test.rs

thread 'main' panicked at src/transform/typecheck_and_unify.rs:210:17:
assertion `left == right` failed
  left: 2
```

N1ark:
Fix #1391 -- the bug came from using the context's number of args when adding a new predicate to `dyn`, leading to the ID of that argument varying depending on the context that created it

Fix #1394 -- the bug came from two places: 
- `TypeDeclRef::as_box`, which panics in monomorphic code since there are no generics
- typechecking concretisation of ADTs would fail in monomorphic code, because both sides would have mismatched IDs

### DIFF
--- a/charon/rustc_trait_elaboration/src/elaboration.rs
+++ b/charon/rustc_trait_elaboration/src/elaboration.rs
@@ -102,7 +102,6 @@ impl<'tcx, Id: ItemId> Candidate<'tcx, Id> {
 #[derive(Clone)]
 pub struct PredicateSearcher<'tcx, Id: ItemId = DefId> {
     pub(crate) elab_ctx: ElaborationCtx<'tcx, Id>,
-    owner: Id,
     pub(crate) typing_env: rustc_middle::ty::TypingEnv<'tcx>,
     /// Local clauses available in the current context.
     candidates: HashMap<PolyTraitRef<'tcx>, Candidate<'tcx, Id>>,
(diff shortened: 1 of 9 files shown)


Nadrieril:
Looks good!

PR #1405, opened 2026-09-01
Fix missing statics!

### LINKED ISSUE #1393: Bug: missing global bodies
Some functions reach the output declared but with no body, not sure why though. Maybe some are externs and not being detected as such? Some seem related to thread statics, so maybe we don't translate that entirely properly? Very unsure ^-^

```rust
thread_local! {
    static X: u32 = 42;
}
fn main() {
    X.with(|_v| {});
}
```

```
$ charon rustc --extract-opaque-bodies --print-ullbc -- test.rs
```

## Generated code

Eight bodies are missing:
- `std::sys::thread_local::destructors::list::DTORS`
- `std::thread::current::CURRENT`
- `std::sys::thread_local::guard::apple::enable::REGISTERED`
- `core::io::error::os_functions::OS_FUNCTIONS`
- `std::panicking::panic_count::GLOBAL_PANIC_COUNT` 

Other programs reach the same state through different statics — `std::alloc::HOOK`, `std::rt::cleanup::CLEANUP`, `std::backtrace::{Backtrace}::enabled::ENABLED`, and `std::hash::random::{RandomState}::new::KEYS`.

This is because we give up in `translate_bodies.rs:205` when the item has no MIR and cannot be const-evaluated:
https://github.com/AeneasVerif/charon/blob/266153b4a07b37db8127b663b240da358d2f00c4/charon/src/bin/charon-driver/translate/translate_bodies.rs#L206-L223

N1ark:
Fixes #1393

The issue was in fact three problems in a trenchcoat! Hurray

- We were trying to read through dangling pointers in uneval, which gave up translation. We now check properly
- Const eval gives up in the present of interior mutability, so we cheat a bit to still const-evaluate.
- Const eval also gives up for poly constants, but if they're trivial we can get them regardless!!

### DIFF
--- a/charon/src/bin/charon-driver/hax/constant_utils/uneval.rs
+++ b/charon/src/bin/charon-driver/hax/constant_utils/uneval.rs
@@ -262,7 +262,7 @@ fn op_to_const<'tcx, S: UnderOwnerState<'tcx>>(
         // Detect statics
         _ if let Some(place) = op.as_mplace_or_imm().left()
             && let ptr = place.ptr()
-            && let (alloc_id, _, _) = ecx.ptr_get_alloc_id(ptr, 0)?
+            && let Some((alloc_id, _, _)) = ecx.ptr_get_alloc_id(ptr, 0).discard_err()
             && let interpret::GlobalAlloc::Static(did) = tcx.global_alloc(alloc_id) =>
         {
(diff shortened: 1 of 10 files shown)


Nadrieril:
Nice!

PR #1406, opened 2026-09-02
Make `Span` tiny

### LINKED ISSUE #1115: Performance: Span is huge (88 Bytes)
in current implementation, Span is huge - it has whopping 88 Bytes. since it is inlined everywhere, it makes a lot of types large.

for comparison, `rustc_span::Span` has 8 Bytes.

didn't do profiling yet - i encountered this issue running clippy from https://github.com/OpenSourceVerif/trust2-contract/pull/65

N1ark:
Fixes #1115 

Make `Span` on the Rust side a single `u64`, to reduce the memory foot print of most types. To do this, we encode the span the following way:

| 63 | 62..47 | 47..27 | 27..17 | 17..10 | 10..0 |
| -- | -- | -- | -- | -- | -- |
| wide | file(16) | beg.line | beg.col | nb lines | end.col |

From compiling syn, serde and regex, a quick analysis shows that basically all spans are not wide (i.e. `generated_from_span` is `None`), and range over very small values. So we use this small layout for the very vast majority of cases, and if the span doesn't fit we set the upper bit (`wide`) to 1 and have a global table of 63-bit index to full span, using the pre-existing (large) AST.

In practice, basically all spans fit the u64 and don't need further work. The crate with the most translated spans, `syn`, with 800K spans, only had 36 wide spans in the global table.

This improves performance a bit and memory usage a lot! The numbers don't seem massive, but peak RSS includes rustc, which is about 70% of total memory usage. e.g. for `syn`, rustc's memory usage is of about 250MB, so we made Charon's usage go from ~200MB to ~150MB! 

| crate | time | peak RSS | charon's RSS |
| --- | -- | -- | -- | 
| regex | -1.4% |  −6.8% (235→219 MB)  | −19.3% (83→67 MB) |
| serde | −1.5% | −8.0% (249→229 MB)  | −18.9% (106→86 MB) | 
| syn | −1.8% | −10.7% (467→417 MB) |  −23.5% (213→163 MB) |

Small analysis, for the curious:

| observation | serde | regex | syn |
| --- | -- | -- | -- |
| spans with `generated_from_span` | < 1% | < 1% | < 1% | 
| `beg.line == end.line` | 90% | 90% | 89% |
| max column | 141 | 145 | 145 |
| max line | 4423 | 4401 | 4380 |

| crate | `Span`s constructed | distinct entries in `WIDE_SPANS` |
| --- | -- | -- |
| serde 1.0.228 | 246,402 | 12 |
| regex 1.11.1  | 135,786 | 25 |
| syn 2.0.104 | 800,203 | 36 |

### DIFF
--- a/charon/src/ast/meta/spans.rs
+++ b/charon/src/ast/meta/spans.rs
@@ -1,7 +1,5 @@
 use crate::utils::dedup::*;
-use derive_generic_visitor::{
-    Break, Continue, ControlFlow, Drive, DriveMut, DriveTwo, VisitMut, Visitor,
-};
+use derive_generic_visitor::{ControlFlow, Drive, DriveMut, DriveTwo, Visit, VisitMut, VisitTwo};
 use serde::{Deserialize, Serialize};
 use serde_state::{DeserializeState, SerializeState};
 use std::collections::HashMap;
@@ -131,8 +129,6 @@ pub struct SpanData {
 // - For serde 1.0.228, out of 246K spans, 12 didn't fit
 // - For regex 1.11.1, out of 136K spans, 25 didn't fit
 // - For syn 2.0.104, out of 800K spans, 36 didn't fit
-// Ordering is the ordering of the packed value: for packed spans the layout above makes this the
-// source order (file, then start, then extent); wide spans sort last.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Span(u64);
 
@@ -177,6 +173,9 @@ mod pack {
     Deserialize,
     SerializeState,
     DeserializeState,
+    Drive,
+    DriveMut,
+    DriveTwo,
 )]
 #[cfg_attr(feature = "charon_on_charon", charon::rename("Span"))]
 #[serde_state(stateless)]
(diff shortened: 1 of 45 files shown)


charon/src/ast/meta/spans.rs
```diff
@@ -91,50 +96,269 @@ pub struct SpanData {
...
+// ```
+// The spans that don't fit this layout -- because they come from a macro expansion, span many
+// lines, or point into a very large file or a very long line -- are stored in `WIDE_SPANS` and
+// referred to by index.
+// Some numbers to back this up:
+// - For serde 1.0.228, out of 246K spans, 12 didn't fit
+// - For regex 1.11.1, out of 136K spans, 25 didn't fit
+// - For syn 2.0.104, out of 800K spans, 36 didn't fit
+// Ordering is the ordering of the packed value: for packed spans the layout above makes this the
+// source order (file, then start, then extent); wide spans sort last.
```
Nadrieril: We need span ordering to be exact when we reconstruct comments. Should be too hard tho, no?

charon/src/ast/meta/spans.rs
```diff
@@ -91,50 +96,269 @@ pub struct SpanData {
...
+}
+
+/// A `Span` has no contents to visit: what it contains is packed into an integer. We must however
+/// let mutable visitors edit the `FileId`s inside, as the multi-target export relies on this to
+/// renumber files.
+impl<'s, V: Visitor> Drive<'s, V> for Span {
+    fn drive_inner(&'s self, _v: &mut V) -> ControlFlow<V::Break> {
+        Continue(())
+    }
+}
```
Nadrieril: This needs to visit the `FileId` too then, lest the mutable and immutable visitors mismatch

charon/src/ast/meta/spans.rs
```diff
@@ -91,50 +96,269 @@ pub struct SpanData {
...
+impl<'s, V: Visitor<Break: Default>> DriveTwo<'s, V> for Span {
+    fn drive_two_inner(&'s self, other: &'s Self, _v: &mut V) -> ControlFlow<V::Break> {
+        if self == other {
+            Continue(())
+        } else {
+            Break(Default::default())
+        }
+    }
+}
```
Nadrieril: Needs to visit the FileIds too

Nadrieril (approved):
Impressive work, happy to merge modulo the comments I added, and after resolving conflicts

charon/src/ast/meta/spans.rs
N1ark: sure, i can; fyi span ordering is not used in comment reconstruction though, bc there you already used `data().beg.line`. the only place that needs `Span: Ord` is `errors.rs`, where you have
```rs
spans.sort(); // Sort spans to display in file order
```

PR #1408, opened 2026-09-03
Avoid hardcoding names in `PathElem::Ident`

### LINKED ISSUE #1169: Don't use made-up strings in item names
We sometimes generate a couple different items from a single rust items. For these we add dummy name fragments, e.g. the vtable for a trait will be named `path::to::Trait::{vtable}`. I think it's fine to print it like this, but this name element should be an enum instead of the literal string `"{vtable}"`. That way all identifiers in a name will be things that come from the source code.

N1ark:
Closes #1169 

This was itching me so thought I'd do it real quick :) 

ci: use https://github.com/AeneasVerif/eurydice/pull/446
ci: use https://github.com/AeneasVerif/aeneas/pull/1334

### DIFF
--- a/charon-ml/src/CharonVersion.ml
+++ b/charon-ml/src/CharonVersion.ml
@@ -1,3 +1,3 @@
 (* This is an automatically generated file, generated from `charon/Cargo.toml`. *)
 (* To re-generate this file, rune `make` in the root directory *)
-let supported_charon_version = "0.1.248"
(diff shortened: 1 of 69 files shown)


N1ark:
did not expect this to cause breakage in aeneas/eurydice... damn

Nadrieril (approved):
Looks very nice, could you fix the conflictsand update the linked PRs?

PR #1415, opened 2026-09-08
Fix some bugs

N1ark:
Fix three bugs encountering when evaluating Charon

- The `Discriminator` translation was wrong, and wasn't marking some tags as invalid. I redid the whole translation there, copying more closely what MiniRust does. I think it also reads a bit more clearly.
- Fix a panic with closures in const and static contexts, which had the wrong generics applied to them, causing a panic.
- Fix globals being removed by the const inlining pass even in the presence of `--raw-consts`

### DIFF
--- a/charon/src/transform/mod.rs
+++ b/charon/src/transform/mod.rs
@@ -324,7 +324,9 @@ impl TransformCtx {
                     }
                 });
                 for pass in passes.iter() {
-                    pass.finalize(self);
+                    if pass.should_run(&self.options) {
+                        pass.finalize(self);
+                    }
                 }
             }
             Pass::FusedStructuredBody(passes) => {
--- a/charon/tests/ui/consts/ref-constant.out
+++ b/charon/tests/ui/consts/ref-constant.out
@@ -1,6 +1,20 @@
 # Final LLBC before serialization:
 
-struct () {}
+// Full name: test_crate::REF::{promoted_const}
+fn {promoted_const}() -> &'_ u32
+{
+    let _0: &'0 u32; // return
+    let _1: u32; // anonymous local
+
+    storage_live(_0)
+    storage_live(_1)
+    _1 = const 42u32
+    _0 = &_1
+    return
+}
+
+// Full name: test_crate::REF::{promoted_const}
+const {promoted_const}: &'_ u32 = {promoted_const}()
 
 // Full name: test_crate::REF
 fn REF() -> &'static u32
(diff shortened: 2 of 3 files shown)


charon/src/bin/charon-driver/translate/translate_types.rs
```diff
@@ -645,45 +651,87 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
...
-                            };
-                            children = vec![(start..=end, discriminator)];
+                        // The untagged variant may itself be in the range of niched variants, in
+                        // which case its niche value must never be encountered.
+                        if niche_variants.contains(untagged_variant) {
+                            let relative =
+                                (untagged_variant.index() - niche_variants.start.index()) as u128;
+                            let val = tag_from_bits(
+                                tag_size.truncate(niche_start.wrapping_add(relative)),
+                            );
```
Nadrieril: This duplicates some of the `tag_for_variant` logic, could we deduplicate somehow?

charon/src/bin/charon-driver/translate/get_mir.rs
```diff
@@ -41,7 +41,7 @@ impl<'tcx> ItemTransCtx<'tcx, '_> {
                 let s = self.hax_state_with_id();
                 Some(if self.monomorphize() {
                     let typing_env = s.typing_env();
-                    let args = item_ref.rustc_args(s);
+                    let args = item_ref.rustc_args_for_def(s);
```
Nadrieril: Hm, this sounds hyper brittle. We should change `rustc_args`, no? just match on the DefKind stored in the hax::DefId. Also, inline consts also have synthetic type params that we sometimes skip, so they might need the same treatment. But more importantly, this forgetting of synthetic args is done on purpose here https://github.com/AeneasVerif/charon/blob/45c1e30e5536a4425e3a1da9407fa2e7a29f6699/charon/rustc_trait_elaboration/src/item_ref.rs#L214-L216

maybe we shouldn't be truncating them at all?

Nadrieril: see https://github.com/AeneasVerif/charon/pull/1418

charon/src/bin/charon-driver/translate/translate_types.rs
Nadrieril: I attempted that here: https://github.com/AeneasVerif/charon/pull/1419

PR #1416, opened 2026-09-08
Improve unsizing algorithm for `Rc` & co

### LINKED ISSUE #855: Feature request: Support unsizing behind custom smart pointers
Rust supports unsizing behind custom smart pointers:
```rust
let x: Arc<[u32; 2]> = Arc::new([0, 1]);
let _: Arc<[u32]> = x;
```

As of writing, this is enabled on a pointer using `#[derive(CoercePointee)]`, but the load-bearing traits are [`CoerceUnsized`](https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html) and [`Unsize`](https://doc.rust-lang.org/std/marker/trait.Unsize.html). For our example here the relevant impls are:
```rust
impl<T: ?Sized + Unsize<U>, U: ?Sized, A: Allocator> CoerceUnsized<Arc<U, A>> for Arc<T, A> {}
impl<const N: usize, T> Unsize<[T]> for [T; N] {} // compiler-generated
```

To support this in Charon, my idea would be to add fake methods to these traits that perform the operation (like we do for `Destruct::drop_in_place`). We'd then fill the methods with auto-generated code that does the right thing, and call these methods when a cast happens.
```rust
/// Types that wrap a pointer.
trait CoerceUnsized<T: ?Sized> {
    // Note: unsure how to express coercion if either side is unsized.
    fn coerce(self) -> T where Self: Sized, T: Sized;
    /// The reverse of `coerce`, using `Unsize::resize`.
    /// SAFETY: `x` must have the correct unsizing metadata.
    unsafe fn uncoerce(x: T) -> Self where Self: Sized, T: Sized;
}
/// Types that can be cast to an unsized type.
trait Unsize<T: ?Sized> {
    /// Unsize a pointer to `Self`.
    fn unsize(*const self) -> *const T;
    /// The reverse of `unsize`.
    /// SAFETY: `ptr` must have the correct unsizing metadata for `Self`.
    unsafe fn resize(ptr: *const T) -> *const self;
}
```

This can be done in several steps. This isn't an easy issue, come discuss on Zulip if you'd like to tackle it.

EDIT: the bit about `Unsize` doesn't actually make sense, see below.

N1ark:
Fix missing unsizing metadata for custom pointers. This doesn't entirely close #855 , as we don't desugar the coercion or pass the field indices, which have to be recomputed in the client. I would like changing this (which should require a minor representation change), but I'll leave it for later :)

### DIFF
--- a/charon/tests/ui/unsize.out
+++ b/charon/tests/ui/unsize.out
@@ -52,6 +52,45 @@ impl "impl_Copy_for_i32" Copy for i32 {
     non-dyn-compatible
 }
 
+// Full name: core::marker::Unsize
+#[lang_item(Unsize)]
+pub trait Unsize<Self, T>
+
+struct () {}
+
+// Full name: core::marker::Destruct
+#[lang_item(Destruct)]
+pub trait Destruct<Self>
+{
+    fn drop_glue<'_0_1>;
+    non-dyn-compatible
+}
+
+// Full name: core::marker::CoercePointeeValidated
+#[lang_item(CoercePointeeValidated)]
+pub trait CoercePointeeValidated<Self>
+{
+    proof ImpliedClause0: (Self: MetaSized)
+    vtable: core::marker::CoercePointeeValidated::{vtable}
+}
+
+// Full name: core::ops::unsize::CoerceUnsized
+#[lang_item(CoerceUnsized)]
+pub trait CoerceUnsized<Self, T>
+{
+    proof ImpliedClause0: (Self: Sized)
+    non-dyn-compatible
+}
+
+// Full name: core::ops::unsize::DispatchFromDyn
+#[lang_item(DispatchFromDyn)]
+pub trait DispatchFromDyn<Self, T>
+{
+    proof ImpliedClause0: (Self: Sized)
+    proof ImpliedClause1: (T: Sized)
+    non-dyn-compatible
+}
+
 // Full name: alloc::alloc::Global
 #[lang_item(GlobalAlloc)]
 pub struct Global {}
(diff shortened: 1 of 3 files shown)


charon/tests/ui/unsize.out
```diff
@@ -199,99 +376,199 @@ fn foo()
...
     ↳⚡ unwind_continue
-    _8 = unsize_cast<Rc<[i32; 2usize]>[{built_in impl MetaSized for [i32; 2usize]}, {built_in impl Sized for Global}], Rc<[i32]>[{built_in impl MetaSized for [i32]}, {built_in impl Sized for Global}], ?>(move _9)
-    conditional_drop[impl_Destruct_for_Rc::drop_glue<'34, [i32; 2usize], Global>[{built_in impl MetaSized for [i32; 2usize]}, {built_in impl Sized for Global}]] _9
-    ↳⚡ conditional_drop[impl_Destruct_for_Rc::drop_glue<'36, [i32], Global>[{built_in impl MetaSized for [i32]}, {built_in impl Sized for Global}]] _8
+    _8 = unsize_cast<Rc<[i32; 2usize]>[{built_in impl MetaSized for [i32; 2usize]}, {built_in impl Sized for Global}], Rc<[i32]>[{built_in impl MetaSized for [i32]}, {built_in impl Sized for Global}], 2usize>(move _9)
+    conditional_drop[impl_Destruct_for_Rc::drop_glue<'50, [i32; 2usize], Global>[{built_in impl MetaSized for [i32; 2usize]}, {built_in impl Sized for Global}]] _9
+    ↳⚡ conditional_drop[impl_Destruct_for_Rc::drop_glue<'52, [i32], Global>[{built_in impl MetaSized for [i32]}, {built_in impl Sized for Global}]] _8
```
N1ark, line 381: the diff is a mess to read x) the important point is that here the `unsize_cast<_, _, ?>` became `unsize_cast<_, _, 2usize>` :D

Nadrieril:
Very cool!

PR #1418, opened 2026-09-08
Fix incorrect generic arg truncations

Nadrieril:
Fixes a bug found by @N1ark

### DIFF
--- a/charon/rustc_trait_elaboration/src/item_ref.rs
+++ b/charon/rustc_trait_elaboration/src/item_ref.rs
@@ -138,6 +138,9 @@ pub struct ItemRefContents<'tcx, Id: ItemId = DefId> {
     pub def_id: Id,
     /// The arguments provided to this item.
     pub generic_args: ty::GenericArgsRef<'tcx>,
+    /// The number of generic arguments excluding the extra inference arguments of closures and
+    /// inline consts.
+    pub proper_arg_count: usize,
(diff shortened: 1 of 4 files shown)


charon/rustc_trait_elaboration/src/item_ref.rs
```diff
@@ -210,11 +213,6 @@ impl<'tcx, Id: ItemId> PredicateSearcher<'tcx, Id> {
         // Normalize the generics.
         let mut generics = normalize(tcx, typing_env, ty::Unnormalized::new_wip(generics));
```
N1ark, line 214: can you make this `let generics` now?

charon/rustc_trait_elaboration/src/item_ref.rs
```diff
@@ -297,7 +303,7 @@ impl<'tcx, Id: ItemId> ItemRef<'tcx, Id> {
         } else {
             0
         };
-        &self.generic_args[start..]
+        &self.generic_args[start..self.proper_arg_count]
```
N1ark, line 306: it seems weird this isn't `start..(start + self.proper_arg_count)`...?

Nadrieril: The trait args are part of the proper args tho

charon/rustc_trait_elaboration/src/item_ref.rs
Nadrieril: we still modify the generics in case we can resolve an assoc item

PR #1425, opened 2026-09-09
Translate anon const as globals

N1ark:
This is an extra step towards #1225

Fix a bug where const pointers offset into an allocation lost the offset:
```rust
static FOO: [u8; 3] = [1, 2, 3];
static FOO_FST: &u8 = &FOO[0] // this stays translated as &FOO, which is ok.
static FOO_SND: &u8 = &FOO[1] // this was translated as &FOO. we now translate it as &2.
```
For now, we simply give up when the offset is not 0, reading the pointee. Ideally, we should modify hax's `ConstantExprKind::Borrow`/`ConstantExprKind::RawBorrow` and their Charon equivalents to contain the offset within the allocation.. or something like that? Not sure where the best place is to do that.

We now support translating anonymous, untyped allocations (`GlobalAlloc::Memory`)! They are translated as `MaybeUninit<[u8; N]>`.


charon/src/bin/charon-driver/hax/constant_utils/uneval.rs
```diff
@@ -378,17 +426,6 @@ fn op_to_const<'tcx, S: UnderOwnerState<'tcx>>(
...
-        // rest of the machinery ICE, so we don't turn them into named globals here.
-        _ if let Some(place) = op.as_mplace_or_imm().left()
-            && let ptr = place.ptr()
-            && let Some((alloc_id, _, _)) = ecx.ptr_get_alloc_id(ptr, 0).discard_err()
-            && let interpret::GlobalAlloc::Static(did) = tcx.global_alloc(alloc_id)
-            && let rustc_hir::def::DefKind::Static { nested: false, .. } = tcx.def_kind(did) =>
-        {
-            let item = translate_item_ref(s, did, ty::GenericArgsRef::default());
-            ConstantExprKind::NamedGlobal(item)
-        }
```
Nadrieril, line 391: That destroys sharing tho, no?

N1ark: its just been moved up! and made less wrong (at the cost of sharing, but at least dereferencing is correct)

N1ark: see `pointee_to_const`

Nadrieril:
First three commits looks very good, happy to merge them as is. The last two I'll need a bit more time to wrap my head around; would you mind splitting them off to another PR?

N1ark:
> First three commits looks very good, happy to merge them as is. The last two I'll need a bit more time to wrap my head around; would you mind splitting them off to another PR?

done! https://github.com/AeneasVerif/charon/pull/1444

Nadrieril:
Very nice!

PR #1428, opened 2026-09-09
Take by-value `self` vtable receivers via `*mut Self`

N1ark:
Based on top of #1427 

This is a bit weird, but is what rustc does and avoid gnarly unsized locals.

While the caller of a trait object that takes `self` by values sees the signature `fn(Self, ...Args)`, on the callee side the signature is actually `fn(*mut Self, ...Args)`, which morally is `fn(&own Self, ...Args)`.  This is how rustc chooses to handle unsized function arguments, and for now it seems like an Okay way of doing, so that consumers can at least support it in some homogeneous way.

I would not be opposed to improve the representation in the future to more explicitly describe these cases, though I'm not sure how atm.

rustc source: https://github.com/rust-lang/rust/blob/70222712809cd5cc1718ed8995914a1cbacb6b92/compiler/rustc_middle/src/ty/instance.rs#L101-L106

soteria rust support (for the curious): https://github.com/soteria-tools/soteria/blob/main/soteria-rust/lib/interp.ml#L1125-L1147

### DIFF
--- a/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
+++ b/charon/src/bin/charon-driver/translate/translate_trait_objects.rs
@@ -1169,11 +1169,19 @@ impl<'tcx> ItemTransCtx<'tcx, '_> {
     /// // call the impl function and assign the result to ret@0
     /// ret@0 := impl_func(target_self@(N+1), arg1@2, ..., argN@N);
     /// ```
+    ///
+    /// For a method that takes `self: Self` by value, the shim receiver is `*mut dyn Trait`, so
+    /// we concretize the pointer and move out of it:
+    /// ```ignore
+    /// target_self@(N+1) := concretize_cast<*mut dyn Trait, *mut TargetReceiverTy>(shim_self@1);
+    /// ret@0 := impl_func(move (*target_self@(N+1)), arg1@2, ..., argN@N);
+    /// ```
     fn translate_vtable_shim_body(
         &mut self,
         span: Span,
         target_receiver: &Ty,
         shim_signature: &FunSig,
+        receiver_is_by_value: bool,
         impl_func_def: &hax::FullDef,
     ) -> Result<Body, Error> {
         let mut builder = BodyBuilder::new(span, shim_signature.inputs.len());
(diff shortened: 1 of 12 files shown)


charon/src/bin/charon-driver/translate/translate_trait_objects.rs
```diff
@@ -395,13 +395,23 @@ impl<'tcx> ItemTransCtx<'tcx, '_> {
                                 .insert(hax::GenericPredicateId::TraitSelf, TraitClauseId::ZERO);
                             ctx.translate_poly_fun_sig(span, sig)
```
Nadrieril: Let's thread an adjusted signature there instead of doing a manual adjustment. I did https://github.com/AeneasVerif/charon/pull/1429 which should help

N1ark: sure :) ill let you merge that and rebase then

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -124,13 +124,49 @@ fn transform_dyn_trait_call(
...
+        dyn_trait_place = place.clone();
+        ctx.statements.pop();
+    }
+
+    // A receiver taken by value has the unsized type `dyn Trait` itself, so it lives behind a
+    // wide pointer, and it's that pointer which carries the vtable. Walk back out to it: out of
+    // the dereference that produced the value, and out of any struct the value is the unsized
+    // tail of, since such a struct's metadata is its tail's metadata. This is the same walk as
+    // `compute_place_metadata`. See `dyn/mono-dyn-call-by-value.out`, where the by-ref call
+    // passes `move _5` but the by-value one passes `move (*_7)`.
```
Nadrieril:
```suggestion
    // `dyn Trait` has a carveout for unsized `self` types. It works by calling the method on a place of type `dyn Trait` directly.
    // The actual shim we generate expects `*mut dyn Trait`, so we must borrow that unsized place.
    // See `dyn/mono-dyn-call-by-value.out`.
```
and then take a borrow, else we have a typecheck problem

charon/src/transform/normalize/transform_dyn_trait_calls.rs
```diff
@@ -124,13 +124,49 @@ fn transform_dyn_trait_call(
...
+        Operand::Copy(place) | Operand::Move(place) => place.clone(),
         Operand::Const(_) => {
             panic!("Unexpected constant as receiver for dyn trait method call")
         }
     };
 
+    // Rustc may move the unsized first argument of a call by value into a temporary local,
+    // e.g. for `StructWithTail<dyn T>>` and `b.field.by_value()`: `_15 = move (*f)._0; by_val(move _15)`.
+    // Not only does that local not have metadata, it is also unsized, which we want to avoid.
+    // We inline it back into the call, giving us: `by_val(move (*f)._0)`.
```
Nadrieril:
```suggestion
    // Rustc may move the unsized first argument of by-value dyn call into a temporary local,
    // e.g. for `StructWithTail<dyn T>>` and `b.field.by_value()`, we get `_15 = move (*b).field; by_value(move _15)`.
    // That's an unsized local, which we is not supposed to happen. To avoid that we inline the
    // place back into the call, giving us: `by_value(move (*b).field)`.
```

PR #1436, opened 2026-09-12
Performance improvements

N1ark:
Some perf improvements, notably for `--monomorphize`:

- Skip translating all trait methods in monomorphised mode: we don't care about methods on trait impls, so we should stop translating them all! This saves *a lot* of translation time, when using `--extract-opaque-bodies`, since hitting a trait impl very quickly pulls dozens of methods, which themselves pull hundreds of methods, etc. I had a small test that used a `for i in 0..50`, and that pulled in 1650 items bc it pulls ranges, iters, display stuff for panics, etc etc etc. when only 50 functions are actually reachable.
- Keep polymorphic items opaque in monomorphic mode: they are only used for names (as they are the parent of methods), and would pull in a bunch of polymorphic items we don't really want.
- To reduce the size of `hax::Repr` (was 48 bytes), make `Align` a `NonZeroU8` (the log2 of the align, + 1), making `Repr` now only take 16 bytes :) the max align allowed by Rust is 2^31, so 8 bits is amply sufficient. `NonZerou8` is because `Repr` has two `Option<Align>`, which allows us to also exploit a niche!


charon/src/bin/charon-driver/translate/translate_crate.rs
```diff
@@ -739,7 +744,7 @@ impl<'tcx, 'ctx> ItemTransCtx<'tcx, 'ctx> {
         let mono =
             self.monomorphize() && (kind.is_for_trait() || !self.item_src.kind.is_for_trait());
         let item_src = TransItemSource::from_item(&item, kind, mono);
-        if enqueue {
+        if enqueue && !self.is_poly_in_mono(&self.item_src) {
```
Nadrieril: I'd rather this acted at the level of opacity computation, rather than ad-hoc here

N1ark: done; poly items are excluded (opaque didn't suffice); this lightly impacts names but I think it's ok ?

Nadrieril:
Very nice

PR #1451, opened 2026-09-17
Fix panic for an `FnDef` with bound vars

jdonszelmann:
I ran into this trying charon on my own project. The original project used `winnow`, a parser library, but the reproducer is barely recognizable as such. To the best of my knowledge, looking at rustc's source and analyzing traces of the compiler, newer rustc versions fix this problem. 

> [!NOTE]
> I've not used an LLM for any part of this PR, or any other PR I make. This includes any related work like research.


Nadrieril:
I updated the rustc toolchain! Could you try your test again on latest charon, and if it passes let's remove the fix and keep just the test

jdonszelmann:
This still exists on the newer nightlies. At some level, higher bound vars are to be expected here and properly instantiating is the way to go. For now this fix at least avoids an ICE.

Nadrieril:
We'll have to translate them eventually but agreed, let's just fix this first.

charon/src/bin/charon-driver/hax/types/ty.rs
```diff
@@ -722,7 +722,11 @@ pub enum TyKind {
 
     #[custom_arm(
         ty::TyKind::FnDef(fun_id, generics) => {
-            let generics = generics.no_bound_vars().expect("bound variables in FnDef");
+            // trying to work out where the binder here even comes from, cause rustc doesn't have it
+            // so you must be rebinding somewhere
```
Nadrieril:
```suggestion
```

Nadrieril:
Charon moves fast, the test output is already outdated :) Could you update it?

jdonszelmann:
note: happy to remove the .gitignore change. I don't like .direnv in history and it was getting annoying

Nadrieril:
I just have a global ignore for `.direnv` so I never noticed ^^ fine to keep it

PR #1454, opened 2026-09-17
Compute guaranteed layouts

N1ark:
Port the layout guarantees part of #1343 here, into the `compute_layout_guarantees` pass. I took some liberties in reformatting the code, i find it a bit clearer this way!

In the process I also fixed a couple small bugs in the original PR:

- the first field of a `repr(C)` enum is in fact not placed after the tag aligned to its own alignment, but to the max alignment of *all fields of this enum*. Basically, these enums are translated as `struct { tag, union { fields... }}`.
- uninhabited fields and variants were ignored in size/alignment calculations, which is incorrect (e.g. `(u32, !)` has alignment 4).

there's still no aggressive normalisation (aka "concretisation"). I think i'd rather keep it this way for now.


charon/src/ast/items/item_ids.rs
```diff
@@ -402,6 +402,15 @@ impl From<FunDeclRef> for FnPtr {
         FnPtr::new(fn_ref.id.into(), fn_ref.generics)
     }
 }
+impl From<&TypeDecl> for TypeDeclRef {
+    fn from(decl: &TypeDecl) -> Self {
```
Nadrieril: Hm, I'd prefer for this to be a method like `self_ref` on `TypeDecl`, as this conversion is only valid in the context of the typedecl itself

charon/src/ast/items/layout_guarantees.rs
```diff
@@ -20,14 +28,13 @@ use serde_state::{DeserializeState, SerializeState};
 pub enum OffsetGuarantee {
     /// Guaranteed to be at offset zero. This applies for `repr(transparent)` and in some `repr(C)` cases.
     AtOffsetZero,
+    /// Guaranteed to be exactly at the given offset.
+    AtOffset(SizeExpr),
```
Nadrieril: We can definitely merge these

charon/src/transform/add_missing_info/compute_layout_guarantees.rs
```diff
@@ -7,6 +8,207 @@ use crate::transform::{TransformCtx, ctx::TransformPass};
 
 pub struct Transform;
 
+struct Guarantees {
+    size: SizeExpr,
+    align: SizeExpr,
+    offsets: IndexVec<VariantId, IndexVec<FieldId, OffsetGuarantee>>,
+}
```
Nadrieril, line 15: Is that struct useful, compared to just taking a `&mut Layout`?

Nadrieril: yea ok I like it

charon/src/transform/add_missing_info/compute_layout_guarantees.rs
```diff
@@ -7,6 +8,207 @@ use crate::transform::{TransformCtx, ctx::TransformPass};
...
+            offsets,
+        });
+    }
+
+    // `repr(Rust)` only guarantees that the fields are aligned and don't overlap, and that the
+    // alignment is at least that of the fields.
+    // <https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.rust.layout>
+    let fields = variants.iter().flat_map(|(_, fields)| fields);
+    let align = type_align(fields.clone().map(field_align).collect());
+    let size = SizeExpr::max_size(fields.map(SizeExpr::size_of).collect());
```
Nadrieril: @firefighterduck's idea was to make this the max of `offset_of(ty, field) + sizeof(field_ty)`, I'd like that

N1ark: done :3

Nadrieril (approved):
Looks very good! Dunno what's up with the aeneas CI, I don't see how this PR could have changed it

N1ark:
the aeneas failure seems to be bc for some ADTs we have a layout that has `ConstantExprKind::OffsetOf`, which has a `TypeDeclRef`, that mentions the type itself, causing those types to be self-recursive, which changes outputs. i just made sure the SCC pass doesn't visit layouts, i think it's fine

PR #1455, opened 2026-09-17
Make structs for all `FullDefKind`s

N1ark:
As discussed on Zulip!

This first step only does two things (each in one commit):
- make a struct for all variants of `FullDefKind`, enabling us to do better encapsulation and lazily calculate fields in the future
- stop pattern matching on fields of all of these structs, as they may become private in the future, in favour of accessor methods. it's not the case right now and some of these will likely stay public but it felt cleaner to just apply the rule for everything (it also made some code nicer:3)


Nadrieril:
let's gooo

PR #1456, opened 2026-09-17
Generate VTables for function pointers

N1ark:
The last part of the vtable series (I forgot to open the PR for this woops)

- Translate the `Fn*` trait methods for function pointers _in monomorphic mode_ -- we use `TransImplSource::Marker` for these, as conceptually they kind of match. I don't mind doing this on a separate source if you'd prefer.
- Tweak VTable generation to also translate VTables of marker trait impls when necessary (in our case this is only used for function pointers).

We can't do this change for polymorphic mode, because:
- the `Args` type might not be concrete, making it impossible for us to untuple/retuple it in the shim
- we can't soundly create a synthetic `DefId` for function pointers, because they may have late-bound regions, that we can't currently represent with the machinery in hax (more specifically: we are missing an early binder).


Nadrieril:
As discussed in DMs, this won't work once traits in mono have keep their `Self` parameter. Also I'd quite like a solution that works for poly too.

I propose a shape-extraction idea: given an `fn(...)` type, make as many lifetimes and types fresh without breaking the higher-order structure, and use that as a `DefId` that takes the fresh parameters as input. So e.g. `for<'a> fn(&'a u32) -> &'a u32` becomes the shape `for<'a> fn(&'a A) -> &'a B` applied to `<u32, u32>`.

It's a bit crazy but I don't see another way to handle the poly case.

Nadrieril:
Cf https://github.com/AeneasVerif/charon/pull/1478 :3

Nadrieril:
Is there anything missing that #1478 doesn't cover?

N1ark:
> Is there anything missing that #1478 doesn't cover?

i dont think so! if there is ill open PRs but this looks fine

Issue #1449, opened 2026-09-16
Feature request: annotate unsafe operations


Nadrieril:
Let's add an is_unsafe() method to Statement so that users can lint on unsafe use, should be easy enough.

Nadrieril:
The unsafe operations are exhaustively listed here: https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html

PR #1457, opened 2026-09-17
Make some of hax's `FullDefKind` lazy

N1ark:
Built on top of #1455

- First, we make all fields of the new `FullDefKind` structs private, adding functions to access them. 
- We then remove all fields that are never accessed (i am a bit doubtful abt the fact we never read a static's unsafety/mutability... this looks like Charon isn't encoding enough info, but oh well)
- Then, we compute all the proofs for the `Call*` traits (and other similar fields) lazy, since they are actually only needed for VTables, which we don't always translate.
- Finally, we make trait items lazy too! This is one of the biggest wins; especially in `--monomorphize`, where we practically never query these items, and they cause massive slowdowns

Some benchmarks! This is time taken for the file setting only the given preset and `--no-serialize`. `bench.rs` is a test that exercises the std (HashMap, HashSet, Box, Arc, Vec...).

| file | fast | soteria | aeneas |
|---|---|---|---|
| bench.rs | 0.25 → 0.19 (−24%) | 2.03 → 1.06 (−48%) | 0.27 → 0.20 (−26%) |
| [address-reuse.rs] | 0.14 → 0.12 (−14%) | 0.70 → 0.39 (−44%) | 0.15 → 0.12 (−20%) |
| [gosim-demo.rs] | 0.14 → 0.12 (−14%) | 0.76 → 0.42 (−45%) | 0.14 → 0.12 (−14%) |
| [result-unwrap.rs] | 0.10 → 0.10 (—) | 0.33 → 0.20 (−39%) | 0.11 → 0.11 (—) |
| [closures.rs] | 0.11 → 0.11 (—) | 0.30 → 0.19 (−37%) | 0.13 → 0.12 (−8%) |

[address-reuse.rs]: https://github.com/rust-lang/miri/blob/master/tests/pass/address-reuse.rs
[gosim-demo.rs]: https://github.com/AeneasVerif/charon/blob/main/charon/tests/ui/gosim-demo.rs
[result-unwrap.rs]: https://github.com/AeneasVerif/charon/blob/main/charon/tests/ui/result-unwrap.rs
[closures.rs]: https://github.com/AeneasVerif/charon/blob/main/charon/tests/ui/closures.rs


charon/src/bin/charon-driver/hax/types/new/full_def.rs
```diff
@@ -803,37 +776,19 @@ impl AssocConst {
 #[derive(Clone, Debug)]
 pub struct Static {
     param_env: ParamEnv,
-    /// Whether it's a `unsafe static`, `safe static` (inside extern only) or just a `static`.
-    safety: Safety,
-    /// Whether it's a `static mut` or just a `static`.
-    mutability: Mutability,
```
Nadrieril: I'd prefer to keep these even if we don't use them at the moment. I'll need the safety for #1449 for example

charon/src/bin/charon-driver/hax/types/new/full_def.rs
```diff
@@ -550,21 +557,91 @@ impl TraitImpl {
         &self.implied_trait_proofs
     }
     /// Associated items, in the order of the trait declaration. Includes defaulted items.
-    pub fn items(&self) -> &[ImplAssocItem] {
-        &self.items
+    pub fn items(&self, s: &impl BaseState<'tcx>) -> Vec<ImplAssocItem> {
```
Nadrieril: From an API perspective it feels really expensive to compute it fresh every time; let's store that in a OnceCell

N1ark: for the record i tried a oncecell exactly here and it had no impact, but yes probably is the safer choice

Nadrieril (approved):
Very happy about this, feels like hax is starting to take shape as a crate that makes sense on its own. One question I have is: maybe we want to put all these inside `OnceCell`s in the cases where the computation is a bit expensive? We'd be making a baby query system essentially.

N1ark:
:D

> maybe we want to put all these inside `OnceCell`s in the cases where the computation is a bit expensive? We'd be making a baby query system essentially.

yeah i guess it's a bit of a footgun if in the future we use these things more and don't realise that we're recomputing stuff. added a `OnceCell` around items, implied trait proofs, and the proofs for callables, we can always tweak this later but I think it's fine as is

PR #1458, opened 2026-09-18
Minor perf improvements

N1ark:
Tiny fixes that give me about 20% perf improvements for Soteria! i don't think there's anything particularly controversial


Nadrieril:
Huh that's a huge improvement for so few lines changed!

PR #1459, opened 2026-09-18
Fix misaligned anon consts

N1ark:
All anon consts were translated as `MaybeUninit<[u8; N]>`, which has align 1, which meant that e.g. `const { &5u32 }` would make an allocation that's 1-aligned! Fixes these by instead making anon consts be `MaybeUninit<[uN; M]>`, where `uN` has the alignment of the allocation.

This does mean if a user specifies e.g. a struct with `#[align(64)]` and then has an anon const with that we can't represent it, since there is no primitive with that alignment... maybe we should have fields `align` + `size` in `GlobalDecl`?...


Nadrieril:
> maybe we should have fields align + size in GlobalDecl?

Yeah that would make sense to me

N1ark:
done! reusing layout's `Size` so we can do fancy things if we want to

PR #1460, opened 2026-09-18
Be explicit about lazy vtables

N1ark:
This is the result of experimenting with skipping more VTables, which turned out to not matter at all for performance. Still, I feel like being more explicit about how we translate the vtable of trait impls is probably nicer.

The status quo is:
- VTables are not enqueued by the trait impl containing it, except in monomorphic mode. However they get a `GlobalDeclId` generated.
- them being named then enqueues that VTable, populating that id in the list of globals.

I find this a little awkward, as we end up with dangling global IDs that don't really point to anything. the mechanism is itself not really made explicit anywhere, as far as i know. 

Instead, we now have an explicit `--eager-vtables`, that makes populating VTables eager. dyn-compatible trait impls by default have `vtable: VTableDecl::Lazy`, which then get replaced if and when the vtable is generated. dyn-incompatible trait impls get `VTableDecl::NotDynCompatible`. Finally, I also added `VTableDecl::Unknown(String)`, for the 2-3 cases where we should maybe translate a vtable and don't


Nadrieril (approved):
nice

PR #1461, opened 2026-09-18
Small speedups and fixes

N1ark:
Nothing crazy again, but this brings a nice 10% improvements in some worst case scenarios.

- don't generate the drops of no-drop types
- don't generate unreachable methods of closures in monomorphic mode
- avoid having hax generate all monomorphic instances of trait declarations, when we only use information that doesn't care about the generics

possibly the first change can be made a flag, if you want to preserve the existing behaviour! the goal is just to reduce the number of drops that Charon generates, as it generates a lot of unreachable items that slow down translation a bit. i also want to focus on improvement the performance per item, but that is proving increasingly hard as im running out of easy tricks :P 

ci: use https://github.com/AeneasVerif/eurydice/pull/452


charon/src/bin/charon-driver/translate/translate_items.rs
N1ark, line 1457: to clarify, we replace the `hax_def` above with `poly_hax_def`, because we only access the id and kind of its associated items, which are the same regardless of generics. this avoids costly constructions of the same hax TraitDecl with different sets of instantiated arguments

Nadrieril (approved):
LGTM! I'm just surprised about the soteria preset changes ^^

PR #1463, opened 2026-09-20
Minor perf improvements

N1ark:
Lots of smaller perf improvements. They are all safe contained, minor algorithmic improvements. Across a small set of benchmarks, this branch has a 10-20% perf improvement and 15-30% instruction count reduction!


charon/src/errors.rs
```diff
@@ -169,35 +169,28 @@ enum DepNode {
...
     }
 }
 
 impl std::fmt::Display for DepGraph {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
-        for (from, to, _) in self.dgraph.all_edges() {
+        for (from, to) in &self.edges {
             writeln!(f, "{from:?} -> {to:?}")?
         }
         Ok(())
```
Nadrieril: I don't think we use that impl, let's remove it since it's not order-stable anymore

Nadrieril (approved):
Happy with 9/10 commits! The last one feels redundant with `dyn_visit_in_body`

Nadrieril:
Got conflicted with another PR :)

PR #1464, opened 2026-09-20
Cache trait proofs

N1ark:
Add caching of types trait proofs across owners! We also now only compute trait proofs lazily (with a cache). This is slightly delicate, as this only applies to types/proofs that are entirely concrete *and* for which the environment has no concrete `where` clauses, as those would take precedence over any resolution.

I added a test for this corner case, to be safe. This brings an 5-20% perf improvement on monomorphic benchmarks, and a 12-20% instruction count reduction.


charon/rustc_trait_elaboration/src/item_ref.rs
```diff
@@ -148,6 +148,9 @@ pub struct ItemRefContents<'tcx, Id: ItemId = DefId> {
     pub needs_explicit_self_clause: bool,
     /// Whether this contains any reference to a type/lifetime/const parameter.
     pub has_param: bool,
+    /// Whether this mentions the generic parameters of the item under which the reference was made.
+    /// If `false`, the trait proofs for this reference don't depend on the owner.
+    pub has_owner_param: bool,
```
Nadrieril, line 151: That comment doesn't follow: not mentioning the generic params is not enough for the trait proofs not to depend on the owner, as you mention in the PR description

PR #1465, opened 2026-09-20
Avoid creating Hax `Def`s when possible

N1ark:
Building a Hax `FullDef` is expensive! In particular, we calculate a lot of them eagerly, when we may not really need that item. We also often use `hax_def`, which monomorphises, when `poly_hax_def` suffices; this should be preferred, to improving sharing in monomorphic mode.

This PR avoid eagerly creating `FullDef`s when possible, favouring either using `tcx` or using `hax::DefId.kind` when possible.

Separately, the last PR adds a `HasErasedOrBodyRegions::{Yes,No}` enum to the type cache; types that have no erased/body regions aren't visited by the lifetime freshener.

This PR brings a 2-8% perf improvement and a ~2% instruction count reduction. Nothing big but it's practically free :D


charon/src/bin/charon-driver/hax/types/new/item_ref.rs
```diff
@@ -272,6 +272,33 @@ impl ItemRef {
     pub fn with_def_id<'tcx, S: BaseState<'tcx>>(&self, s: &S, def_id: &DefId) -> Self {
         self.mutate_def_id(s, |d| *d = def_id.clone())
     }
+
+    /// The item whose generics this item's generics extend, if any: the impl or trait of an
+    /// associated item, the enclosing item of a closure or anonymous constant, the type of a
+    /// variant or constructor. It is instantiated with the corresponding prefix of our generic
+    /// args. This is cheap: it doesn't require computing the `FullDef`.
```
Nadrieril:
```suggestion
    /// The parent of this item if the item inherits the typing context from its parent (e.g. closures and methods).
```

charon/src/bin/charon-driver/translate/translate_generics.rs
```diff
@@ -61,7 +62,23 @@ pub(crate) struct BindingLevel {
     /// Cache the translation of types. This harnesses the deduplication of `Ty` that hax does.
     // Important: we can't reuse type caches from earlier binders as the new binder may change what
     // a given variable resolves to.
-    pub type_trans_cache: HashMap<hax::Ty, Ty>,
+    pub type_trans_cache: HashMap<hax::Ty, (Ty, HasErasedOrBodyRegions)>,
```
Nadrieril: For this one, I'd like to do the more general solution of storing a couple flags inside `Ty` itself, like rustc does

N1ark: sure :) for now i can add a `u8` newtype and we do bitflags if you want? how does that sound

Nadrieril: Let's make this a separate PR, I'm thinking some kind of `WithCachedTypeInfo<TyKind>` that gets recomputed at interning time, that also applies to `TraitRef` and that visitors take into account

N1ark: done :3 ill let you do that then, nws

PR #1468, opened 2026-09-20
Make auto-trait ordering stable in trait objects

N1ark:
Turns out the ordering the compiler gives us is not stable across targets or compiler versions :P


Nadrieril:
For reference (thanks @N1ark):
- [where the sorting happens](https://github.com/rust-lang/rust/blob/923c95cdf5ba65cea505aa2ea829f578e1506ed8/compiler/rustc_middle/src/ty/predicate.rs#L223);
- [how the compiletest output stays stable](https://github.com/rust-lang/rust/blob/923c95cdf5ba65cea505aa2ea829f578e1506ed8/compiler/rustc_middle/src/ty/print/pretty.rs#L1474-L1481).

PR #1469, opened 2026-09-21
Fix panics in foreign statics and consts with fn ptrs

N1ark:
Fix two panics encountered while running the Miri suite:
- a panic when trying to evaluate foreign statics
- a panic when translating a constant (with `--consts values`) that has a function pointer that is not just a function (e.g. closures, or builtin impls)


Nadrieril:
Very nice fixes

PR #1473, opened 2026-09-22
Global cache of concrete trait resolutions

N1ark:
Separately from #1464 -- add caching to the trait solver! This improves perf on `--preset soteria` by 5-15%, but doesn't do much to `--preset aeneas`


Nadrieril (approved):
Ya that looks good

